PR #13, opened 2011-07-25
Change toJSON and fromJSON instances for Either.

jhance:
Previously, fromJSON . toJSON == id was violated by the Either
instances. In addition, if both types are parsable from the same
JSON, then the Left is always given favor - which is probably not
desirable. Instead, differentiate the instances using an array.

This allows for "Either a a" to actually work instead of always
chosing Left in FromJSON.

See https://github.com/mailrank/aeson/issues/11


amigalemming:
What about providing an `Alternative` data type that maintains the old behaviour of `FromJSON Either`?

PR #209, opened 2014-06-28
Issue 175 (Order of ZonedTime parsers)

### LINKED ISSUE #175: Order of ZonedTime parsers
The [current instance for `ZonedTime`](https://github.com/bos/aeson/blob/be4cfc8839d8e62c67bdce9c3df3ceeecc0a821a/Data/Aeson/Types/Instances.hs#L491) uses:

```haskell
        alternateFormats = --(\x -> traceShow ("alternateFormats", x) x) $
          dateTimeFmt defaultTimeLocale :
          distributeList ["%Y", "%Y-%m", "%F"]
                         ["T%R", "T%T", "T%T%Q", "T%T%QZ", "T%T%Q%z"]
```

I have an application where JSON date parsing takes around 30% of the CPU time (short events with time stamps). It uses Javascript's `new Date().toISOString()`.

When I moved `"T%T%Q"` to the front of that list, time parsing became much faster since `"%FT%T%QZ"` is the format that `toISOString()` outputs.

Given that `toISOString()` is a very common format when you care about zoned times, maybe it makes sense for it to be moved to the front by default?

--

Another question: In any case, `dateTimeFmt defaultTimeLocale` is the first entry in that list. Why does that make sense? It is `"%a %b %e %H:%M:%S %Z %Y"` - is this actually a valid/sensible JSON date (I don't think one can say "valid" since there are no real dates in JSON and they are just strings - but sensible in the way of "does any of the common JS functions create it like this")? Maybe it should come after the other options?

Thanks!

zerobuzz:
This should fix #175.  Please feel free to cherry pick, or let me know what you want changed and I'll compose a new pull request.

(In particular, I have placed the relevant changes, various style and taste issues, and the slightly insane test case in separate commits.  I'm ok with dropping anything as long as I can get 1fb1e78 in.)


bos:
Applied with rebase, thanks.

Issue #215, opened 2014-07-29
Leading whitespace breaks parsing


tac-tics:
In Data.Aeson.Lens, I noticed that leading whitespace causes the parse to fail:

    *Main Data.Aeson.Lens> (" {}" :: String)^?! _Object
    fromList *** Exception: (^?!): empty Fold
    *Main Data.Aeson.Lens> ("{}" :: String)^?! _Object
    fromList []

The spec doesn't make much commentary on where whitespace is allowed ("between any pair of tokens", so perhaps technically no). However, leading whitespace is valid in Python's JSON library. And in Aeson, trailing whitespace causes no issue.

Mr. Kmett says that this behavior is inherited from this package. 

This bug cost me maybe an hour to track down. It was easy to work around, but I'm sure others are likely to bump into it.

PR #218, opened 2014-09-02
Move skipSpace from value callsite to inside the definitions of value and value'

tac-tics:
A three-liner for addressing: https://github.com/bos/aeson/issues/215#issuecomment-52565616

I tested this with the following file:

    module Main where
    
    import Data.Text
    import Control.Lens
    import Data.Aeson.Lens
    
    main = print $ (pack " {}") ^?! _Object

I don't know the Haskell testing frameworks well enough to add the test to the project. If anyone could give me some basic direction on the matter, I would add them.


bos:
Applied manually, thanks.

PR #225, opened 2014-10-16
Add missing instances for Data.Sequence.Seq

jcpetruzza:
They behave just like the instances for `[a]`


bos:
Applied manually, thanks.

PR #226, opened 2014-10-25
ToJSON and FromJSON instances for Identity

supki:
I've found these useful when dealing with the particular kind of parameterized types, e.g.

```
data T f = T { foo :: f String, bar :: f Int, ... } 
```


bos:
Applied by hand, thanks.

Issue #83, opened 2012-07-05
Unexpected behavior of .:?


lykahb:
I want to have distinction between absence of field and its null value. 

```haskell
data Blah = Blah (Maybe (Maybe String))
instance FromJSON Blah where
  parseJSON (Object v) = Blah <$> v .:? "blah"
```

```haskell
Blah Nothing -- if field not mentioned
Blah (Just Nothing) -- if field has null value
```

However, the result of parsing is never `Blah (Just Nothing)` because `.:?` uses `(Maybe a)` instance directly.

```haskell
(.:?) :: (FromJSON a) => Object -> Text -> Parser (Maybe a)
obj .:? key = case H.lookup key obj of
               Nothing -> pure Nothing
               Just v  -> parseJSON v
            -- Just v  -> Just <$> parseJSON v would have expected behavior
```

lykahb:
Writing a testcase was harder than I expected since monadic Parser does not play well with properties.

isExistingKeyJust :: Value -> Bool
isExistingKeyJust (Object obj) = all (\k -> isJust $ obj .:? k) $ H.keys obj -- does not typecheck
isExistingKeyJust _            = False

gregwebs:
This is committed as d0414be92ee6bf5b4c057978955b89d111767dab
However, it is not in the Changelog even though this is a huge breaking change.
I personally like the change, however, I think it may cause so much breakage that the existing behavior should be maintained and a different operator should be added instead., perhaps `.:??`

gregwebs:
Should this be closed now?

PR #232, opened 2014-12-18
Fix .:? behavior for #83

lykahb:
This commit fixes the .:? operator and correspondingly changes parsing code to preserve JSON roundtrip properties.

### DIFF
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -94,8 +94,8 @@ import qualified Data.Aeson.Encode.Functions as E
 import qualified Data.Aeson as A
 -- from base:
 import Control.Applicative ( pure, (<$>), (<*>) )
-import Control.Monad       ( return, mapM, liftM2, fail )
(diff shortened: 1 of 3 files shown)


bos:
I had a history hiccup, and need you to rebase on top of the new history. Sorry!

Also, if you could please include an example in your commit description of what used to fail and now works, that will be important in understanding what this change actually does.

And finally, this PR is functionally identical to #248 but structurally different – please discuss with @sol which one should actually go in.

lykahb:
The change in Instances.hs is identical in my and @sol pull request. The difference is how TH and Generics treat nested Maybes. Without joins the behavior would change when (.:?) returns Just Nothing. The tests catch this and fail.

I've added a more detailed description.

bos:
Rebased and applied as d0414be92ee6bf5b4c057978955b89d111767dab, thanks! Any chance you'd like to submit a regression test for it? :-)

PR #220, opened 2014-09-23
Add json path to parse error messages

maximkulkin:
Every time you parse some JSON and get error message, it is very painful to track where the error is because existing error messages do not point to the place in JSON where parse error occured. JSON *syntax* errors are reported from Attoparsec parser and it gives position. But JSON *conversion* errors have no support for tracking position/context so quality of error messages is poor.

This change adds json path to error message:
1. Parser tracks current JSON path (in reverse order)
2. An `<?>` operation is added to annotate Parser with JSON context (JSON path). This operation is named after Attoparsec's `<?>` which serves similar purpose.
3. Built-in FromJSON instances are updated to provide context information
4. `.:` and `.:?` operations automatically add context information.

Context is reported in form of [JSON-Path](https://code.google.com/p/json-path/) expression e.g. `$.store.book[0].title`:
```
> eitherDecode storeJsonString :: Either String Store
Left "Error in \"$.store.book[0]\": key \"title\" not present"
```

Caveats:
1. There is no tests for error message yet. I believe those should be done in normal xUnit assertion style instead of QuickCheck property style.
2. Only first error is reported. Further work can include supporting collection of error messages and reporting all of them at once. This can be useful e.g. for validating HTTP JSON requests and reporting all errors at once for displaying them to user on a web form.

PR #234, opened 2015-01-05
Add a flag to omit empty content fields when generating tagged objects

plaprade:
This patch adds a flag to the SumEncoding option when generating JSON instances through Template Haskell or GHC Generics. The new flag prevents the generation of an empty contents field for nullary constructors in the presence of mixed constructors.

#### Motivating Example

```haskell
module Main where

import Data.Aeson
import Data.Aeson.TH

data A = A | B { value :: String }
    deriving Show

$(deriveJSON defaultOptions ''A)
```

This code will generate the following JSON data for A and B respectively:

`encode A`
```json
{
  "tag": "A",
  "contents": []
}
```
`encode (B "Hello World")`
```json
{
  "tag": "B",
  "value": "Hello World"
}
```

The contents field for the constructor **A** is redundant as nullary constructors do not have any arguments to encode. This information is available when generating the JSON instances. To produce cleaner JSON output, I added the `omitEmptyContents` field in the `SumEncoding` type.

#### New Behavior

```haskell
module Main where

import Data.Aeson
import Data.Aeson.TH

data A = A | B { value :: String }
    deriving Show

$(deriveJSON 
    defaultOptions{ sumEncoding = defaultTaggedObject{ omitEmptyContents = True } } 
    ''A
 )
```

Which will produce the following JSON data for A and B:

`encode A`
```json
{
  "tag": "A"
}
```
`encode (B "Hello World")`
```json
{
  "tag": "B",
  "value": "Hello World"
}
```

This new flag only works for union types where there are at least 2 constructors. A type with only 1 nullary constructor will be encoded to an empty array, which is the same behavior as before this patch. The `omitEmptyContents` flag defaults to `False` for backwards compatibility.

Any comments are appreciated.

### DIFF
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -403,11 +403,12 @@ sumToEncoding opts multiCons conName exp
         case sumEncoding opts of
           TwoElemArray ->
               [|Array|] `appE` ([|V.fromList|] `appE` listE [conStr opts conName, exp])
-          TaggedObject{tagFieldName, contentsFieldName} ->
-              [|object|] `appE` listE
-                [ infixApp [|T.pack tagFieldName|]     [|(.=)|] (conStr opts conName)
-                , infixApp [|T.pack contentsFieldName|] [|(.=)|] exp
-                ]
+          TaggedObject{tagFieldName, contentsFieldName, omitEmptyContents} -> do
+              isEmptyExp <- (==) <$> exp <*> [e|toJSON ([] :: [()])|]
+              let tag = infixApp [|T.pack tagFieldName|] [|(.=)|] (conStr opts conName)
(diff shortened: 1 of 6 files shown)


basvandijk:
This is a good feature to have.

I reviewed and tested the patch and I'm ok with merging this in.

@bos since this modifies the `SumEncoding` type this requires a major version bump (0.8.0.2 -> 0.9.0.0).

I'm going to make a 0.9.0.0 release which will also include at least: #220.

plaprade:
I dropped an inline statement by mistake and added it back in.

plaprade:
Hi, my pull request was a bit off from the last Aeson release so I made the necessary changes and rebased it back on the current master.

plaprade:
This pull request doesn't work anymore. I'll close it for now until I have time to fix it.

Issue #180, opened 2014-01-29
Consider adding back escaping of '<' and '>' to prevent XSS injection


meiersi:
I've just noticed that aeson-6.2.1 also escaped '<' and '>' characters (see for example https://github.com/bos/aeson/blob/6bae92494c1aeae4befdf2a87e5d9cd026b40e8b/Data/Aeson/Encode.hs#L79).

The new encoder does not do that. @bos, should I re-add the behaviour? When exactly is this escaping preventing an XSS injection? Note that in terms of encoding size this escaping is quite costly  when embedding HTML documents in JS strings. So, I'd like to understand why it is needed. I also suspect that it would be sufficient to only escape '<', as then it would be impossible to start a tag.

avieth:
+1 for *not* escaping these, because there are applications which demand human-readable JSON. For instance, programmatically building or altering a node.js package.json file. It's not a huge cost: the package.json file will be interpreted correctly even with the `<, >` characters escaped, but it will be *slightly-less* readable.

lpsmith:
Well, you can't build a correct json-in-html escaping function from a correct json escaping function and a correct html escaping function.   So while I agree that escaping angle brackets is undesirable in some cases,   it's also necessary in a very common use case.

I don't really have an complete answer on this count.    It's kind of an ugly problem that's arisen from the lack of syntax engineering that went into embedded JavaScript.

snoyberg:
I was unaware of this change, and it's subtly introduced a potential for XSS attacks in Yesod. See this thread:

https://groups.google.com/d/msg/yesodweb/M2VS5OTwyPg/AnN5nAm1AgAJ

Note that this is _not_ about embedding unsanitized data in an innerHTML. This can be triggered by a perfectly safe usage of sanitized data being embedded inside an innerText. The problem comes from a string like:

    "</script><script>alert('hacked!')</script>"

Before this change, the less than signs would be escaped, meaning this was safe to embed inside a `<script>` tag, and the resulting string could be used for `innerText`. That's no longer the case.

mbj:
JSON encodes opaque generic strings that are represented in UTF-8. A JSON encoder should use the most efficient representation, only triggering escapes when technically required. This follows the rule of least power, a decent engineering axiom.

Many data formats, including HTML can be encoded into that generic JSON strings, even JSON itself. Hence this library nor any other JSON encoder should consider to add format specific special cases.

[...]

PR #237, opened 2015-03-10
encodeToTextBuilder does not escape > and <

avieth:
`encodeToByteStringBuilder` and `encodeToTextBuild` are inconsistent: the former will not escape `>` or `<`, but the latter will!

```

import Data.ByteString.Builder (toLazyByteString)
import Data.Text.Lazy.Encoding (encodeUtf8)
import Data.Text.Lazy.Builder (toLazyText)
import qualified Data.ByteString.Lazy as BL
import Data.Aeson hiding (encode)
import Data.Aeson.Encode

theJson :: Maybe Value
theJson = decode "{ \"foo\" : \"< bar >\" }"

main = do
  let Just jsonValue = theJson
  let encoded1 = encodeUtf8 (toLazyText (encodeToTextBuilder jsonValue))
  let encoded2 = toLazyByteString (encodeToByteStringBuilder jsonValue)
  -- > and < are escaped.
  BL.putStr encoded1
  BL.putStr "\n"
  -- > and < are not escaped.
  BL.putStr encoded2
```

This program prints two different strings, but I would expect them to be the same. With this patch, they are the same.

Relevant to issue #180


bos:
Applied by hand, thanks.

Issue #240, opened 2015-03-20
Output UTCTime to microsecond precision


timmytofu:
Would you consider this?

```
formatMillis :: (FormatTime t) => t -> String
formatMillis = take 3 . formatDecimal

formatMicros :: (FormatTime t) => t -> String
formatMicros = take 6 . formatDecimal

formatDecimal :: (FormatTime t) => t -> String
formatDecimal = formatTime defaultTimeLocale "%q"

instance ToJSON UTCTime where
    toJSON t = String $ pack $ formatTime defaultTimeLocale format t
      where
        format = "%FT%T." ++ formatMicros t ++ "Z"
    {-# INLINE toJSON #-}

instance FromJSON UTCTime where
    parseJSON = withText "UTCTime" $ \t ->
        case parseTime defaultTimeLocale "%FT%T%QZ" (unpack t) of
          Just d -> pure d
          _      -> fail "could not parse ISO-8601 date"
    {-# INLINE parseJSON #-}
```

or perhaps even not limiting it at all, except implicitly to the picosecond resolution of UTCTime.

I see in the source that you mention ECMA262 keeping dates to millisecond precision, and this is true. However, ISO-8601 does not limit precision to three places, and JS will parse a longer decimal fine, it will merely truncate it to milliseconds.

For us, the advantage of having that extra precision (in the string) is that postgres and mysql both have microsecond precision, and oracle defaults to microsecond precision (though it allows nano as well). So, if you're using a timestamp as the [primary dimension of a paging cursor](http://use-the-index-luke.com/sql/partial-results/fetch-next-page), those extra digits can be significant. If I have an event with timestamp `2015-01-01T12:12:12.123456` and it gets truncated to `….123`, and events with timestamps between `….123001` and `….123456` will get skipped when the truncated `….123` timestamp comes back on the next request (barring us doing some explicit database malarky).

Granted this only works if the consumer of this JSON returns the timestamp cursor as received, rather than parsing and re-stringifying (if ECMAScript), and granted we could accomplish this with a little newtype wrapper with the instance above, but I think this is a pretty common usecase of timestamps and having to figure out where to hammer in the newtype seems like unneccessary extra work.

[...]

PR #241, opened 2015-03-24
Updates UTCTime ToJSON instance to output full picosecond resolution

timmytofu:
See bos/aeson#240.

Microsecond is good enough for my needs but seems arbitrary, so this doesn't limit at all. All output will have twelve decimal places - there's no way to get the actual precision out of UTCTime.


bos:
Applied, thanks.

PR #242, opened 2015-03-27
Option to unwrap unary record constructors (Fixes #228)

### LINKED ISSUE #228: deriveJSON: Add an option to reduce newtype wrappings
Suppose that I have newtype driving of Text (for type safety at compilation time):

    import Data.Text (Text)

    newtype PkgName = PkgName {unName :: Text}
      deriving (Show, Eq, Typeable, Data)

    $(deriveJSON defaultOptions ''PkgName)

The generated JSON representation includes unName in a redundant level. It would have been nice to have the option to eliminate it, because we are simply only proxying the type at compilation time and don't need the proxing at run-time.

    case dec of
       DataD    _ _ tvbs cons _ -> f tvbs cons
       NewtypeD _ _ tvbs con  _ -> f tvbs [con]

I see we are basically passing the single constructor to newtype as a single-item list and ignore the fact it's a newtype. So I guess we need something more elaborate there.

JPMoresmau:
This provides an additional option, to unwrap record that have only one field (like newtypes). It does not generate an object but only the value.
Quickchecks tests have been added using the already present Approx newtype, for both TH and GHC Generic functions.

### DIFF
--- a/benchmarks/aeson-benchmarks.cabal
+++ b/benchmarks/aeson-benchmarks.cabal
@@ -66,6 +66,7 @@ executable aeson-benchmark-typed
     bytestring,
     criterion >= 1.0,
     deepseq,
+    ghc-prim,
     text,
     time
 


basvandijk:
Awesome! I just needed this at work.

I will try to review and merge it this weekend.

PR #244, opened 2015-04-10
Add AtKey and AtIdx

phadej:
While working with not-so-nice JSONs, I found a bit of type-level hackery useful.

What you think, could it be added to aeson, or should I make a separate package.

P.S. the tests are missing, but there are simple examples in haddock comments.


bos:
Thanks for the patch, but I'm afraid I'm not accepting code like this. I generally direct people to the `lens-aeson` package for needs like this.

Issue #246, opened 2015-04-15
encodeToTextBuilder hangs


andrewthad:
Here is a minimal example:

    import Prelude
    import Data.Aeson
    import Data.Aeson.Encode
    import Data.Text.Lazy.Builder as Builder
    import Data.Text.Lazy as LText
    import Data.Text as Text
    import Data.Text.IO as Text

    main = Text.putStrLn $ LText.toStrict $ Builder.toLazyText $ encodeToTextBuilder $ object [Text.pack "test" .= (2 :: Int)]

If I compile and run this, it does not print anything and never terminates. The package set I have installed is Stackage 2.1. The only unusual thing about my GHC install is that it is compiled with `integer-simple` instead of `integer-gmp`. If I call `encode` instead of `encodeToTextBuilder`, this problem does not happen. Can anyone reproduce this issue?

-- EDIT

Some more information. It only happens when trying to encode numbers (I'm not sure which kinds of numbers though). For example, if you replace `main` with the following:

    main = Text.putStrLn $ LText.toStrict $ Builder.toLazyText $ encodeToTextBuilder $ object [Text.pack "test" .= Text.pack "bla"]

Then it doesn't hang. It renders `{"test":"bla"}` as it should.

andrewthad:
yesodweb/yesod#967

zeitgeist:
I stumbled upon the very same issue some time ago. For reasons unknown to me, I did not file a bug report. :(

andrewthad:
Thank goodness it isn't just me.

zeitgeist:
@andrewthad I was using our company's build infrastructure, that is plain old GHC 7.8.4 compiled from sources w/ integer-simple

andrewthad:
Oh, so it's almost certainly a problem with it being incompatible with integer-simple. Either that or there's a bug in integer-simple.

andrewthad:
Or more likely in `scientific`.

basvandijk:
It's not `scientific` it's the `decimal` builder in `text` that is causing an infinite loop:

1. `encodeToTextBuilder` [calls](https://github.com/bos/aeson/blob/master/Data/Aeson/Encode.hs#L102) `decimal` on the `Integer` coefficient.
2. `decimal` [has the rule](https://github.com/bos/text/blob/master/Data/Text/Lazy/Builder/Int.hs#L63):  `{-# RULES "decimal/Integer" decimal = integer 10 :: Integer -> Builder #-}`
[...]

PR #249, opened 2015-04-29
Avoid (10 ^ e) on fromScientific

meteficha:
@basvandijk [noted that fromScientific used `10 ^ e`](https://github.com/bos/aeson/issues/246#issuecomment-93599848).  This operation could lead to a DOS due to a large exponent.

This PR changes the implementation to always use `formatScientificBuilder`, even when `base10Exponent s >= 0`.  The code from the scientific package already takes care when pretty printing huge numbers.

Note that using just `scientificBuilder` wouldn't be enough since we'd change the formatting of large integers to that of a large `Double`.  This commit does not change the output format at all, and all tests are green.

### DIFF
--- a/Data/Aeson/Encode.hs
+++ b/Data/Aeson/Encode.hs
@@ -29,10 +29,9 @@ module Data.Aeson.Encode
 
 import Data.Aeson.Types (Value(..))
 import Data.Monoid (mappend)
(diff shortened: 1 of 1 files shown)


meteficha:
I'm not sure if this is serious enough to warrant an announcement on relevant mailing lists asking users to upgrade.

PR #254, opened 2015-05-13
Use a better version of camelTo

k-bx:
For example, it handles a case "CamelAPICase" into "camel_api_case"

I'm not sure if we should just change current camelTo, or deprecate it for a new function with different name.

### DIFF
--- a/Data/Aeson/Types.hs
+++ b/Data/Aeson/Types.hs
@@ -62,6 +62,7 @@ module Data.Aeson.Types
     , Options(..)
     , SumEncoding(..)
     , camelTo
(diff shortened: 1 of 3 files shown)


bos:
I like this, but it's pretty cryptic.

Please give it a different name than the current camelTo function, e.g. camelTo2 – and please add some unit tests.

k-bx:
@bos ok done, please review!

bos:
Very nice, thanks for revising.

k-bx:
No problem, thanks for getting back to the tickets-pile :)

PR #255, opened 2015-05-18
Add possible to Aeson

tolysz:
Possible is a value which when `Missing` does not generate any entry in serialised JSON.
It is useful for patching and interacting with services which support PATCH semantics.

I use it a lot with (public ones):
 https://github.com/tolysz/google-api
 https://github.com/tolysz/video

--
I had to restore my github fork, so previous PR get lost, I tis the same but I have re-base it.


basvandijk:
I feel uneasy about adding the `Missing` constructor to `Value`. It doesn't represent a syntactical category of JSON which all the other constructors do.

Also doesn't it also allow you to create invalid JSON as in:
```haskell
encode $ Array (V.fromList [Missing, Missing])
``` 
resulting in `[,]`? That seems undesirable!

tolysz:
Without `Missing` there is no way to remove leafs from `Value` tree in a generic way.
You can write `ToJSON` for some type but it has to return something of type `Value`.
Now you can get around that working _level up_

    data A = A { field :: Possible Text }

In this example there is no way for `Possible` to decide that it does not want its output to be visible in the serialised `json`.
Currently there is no way to signal _up_ that you want to remove the output, there is no such constructor in the type for `Value`.

Sure you can write by hand serialise for `A`, or any other data type which will use it, but then you have to keep logic in many places.

But what if we want to go:

    data B b = B { field :: b }


As for `[,]` it has some uses with real-time updates, i.e. when you do not want to resend all data which did not changed; and this would be just one way. And Patch for array could be implemented in this way. Thus I would not add any extra code to forbid it; maybe fix the `FromJSON` to handle parse for it if people would start using it? Possibly suggest to `JSON RFC` that this could be a way to encode _sparse arrays_ .

The `Missing` value corresponds to the `undefined` value in JavaScript (`JSON.stringify({a:undefined}) === '{}'`) does not output anything. As the missing key in JS object have `undefined` as its value.

Adding `Missing` would allow to create `Patch` operations.

mgsloan:
Since this isn't part of JSON, it doesn't seem like it belongs in aeson.  Currently, aeson supports serializing and parsing JSON, this would extend it to support a superset of JSON.  I'd suggest creating a fork which supports this.

tolysz:
@mgsloan It is very easy to throw errors if you use `Missing` in arrays.
There is no word in JSON specs that you need to include __all__ fields of serialised object in JSON string representation; and that only this will be the only correct way of serialising and all other will not be part of JSON.
This is like arguing that `0` is not a natural number; By asking to show `zero` sheep or saying if you do not have sheep you do not need a number for it; you can only count sheep if you have them ;) and it is not a part of how you count them if you do not have them. (I know Nats start with 0 or 1 depending on defs)

`Missing` is only needed if you want to _compose_ types from smaller components; as it allows for type(component) to decide if it want to have some output. This would be very clear if type `Value` would be given more like a tree (it is a tree).
The way that current serialise works, it collapses leafs to type Value before output the result; so the lack of display needs to be of type `Value` if one wants to have an option to output nothing. Overriding other constructors is wrong as they _have_ output;

I do have branch & compile all libraries against the modified version. But it feels so much natural to have just this extra constructor for the Value type; and you do not store fields that have no additional information for you. But as long as Haskell have no 'ML' module system...

bos:
Thanks for submitting this pull request. I'm afraid I won't accept it, as I don't like the approach. Changing the fundamental 1:1 mapping between the `Value` type and the definitions in JSON is just not a clean approach, and does not seem well justified.

tolysz:
Sure, your call. Now I would need to ask JSON RFC to clearly include epsilon production in their format description and explicitly say that not all possible names have to be included every time.

I guess I can not argue it clearly why it is important ;) 
https://github.com/tolysz/google-api/blob/master/src/Network/Google/Api/Youtube/Videos.hs
`Possible` is similar to `Maybe` but have an option not to produce any output in `JSON`.
This is useful as I can use the same type to patch the content by submitting only data that changed. If I used `Maybe` it would delete it. With `Missing` included in data type I can make it in a generic way;  

I will try to come up with a better argument for it or maybe I will understand why I am wrong ;)

basvandijk:
> I guess I can not argue it clearly why it is important ;) 

I agree with you that it's important that we have a good way of handling null fields in aeson and that we can improve on that. However, I also agree with Bryan that Value needs to be a 1:1 mapping to JSON.

> Possible is similar to Maybe but have an option not to produce any output in JSON.

This is more a matter of personal opinion / taste but I don't like specifying encoding information (like 'Possible') in Haskell types. Haskell types should only describe information that is relevant to the application and shouldn't describe encoding-only information like Possible. I think the latter should be moved to the encoders like `deriveJSON`.

tolysz:
I will have a look at that generics approach in `aeson-0.10`, but from what I see you either do not use `null` or use it everywhere (if you want to derive it in a generic way).
With `deriveJSON` I would need to have a generic way to decide if the pair should be output: the same way as the `omitNothingFields` handles, but with decisions on field level (again if my custom type reduce to `Missing` while going `toJSON` the encoder should use a similar logic nut this time it is explicit).

PR #264, opened 2015-06-23
Fixes bug where constructor- and type name are swapped in error message

roelvandijk:
Since all parameters of ```parseTypeMismatch'``` are of type ```String``` it is easy to make a mistake. I choose the order of the error message itself as the right order.

Interestingly ```parseTypeMismatch``` also has a different order than the error message, but it is also called with the parameters in this order.


bos:
Nice, thanks.

Issue #171, opened 2013-12-14
Parser accepts garbage values for `Int` et al.


lfairy:
I've noticed two issues with the current conversion code:

1. Floating point values are silently coerced using `floor`:

        > decode "[1.1]" :: Maybe Int
        Just [1]
        > decode "[-1.1]" :: Maybe Int
        Just [-2]

2. The result isn't checked for overflow:

        > decode "[10000000000000000000]" :: Maybe Int
        Just [-8446744073709551616]

Is there a good reason for this behavior (performance, compatibility)?

PR #265, opened 2015-07-11
Use `truncate` instead of `floor` in Integral instances

lfairy:
Whether we should accept fractional parts at all is up to debate (#171).
But rounding toward negative infinity is wrong either way.

Truncation is the default in most languages, so it will cause the least
surprise to users of the library.


bos:
Thank you.

PR #266, opened 2015-07-12
Faster UTC time stamp parser

nurpax:
Request for comments!  I didn't add docs for the timestamp formats yet.  Wanted to get this out for review in case you have some early comments.

--

Replace the old 'parseTime' based UTCTime parser with an attoparsec
based parser.  This is roughly 5x faster than the old one.

It also allows decoding some timestamp formats that the old version
rejected.  The new version still requires all timestamps to specify a
timezone when decoding to a UTCTime type.

Also add a few unit tests for both what should be accepted and also
for things that should be rejected.

A lot of this code was bummed from the postgresql-simple library -
thus credits to @lpsmith too!


lpsmith:
At a glance,  I do very much agree with the overall spirit of the request.   And I see you've tweaked the parsers a bit to make them a little bit more lenient and robust,   some of which I might adopt back into postgresql-simple.    A few comments:

1.   I really do think that specifying an offset from UTC should be required for the UTCTime instance,  period,  and that aeson should stay firm on this requirement.

2.   Is there a FromJSON instance for LocalTime?   Perhaps it would be a good idea to add one as a way of dealing with timestamps that do not specify an offset.

3.  In light of the fact that the IANA tz database does specify offsets from UTC that contain seconds for almost all timezones,   perhaps we should parse offsets with seconds as well?

My third point is admittedly a fairly esoteric point;  if I recall correctly ISO 8601:2004 does not allow for seconds in the offset,  but the tz database is a very strong de-facto standard that is also worth consideration here.    Also,  all tz timezones moved to an offset that is an integer number of minutes from UTC by the 1970s.    Thus,  if you assume a provided offset actually corresponds to something in the tz database,  this issue only occurs when dealing with fairly old timestamps.

Michael Snoyman and I tackled this issue in lpsmith/postgresql-simple@b4624db41048e42e993d3bea820e907a010fde00,   although this code is making a number of assumptions that the input is sufficiently well-formed;   which is probably ok in the context of receiving data from a pg backend,  but wouldn't be a good idea to make in the context of some random bit of JSON.

lpsmith:
Hmm,  looking through the tz database,  I see that many offsets are specified with as `±H`, `±H:MM`, or `±H:MM:SS`,  perhaps this syntax should be accommodated as well.   I should read parts of ISO 8601:2004 again.

lpsmith:
Ok,  ISO 8601:2004 does explicitly require two digits in the month, day, time of day hour, minute, second, and offset hour and minute fields,  though some of these fields may be omitted altogether.   Notably,  however,  commas may be used as well as periods,  and for the time of day,  you can specify fractional hours (but then cannot specify minutes or secconds),   or you can specify fractional minutes (but then cannot specify fractional seconds.)    And of course,  there's the extended format,  which includes hyphens and colons,  (which most people use),   and the basic format,  which omits hyphens and colons.    I don't think you are allowed to have say,  a date in a basic format and then a time of day in the extended format,  or vice-versa, however.

How much of ISO 8601 do we want to support?

nurpax:
> I really do think that specifying an offset from UTC should be required for the UTCTime instance, and that aeson should stay firm on this requirement. (And yes, tacking on a Z at the end qualifies as specifying an offset, so that's fine.)

That is how to code works now.  The earlier aeson UTCTime decoder did not accept a +/-HH[:MM] offset, only Z was accepted.  Inputs without any tz offset are rejected.

> Ok, ISO 8601:2004 does explicitly require two digits in the month, day, time of day hour, minute, second, and offset hour and minute fields, though some of these fields may be omitted altogether

The proposed parser requires two digits for these.  The offset ':MM' part is optional so something like "2015-01-01T12:30:00.00-02" is also allowed.  

You've seen single digit hour offsets in some inputs, are you saying we should accept those or just go with the ISO 8601:2004?

The unit tests I added https://github.com/nurpax/aeson/commit/a044abf4329852a10276961bcfd553cf19924a43 list various valid and invalid inputs.  Need to amend those once we nail down what the parser should accept and what it shouldn't.

> Notably, however, commas may be used as well as periods, and for the time of day, you can specify fractional hours (but then cannot specify minutes or secconds), or you can specify fractional minutes (but then cannot specify fractional seconds.)

I'm Finnish and we're supposed to use a comma instead of period for the decimal point.  There is no shortage of trouble this quirk has caused.  However, I've never come across JSON data that'd use commas instead of periods as the decimal point (be it floating point or timestamps).  My vote would be -1 on allowing commas here.

I think I'd also be -1 on fractional hours and minutes.  Date&time is already hard to get right without pulling tricks like this. :)

lpsmith:
With regard to single digit hour offsets,  I'm inclined to stick with ISO 8601,  as really,  the source files for the tz database don't follow ISO 8601 in any way.   (But it did make me stop and think for a bit.)
 
To me,  the bigger question is whether or not we should go beyond ISO 8601 and support ±HH:MM:SS offsets,  (I'd give a moderate +1 for this),   whether or not to add a FromJSON instance for LocalTime (I'd give this a strong +1),  and whether or not to support the basic syntax without hypens and colons.   (I'd give this a weak -1)

lpsmith:
As for period versus comma,  I'm pretty neutral on that topic,  but it would be very easy to support reliably without major changes to the parser,  unlike fractional hours,  fractional minutes,  or the basic syntax.

lpsmith:
Ok,  looking at this code some more,  I noticed that the conversion to picoseconds isn't as fast as it could be.   (Actually,   now that I think about it,  I probably had that thought when writing it in the first place,  but just wanted to get something reasonable working.)

Anyway,  when I run your benchmark in a vanilla build of your fork:

~~~
$ ./dist/build/aeson-benchmark-dates/aeson-benchmark-dates
benchmarking utc-time
time                 47.19 aeson-benchmark-dates: <stdout>: commitBuffer: invalid argument (invalid character)
~~~

And when I run an optimized version I threw together:

~~~
$ ./dist/build/aeson-benchmark-dates/aeson-benchmark-dates
benchmarking utc-time
time                 43.34 aeson-benchmark-dates: <stdout>: commitBuffer: invalid argument (invalid character)
~~~

Not sure what the error message is about,  but it is being written on stderr,  not stdout.

Anyway,  my optimizations are at lpsmith/aeson@dcca0cc3f52b36223c7de71cf18b6feaca681503.

lpsmith:
Ok,  after setting my LANG variable,  here's the results of your pull request as is:

~~~
$ ./dist/build/aeson-benchmark-dates/aeson-benchmark-dates
benchmarking utc-time
time                 46.27 μs   (46.16 μs .. 46.41 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 46.24 μs   (46.19 μs .. 46.33 μs)
std dev              212.6 ns   (84.35 ns .. 375.5 ns)

benchmarking utc-time fract
time                 60.53 μs   (60.51 μs .. 60.56 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 60.67 μs   (60.56 μs .. 60.93 μs)
std dev              539.3 ns   (93.69 ns .. 919.6 ns)
~~~

With my optimized patch:

~~~
$ ./dist/build/aeson-benchmark-dates/aeson-benchmark-dates
benchmarking utc-time
time                 43.26 μs   (43.23 μs .. 43.29 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 43.34 μs   (43.29 μs .. 43.48 μs)
std dev              251.6 ns   (126.8 ns .. 431.4 ns)

benchmarking utc-time fract
time                 49.28 μs   (49.20 μs .. 49.39 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 49.21 μs   (49.19 μs .. 49.27 μs)
std dev              124.1 ns   (70.88 ns .. 230.5 ns)
~~~

nurpax:
Nice!

I noticed that Data.Attoparsec.Text has a `rational` which I wanted to use here.  However, it accepts  numbers with the exponent part (e.g. 1.0e1) which AFAICT is not allowed by ISO 8601 so it couldn't be directly used.

Once things settle down here, I can cherry-pick your change into this PR.

lpsmith:
Yeah, exponents are definitely not ISO 8601.    And unless someone's done it already in the last few years,  Data.Fixed is in desperate need of some TLC,  especially with regard to optimizations (some of the operations are implemented in ridiculously suboptimal ways),  and really the newtype constructor should be exposed for cases like this,  even if it's considered semi-internal.

bos:
This is very cool. I applied your change, but the original commit ID is lost for some reason – so you can see your commits in there, but the hashes are different.

I subsequently went in and did a nearby reimplementation to make the code both simpler and faster, and migrated it to [a module of its own](https://github.com/bos/aeson/blob/master/Data/Aeson/Parser/Time.hs).

### Performance

@nurpax 

```
benchmarking whole
time                 21.07 μs   (20.88 μs .. 21.25 μs)
allocated            88599 b    (88588 .. 88611)

benchmarking fractional
time                 28.92 μs   (28.66 μs .. 29.21 μs)
allocated           110285 b    (110269 .. 110304)
```

@bos

```
benchmarking whole
time                 15.78 μs   (15.66 μs .. 15.94 μs)
allocated            78950 b    (78941 .. 78961)

benchmarking fractional
time                 22.95 μs   (22.81 μs .. 23.10 μs)
allocated            98792 b    (98781 .. 98803)
```

It looks like my rewrite is about 25% faster, and uses 10% less memory.

### Future directions

There's some further work that needs to be done for sure, and here's what I currently think about the various possibilities:

* Having support for `LocalTime` and a few other types probably makes sense. Pull requests welcome :-)

* I threw away a bunch of really weird long-standing formatting options that frankly made no sense to me, and for which I've no idea why I accepted them in the first place. Stuff like a year, month, and time, *but no day*.

* I'm not really interested in supporting unused variants of ISO-8601, because adding code to support stuff that nobody needs adds clutter and slows the code down. If people show up with variants that are actually used in the wild, that's a different story.

lpsmith:
Would you be interested in the unsafeCoerce-based conversion to Pico?   It's still worth an additional 10-20% in my tests.

PR #269, opened 2015-07-20
Add ToJSON/FromJSON instances for LocalTime

lpsmith:
This provides a relatively convenient way of dealing with timestamps that don't specify an offset,  while emphasizing the correct ISO 8601 semantics of such a syntax.


bos:
Thanks! This is superseded by 684baea3976238acdbe416a362dd9e10e0d763bb, which is faster and has tests.

PR #270, opened 2015-07-21
Utc offset fixes

lpsmith:
The biggest problem here is that you have a sign error on the minutes component when the offset is negative,  which the test suite did not catch.   You may be interested in taking a look at [postgresql-simple's time test suite](https://github.com/lpsmith/postgresql-simple/blob/b4624db41048e42e993d3bea820e907a010fde00/test/Time.hs).

Also, as I understand ISO 8601:2004,  ±HHMM is not a valid syntax for specifying
a UTC offset in the context we are using it.   Only ±HH and ±HH:MM are
valid in the extended ISO 8601 syntax,   and ±HH is a commonly used variant.

I probably would weakly prefer to not accept ±HHMM, however,  because the
the ToJSON ZonedTime instance has been emitting this syntax since aeson
version 0.6.1.0 released on 2012-12-14,   our hands may well be tied on
this matter.

I can only hope that the older FromJSON instances will also accept ±HH and
±HH:MM,  so that we can fix the emitted syntax with minimal disruption,
but knowing the time library, this probably is not the case.

### DIFF
--- a/Data/Aeson/Parser/Time.hs
+++ b/Data/Aeson/Parser/Time.hs
@@ -113,17 +113,22 @@ timeZone = do
   if ch == 'Z'
     then return Nothing
     else do
-      h0 <- twoDigits
-      m <- maybeSkip ':' *> twoDigits
-      let h | ch == '-' = negate h0
-            | otherwise = h0
-      case h * 60 + m of
-        off | off == 0 ->
+      h  <- twoDigits
+      mc <- peekChar
+      m  <- case mc of
+              Just c | c == ':'  -> anyChar *> twoDigits
+                     | isDigit c -> twoDigits
+              _ -> return 0
+      case () of
+        _ | h == 0 && m == 0 ->
               return Nothing
-            | off < -720 || off > 840 || m > 59 ->
+          | h > 23 || m > 59 ->
               fail "invalid time zone offset"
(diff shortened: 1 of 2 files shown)


bos:
Thanks for spotting that bug. I took parts of your change, however I'm not thrilled with the `±HH` so I've omitted that. If you've seen it in the wild in JSON dates, please let me know.

I can't read ISO 8601 (thanks, ISO!), but at least Wikipedia suggests that `±HHMM` is fine.

bos:
Changed my mind about `±HH` since you mentioned somewhere that Postgres will emit it sometimes.

lpsmith:
±HHMM is most certainly valid ISO 8601 in other contexts.   For example,  you can write `2015-07-21T04:44:38-04:00`  (the extended syntax that most people use most of the time)  or `20150721T044438-0400`,  (the basic syntax that the standard itself discourages the use of in "plain text" contexts.)

But it seems that the standard doesn't technically allow the basic and extended formats to be mixed and matched.    Perhaps my interpretation is wrong,  as I can't point to a phrase that specifically prohibits it,  like the prohibition of a space between the time of day and utc offset.    But the overall context,  namely  the examples provided and the contextual use of language, seem to strongly suggest to me that it's not in accordance with the standards.

PR #271, opened 2015-07-21
Performance enhancement for string parser

SeanRBurton:
Finds the size of the string and checks whether it needs to be escaped in a single pass, rather than two as before.


bos:
Looks interesting.

I need a few changes from you, please:
* Drop the `.swp` file
* Squash into a single commit
* Show me what the performance impact is in the commit comment

bos:
Never mind, I've done the rebasing, squashing, and fixing-up myself :-)

SeanRBurton:
Thanks, I did some quick benchmarking and it doesn't come out ahead after
all, but I think I can rescue things.

bos:
Applied as 84ced1530c362646d42fb1f8ea347703859eedc4, thanks.

bos:
I asked for benchmark results because I expected that your rework would be slow, and the results reveal this to be the case. Here's what I did to close most of the gap: c24f9d37623b066d450358c4aa7b0535bd62b0b9

Unless I find another trick up my sleeve to get past parity, I'll be reverting both commits.

SeanRBurton:
One problem is that scan is written with a few performance tweaks which
aren't duplicated for runScanner.

bos:
I don't think those are all that relevant, but I wrote https://github.com/bos/attoparsec/commit/b1c6694447321db82e5a41d36653fcb8dd8d97c9 just to be safe.

I also have some GHC >= 7.8 unboxing tricks up my sleeve that look very promising. Stay tuned, your idea still has legs :-)

bos:
Performance numbers with GHC 7.10.1 on my MBP, using `CompareWithJSON`'s `decode/en/aeson/strict` result.

Baseline:

```
time                 1.671 ms   (1.656 ms .. 1.688 ms)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              4252843.762 (4251621.741 .. 4254161.104)
```

@SeanRBurton:

```
time                 2.054 ms   (2.011 ms .. 2.102 ms)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              7174189.077 (7172735.587 .. 7175725.079)
```

With e9623d0c61fd0f06e538a5491749e47d0d55a517:

```
time                 1.450 ms   (1.433 ms .. 1.470 ms)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              4332621.071 (4331339.889 .. 4333745.585)
```

Slightly increased memory allocation, but 10% faster than the original baseline. Sean, I believe this validates your approach, it just needed a little GHC-specific TLC to close the gap. Thanks for thinking of the original patch!

SeanRBurton:
Also note that you can use a three  constructor enum rather than a pair of
bools.

PR #276, opened 2015-07-23
Add ToJSON and FromJSON instances for NominalDiffTime

abbradar:
This serializes NominalDiffTime as a number of seconds, which sounds like a sensible encoding for a "length of time" thing to me.

I have a feeling that this instance isn't present due to some reason other than simply being not implemented, but my search of any discussions resulted in nothing -- sorry if that's indeed the case.


bos:
Nice work, thanks.

PR #278, opened 2015-07-27
Support deriving data families with Template Haskell

RyanGlScott:
`Data.Aeson.TH` currently only supports plain data types. This pull request extends the Template Haskell engine to allow deriving data family instances.

Since it is impractical to pass an entire data instance's type (e.g., `Family Int a`) to Template Haskell, this works by passing in the name of a data family instance constuctor to `deriveJSON`/`mk-`/etc., which infers the instance it corresponds to:

```haskell
data family DF a
data instance DF Int = DF1 Int
                     | DF2 Int Int
                     deriving Eq

$(deriveJSON defaultOptions 'DF1)
-- Alternatively, one could pass 'DF2 instead
```


bos:
cc @basvandijk

bos:
Looks good, thanks!

bos:
Nice work, thanks!

PR #280, opened 2015-08-24
make the key show up in the error message

gregwebs:
I have added this as an override to a couple projects already. We get a lot of reports from users confused about why their JSON/YAML did not parse because it doesn't tell them the key that failed.

Feel free to modify the format of the error message to your liking. The important thing is just to have the key in there somehow.


gregwebs:
ping. I would really like to get this in the next release because it trips up everyone.

bos:
Thanks!

PR #282, opened 2015-09-13
Fix GHC HEAD build

RyanGlScott:
The `template-haskell` API has changed considerably in version 2.11.0.0 due to the introduction of injective type families. Barring any other major changes, this patch should be sufficient to make `aeson` build on GHC HEAD.


bos:
Thanks.

PR #283, opened 2015-09-17
Encode Timestamp subseconds with all available digits

### LINKED ISSUE #274: Precision and rounding when encoding fractional seconds
Why perform any rounding or truncation of fractional seconds at all?   If you simply send all available digits,  then I can truncate or round timestamps myself before I send them to your encoder.    But if you do it for me,  and I want to send more precision than your encoder supports,  then I have to bypass your encoder altogether.

lpsmith:
Would resolve #274

Benchmarks don't appear to be much different before or after.   Unfortunately I can't really post the numbers since I accidentally closed one of the terminal windows with the numbers and my laptop just now stopped being able to work off of AC power,  battery only,  which means very unreliable numbers.


bos:
Nice, this actually speeds things up a little too.

PR #284, opened 2015-09-18
Add throwDecode family of decoding functions

phadej:
I find myself often reimplementing those to use with web apis:

```hs
req <- makeTheRequest
res <- httpLbs req mgr
throwDecode (responseBody res)
```


bos:
Thanks for the patch. Unfortunately, I'm not going to accept it as it adds too many mostly redundant entry points that save very little work.

PR #286, opened 2015-09-19
Added test modules to 'other-modules' section

dmjio:
`cabal unpack aeson` had some missing modules which caused the test project to be unable to build.


bos:
Applied, thanks.

Issue #287, opened 2015-09-19
Breaking changes on 0.10 regarding Null field parsing on Maybe types


dmjio:
```haskell
-- 0.9
λ: newtype A = A { val :: Maybe String } deriving Show
(0.01 secs, 3,077,712 bytes)
λ: instance FromJSON A where parseJSON (Object o) = A <$> o .:? "val"
(0.00 secs, 1,032,488 bytes)
λ: eitherDecode "{ \"val\" : null }" :: Either String A
Right (A {val = Nothing})
(0.00 secs, 1,552,616 bytes)
-- 0.10
λ> newtype A = A { val :: Maybe String } deriving Show                                                                                                                            
(0.01 secs, 3,086,560 bytes)                                                                                                                                                      
λ>  instance FromJSON A where parseJSON (Object o) = A <$> o .:? "val"                                                                                                                                                                                                                                                                       
(0.01 secs, 2,069,808 bytes)                                                                                                                                                      
λ> eitherDecode "{ \"val\" : null }" :: Either String A                                                                                                                           
Left "Error in $.val: failed to parse field val: expected String, encountered Null"                                                                                               
(0.01 secs, 2,622,128 bytes)        
```

Is this intended behavior? Normally `Maybe` types would return `Nothing` in the presence of a `Null` field (Under 0.9).

This change causes the `stripe-haskell` project to fail 95% of tests. :sob:

dmjio:
Seems like this is the culprit, https://github.com/bos/aeson/commit/d0414be92ee6bf5b4c057978955b89d111767dab

Can we keep the existing operator `(.:?)` with `0.9.0` behavior, and as @gregwebs recommends, add a `(.:??)` operator for this new behavior ?

lykahb:
You use a wrong operator. Field val has type `Maybe String`. In the return value of the `(.:?)` `Maybe` is related only to the presence of the key. The operator has never not handled nulls. The key exists so it tries to convert null into `String`.
If the key val is mandatory, you can use `(.:)`. If it is optional, `fmap join (o .:? "val")` will use the right FromJSON instance.

dmjio:
@lykahb I see, so previously `fmap join (o .:? "val")` was the default behavior (`Null` would be coerced into `Nothing`). `Null` and key existence were conflated. Now it's separate. Cool.

> Maybe is related only to the presence of the key

In `0.10` yes, in `0.9` it could be related to the value too, right? Example below.
```haskell
[...]

PR #288, opened 2015-09-19
Add (.:!) and partially revert d0414be92ee6bf5b4c057978955b89d111767dab

phadej:
Handle #83 by introducing `.:!` as @gregwebs proposed (I bikeshedded `.:??` to `.:!`). Fixes undocumented breaking change in `0.10`

Related: https://github.com/bos/aeson/issues/287

### DIFF
--- a/Data/Aeson.hs
+++ b/Data/Aeson.hs
@@ -78,6 +78,7 @@ module Data.Aeson
     , foldable
     , (.:)
     , (.:?)
+    , (.:!)
     , (.!=)
     , object
     -- * Parsing
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -101,7 +101,7 @@ import Data.Aeson.Types ( Value(..), Parser
                         , defaultTaggedObject
                         )
 import Data.Aeson.Types.Internal (Encoding(..))
-import Control.Monad       ( join, liftM2, return, mapM, fail )
+import Control.Monad       ( liftM2, return, mapM, fail )
 import Data.Bool           ( Bool(False, True), otherwise, (&&), not )
 import Data.Either         ( Either(Left, Right) )
 import Data.Eq             ( (==) )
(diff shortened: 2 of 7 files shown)


dmjio:
+1

gregwebs:
great, this needs to be merged. only problem is that we have a bunch of operators with slightly different semantics. A big improvement would be to give a non-operator name to all of them. And better yet would be to make combinators to describe how maybe is treated.

```
undefinedAsNull "maybeField"
```

phadej:
Probably shouldn't be part of this PR, but I'd like to have names too.

How about

```hs
parseField = (.:)
parseOptionalField = (.:?)
parseNotNullOptionalField = (.:!) -- this one is huge :(
withDefault = (.!=)
```

gregwebs:
Can we get this released now as version 0.11 to avoid breaking everyone with 0.10 ?

k-bx:
@phadej this needs to be rebased

phadej:
Rebased (and fixed the build `sdist`): https://travis-ci.org/phadej/aeson/builds/85721736

Please, setup travis for all of us!

nikomi:
+1!

This was almost 2 months ago - is anything happening with this?

bos:
Applied, thanks.

gregwebs:
@phadej does it make sense to apply this now that 0.10 has been released or is this a breaking change back in the other direction?

phadej:
@gregwebs, if the next release is `0.11`...

gregwebs:
But what is the point of making a 0.11 release that makes things like 0.9 if 0.10 is out? Isn't this merge 3 months late?

phadej:
I cannot really comment on that, as I'm author and user of `aeson-compat` which makes `(.:?)` in 0.10 behave like it was in 0.9. IMO 0.9 behaviour is more useful, your mileage may vary.

gregwebs:
This is obviously better and it would have been great to be merged before the 0.10 release. However, many people changed their code to work with 0.10. Those that did not take a compatible approach will just need to change it back to 0.9 behavior now.

tolysz:
I am just testing if this PR actually fixes my `Maybe ~ Null` case; otherwise my patch makes `aeson` behave correctly (i.e. pre 10 )

codygman:
Should this close https://github.com/bos/aeson/issues/287?

wereHamster:
@gregwebs there are also many people who haven't upgraded their package yet (due to this and other problems with the 0.10 release), for them keeping the 0.9 semantics would be preferable.

nikomi:
+1

I assumed the null-handling API change to be an unintended glitch that would be corrected in the next version, probably marking 0.10.0.0 as deprecated.

PR #299, opened 2015-10-05
Fix time encoding bugs

### LINKED ISSUE #298: Serialized times are not quoted
> x <- getCurrentTime
    > encode x
    "2015-10-05T00:15:27.673442Z"
    > encode [x]
    "[2015-10-05T00:15:27.673442Z]"
    > decode (encode [x])
    Nothing

Serializing times results in naked time info which isn't valid json. They should be quoted in the encoded output.

benweitzman:
Time datatype were being encoded without surround quotation marks, which did not yield valid JSON. This fixes the bug, and also fixes the tests, which were not catching this bug.

This fixes #298

### DIFF
--- a/Data/Aeson/Encode/Builder.hs
+++ b/Data/Aeson/Encode/Builder.hs
@@ -23,6 +23,7 @@ module Data.Aeson.Encode.Builder
     , text
     , string
     , unquoted
+    , quote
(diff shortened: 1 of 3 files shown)


basvandijk:
Thanks for the fix!

PR #302, opened 2015-10-13
Parse HH:MM time in zonedTime and utcTime

timbodeit:
Some APIs around the Web give date information in `YYYY-MM-DDTHH:MM±hh:mm` format.  
Example: The [QPX Express API](https://developers.google.com/qpx-express/?hl=en) by Google.

As `HH:MM` is a valid time format in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), I find this to be a valid format that should be accepted by aeson.

Changes:
- Make parsing the seconds optional in `timeOfDay`
- Adjust documentation indicating the allowed format
- Add corresponding tests

### DIFF
--- a/Data/Aeson/Parser/Time.hs
+++ b/Data/Aeson/Parser/Time.hs
@@ -62,12 +62,12 @@ twoDigits = do
   let c2d c = ord c .&. 15
   return $! c2d a * 10 + c2d b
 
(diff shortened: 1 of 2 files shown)


dsmatter:
:+1:

basvandijk:
Good stuff. Thanks!

PR #303, opened 2015-10-15
add instances/tests for TimeOfDay, make test for LocalTime actually arbitrary

benweitzman:
This changes adds FromJSON/ToJSON instances for TimeOfDay (the parser/builder is already present since TimeOfDay appears as components of other time datatypes with instances).

Also, the arbitrary instance for local time was not arbitrary at all, this fixes that.

### DIFF
--- a/tests/Instances.hs
+++ b/tests/Instances.hs
@@ -6,15 +6,16 @@ module Instances where
 import Types
 import Data.Function (on)
 import Control.Monad
-import Test.QuickCheck (Arbitrary(..), Gen, choose, oneof, elements)
+import Test.QuickCheck (Arbitrary(..), Gen, choose, oneof, elements, Positive(..))
 import Data.Time.Clock (DiffTime, UTCTime(..), picosecondsToDiffTime)
 import Data.Fixed (Pico)
 import Data.Time (ZonedTime(..), LocalTime(..), TimeZone(..),
                   hoursToTimeZone, Day(..), TimeOfDay(..),
-                  NominalDiffTime)
+                  NominalDiffTime, makeTimeOfDayValid)
 import qualified Data.Text as T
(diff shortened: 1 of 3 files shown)


agrafix:
:+1: , I'm looking forward to the merge, thanks for this!

tests/Instances.hs
```diff
@@ -27,8 +28,15 @@ instance Arbitrary Text where
 instance (Ord k, Arbitrary k, Arbitrary v) => Arbitrary (Map.Map k v) where
     arbitrary = Map.fromList <$> arbitrary
 
+instance Arbitrary TimeOfDay where
+    arbitrary = do
+      Positive h <- arbitrary
+      Positive m <- arbitrary
+      Positive s <- arbitrary
+      return $ fromMaybe (TimeOfDay 0 0 0 ) (makeTimeOfDayValid h m s)
+
```
basvandijk: Hi @benweitzman , thanks for the PR and sorry for the long delay in reviewing this.

I think the following Arbitrary instance for TimeOfDay is a bit better since it doesn't generate the default `TimeOfDay 0 0 0` when the generated TimeOfDay is invalid:
  
```haskell
instance Arbitrary TimeOfDay where
  arbitrary = do
    h <- choose (0,23)
    m <- choose (0,59)
    s <- choose (0,59)
    return $ TimeOfDay h m s
```

If you agree, could you change that and update your PR so I can merge it.

benweitzman:
@basvandijk Your suggested is in, btw.

bos:
Thanks for the contribution. I made various cleanups afterwards, as https://github.com/bos/aeson/commit/4429b86c2d7bae8e57b8bfa0c0402c64a9a6e26e.

Issue #300, opened 2015-10-06
TaggedObject encoding for Nullary constructor includes "contents"


devwout:
Suppose I want to encode the following type as JSON using TaggedObject SumEncoding.

```haskell
data MyType = Zero | One String
     deriving (Generic)
instance Aeson.ToJSON MyType
```

The output of converting `One "bla"` to JSON is `{ "tag": "One", "contents": ["bla"] }` , which is sensible.

However, when converting `Zero` to JSON, you get `{ "tag": "Zero", "contents": [] }`. While technically correct, the "contents" key is completely unnecessary. Worse even, the "contents" key is required for the type to be parsed again.

This is awkward if one wants to use the JSON format as an external interface. Of course, it is possible to manually implement the FromJSON and ToJSON instances for that type, but that is tedious work. 

I cannot imagine a case where one would like the empty "contents" key to be there. Is this intended behavior? Is there something I am missing?

devwout:
Below is a patch that changes the behavior of the generic implementation for Nullary constructors in the TaggedObject SumEncoding. Referring to my example above, it should encode `Zero` as `{ "tag": "Zero" }`.

Does this make sense? I am pretty new to Haskell, so sorry if it doesn't.

```diff
diff --git a/Data/Aeson/Types/Generic.hs b/Data/Aeson/Types/Generic.hs
index 965fe30..669ee1b 100644
--- a/Data/Aeson/Types/Generic.hs
+++ b/Data/Aeson/Types/Generic.hs
@@ -344,6 +344,13 @@ instance (Selector s, ToJSON a) => RecordTo (S1 s (K1 i (Maybe a))) where
     recordToEncoding opts m1 = fieldToEncoding opts m1
     {-# INLINE recordToEncoding #-}
 
+instance RecordTo U1 where
+    recordToPairs _ U1 = empty
+    {-# INLINE recordToPairs #-}
+
+    recordToEncoding _ U1 = mempty
+    {-# INLINE recordToEncoding #-}
+
 fieldToPair :: (Selector s, GToJSON a) => Options -> S1 s a p -> DList Pair
 fieldToPair opts m1 = pure ( pack $ fieldLabelModifier opts $ selName m1
                            , gToJSON opts (unM1 m1)
@@ -663,6 +670,10 @@ instance (Selector s, FromJSON a) => FromRecord (S1 s (K1 i (Maybe a))) where
                     selName (undefined :: t s (K1 i (Maybe a)) p)
[...]

PR #307, opened 2015-10-20
TaggedObject encoding for Nullary constructor does not have "contents" field

devwout:
As requested in #300 , implemented this for both GHC Generics and TemplateHaskell .

### DIFF
--- a/Data/Aeson/Types/Generic.hs
+++ b/Data/Aeson/Types/Generic.hs
@@ -206,6 +206,9 @@ instance ( IsRecord           a isRecord
 class TaggedObjectPairs' f isRecord where
     taggedObjectPairs' :: Options -> String -> f a -> Tagged isRecord [Pair]
 
+instance OVERLAPPING_ TaggedObjectPairs' U1 False where
+    taggedObjectPairs' _ _ _ = Tagged []
+
 instance (RecordToPairs f) => TaggedObjectPairs' f True where
     taggedObjectPairs' opts _ = Tagged . toList . recordToPairs opts
 
@@ -233,7 +236,6 @@ instance ( IsRecord         a isRecord
         (builder tagFieldName <>
          B.char7 ':' <>
          builder (constructorTagModifier opts (conName (undefined :: t c a p)))) <>
-        B.char7 ',' <>
         ((unTagged :: Tagged isRecord B.Builder -> B.Builder) .
          taggedObjectEnc' opts contentsFieldName . unM1 $ v) <>
         B.char7 '}'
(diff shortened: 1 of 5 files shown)


Data/Aeson/Types/Generic.hs
```diff
@@ -201,7 +201,6 @@ instance ( IsRecord      a isRecord
         (builder tagFieldName <>
          B.char7 ':' <>
          builder (constructorTagModifier opts (conName (undefined :: t c a p)))) <>
-        B.char7 ',' <>
```
bos: This is a bit mysterious. Can you explain what's going on here?

bos: I think I see what's going on: you moved the comma insertion elsewhere. Is that correct?

(Ideally, the movement of comma insertion would have been a separate standalone commit, and it would thus have been obvious what was going on.)

devwout: Yes indeed, I moved the comma insertion. Otherwise, the encoding for nullary constructor would be `{"tag": "X",}` (trailing comma). I agree a separate commit would have made this more clear, I'll keep it in mind.

Data/Aeson/Types/Generic.hs
```diff
@@ -211,11 +210,18 @@ class TaggedObject' f isRecord where
     taggedObjectPairs' :: Options -> String -> f a -> Tagged isRecord [Pair]
     taggedObjectEnc' :: Options -> String -> f a -> Tagged isRecord B.Builder
 
+instance OVERLAPPING_ TaggedObject' U1 True where
```
basvandijk: Hi @devwout, I think it would be better not to treat U1 as a record. I think the following is simpler.

Encoding:

```haskell
instance OVERLAPPING_ TaggedObject' U1 False where
    taggedObjectPairs' _opts _contentsFieldName =
        Tagged . const []
    {-# INLINE taggedObjectPairs' #-}

    taggedObjectEnc' _opts _contentsFieldName =
        Tagged . const mempty
    {-# INLINE taggedObjectEnc' #-}
```

Decoding:

```haskell
instance OVERLAPPING_ FromTaggedObject'' U1 False where
    parseFromTaggedObject'' _opts _contentsFieldName = Tagged . const (pure U1)
    {-# INLINE parseFromTaggedObject'' #-}
```

Could you try if that works and update your PR?

devwout:
Hi @basvandijk 

I updated my pull request based on your suggestions and it still works as expected. Explicitly specifying the case for nullary constructors is indeed simpler.

If you prefer me to amend these two commits to the previous one, tell me. We are all busy, I understand this took a while.

bos:
@devwout if you could please squash and rebase your commits, I'd appreciate it. Thank you.

devwout:
Hi @bos 

The pull request is rebased and squashed into two commits. I think it is best to commit the tests for existing functionality separately.

PR #308, opened 2015-10-20
ToJSON and FromJSON instances for Version type

Daniel-Diaz:
This can be useful, for example, to interface with the Hackage API.

### DIFF
--- a/Data/Aeson/Types/Instances.hs
+++ b/Data/Aeson/Types/Instances.hs
@@ -1472,6 +1472,9 @@ instance ToJSON Version where
     toJSON = toJSON . showVersion
     {-# INLINE toJSON #-}
 
+    toEncoding = toEncoding . showVersion
+    {-# INLINE toEncoding #-}
+
 instance FromJSON Version where
     {-# INLINE parseJSON #-}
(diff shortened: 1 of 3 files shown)


basvandijk:
Hi Daniel, thanks for the pull request! I agree this is useful to have. A few points:

* Can you also define an explicit `toEncoding`. This will be more efficient than the default which goes via `toJSON`.

* It would be nice if you can also add a few test cases for encoding and decoding `Versions`.

* I'm wondering what to do about whitespace after a version like `"1.2.3  "`. Should this parse to a `Version`? I'm inclined to say no. What do you think? If no, we should change `go` to:

```haskell
        go [(v,"")] = return v
        go (_ : xs) = go xs
        go _        = fail $ "could not parse Version"
```

Daniel-Diaz:
@basvandijk Thanks for reviewing the PR.

* Explicit definition of ``toEncoding`` added.

* I have also added a test using ``roundTripEq``. I hope this is OK.

* You are right, the parser was not only accepting trailing whitespaces, but any characters after the longest version possible has been parsed. I applied the idea you proposed.

basvandijk:
@Daniel-Diaz thanks for the updates and sorry about the delay in merging this.

PR #316, opened 2015-11-25
Add instances of ToJSON and FromJSON for Ordering

cgibbard:


Data/Aeson/Types/Instances.hs
```diff
@@ -182,6 +182,21 @@ instance FromJSON Bool where
...
 
+instance ToJSON Ordering where
+  toJSON o = String $ case o of
+    LT -> "LT"
+    EQ -> "EQ"
+    GT -> "GT"
+
+instance FromJSON Ordering where
+  parseJSON v = case v of
+    String s -> case s of
```
basvandijk: Could you replace the explicit case analysis on `v`with [withText](http://hackage.haskell.org/package/aeson-0.10.0.0/docs/Data-Aeson-Types.html#v:withText). That's a bit shorter and consistent with other text-based encodings.

cgibbard:
There you go

basvandijk:
Thanks. I made a small performance [tweak](91ec3be856b2457b88afd5ca56d634ae1f71b03b) and merged your commits.

ryantrinkle:
@basvandijk Do you think we can get this on hackage soon?

Issue #323, opened 2015-12-12
Extend Value type. 


tolysz:
Hi,

I know we had this discussion a couple of times already; but I really try to use `aeson`

    data Series = Empty
            | Value Encoding
            deriving (Typeable)

I see that `Empty` is only used for arrays; but I would like to have some kind of indication at the `Value` level that the result should be serialised as `Empty` in arrays and as omitted key in  `Object`s
(not just once `omitNothingFields` possibly different for each)

My proposal requires adding an Extra constructor to `Value` one which will represent removed/missing/one_to_be_removed key.

This extra constructor is needed if we want to be able to build more generic types.

say we have and `instance ToJSON a => Map Text a`
There is no way for `a` to tell `aeson` how it should serialise itself (i.e. `omitNothingFields` is not modifiable)  

My proposal is to add `Missing` constructor to `Value`  which will serialise same currently `Null` with `omitNothingFields=True` and the default for `Null` be `omitNothingFields=False`. 

This requires `.:??` extra operator which would return `Missing` (or `Omitted`) if there is no field and `Null` if there is `null`
and the `.:?` will behave as it currently does.

There are practical cases where it is useful; but you know that as you have `omitNothingFields`; but it not allow for the field to be `null` or `Omitted` depending on the actual value.

e.g.
   
     With a type `Maybe (Maybe a)` (Or any other type with the same universal property) I want to have:

    instance ToJSON a => Maybe (Maybe a)
    toJson Nothing = Omitted
      toJson (Just (Nothing)) = Null
      toJson (Just (Just a)) = toJson a

(I use it extensively https://github.com/tolysz/google-api ) I use patched `aeson-8.0.0.2` but I would like to port it upstream it compiles all libraries. 

Modification should be quite straightforward:
Add extra constructor; update encoders/decoders; update TH and Generics...

phadej:
I don't think that adding a constructor to `Value` is good idea. There are various problems in your request, you try to solve with it:

### Distinguishing between missing and `Null` value:

with next release, there is `(.:!) :: FromJSON a => Object -> Text -> Parser (Maybe a)` which you can use to parse field into `Maybe (Maybe b)`. The `Nothing` will be missing value, `Just Nothing` would be `Null` and `Just (Just x)` present value.

### `instance ToJSON a => Map Text a`

Omitting some values from `Map` would break `toJSON . fromJSON` identity. I personally don't like that,

### Records

[...]

PR #324, opened 2015-12-12
Sample implementation for #323

tolysz:
Hi,

I am not sure if there is a shorter way to address this extension; I had to add an extra `basic` type next to `Maybe` so I can capture the `Omitted` bits.

I tried to keep the new `Omitted` values as much internally as possible; so it should not even show in generated  `toJSON`.

The library should be functionally exactly the same modulo when using `Possible` rather then `Maybe` in constructors. List of `[Possible]` will serialize in a funny way but someone might actually want it this way (One can always use lists of `[Maybe]` there...

### DIFF
--- a/Data/Aeson.hs
+++ b/Data/Aeson.hs
@@ -79,6 +79,7 @@ module Data.Aeson
     , foldable
     , (.:)
     , (.:?)
+    , (.:??)
     , (.:!)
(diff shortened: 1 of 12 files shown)


tolysz:
rebased

bergmark:
See discussion in #323

Issue #325, opened 2015-12-12
Failed to parse field some-string: expected Text, encountered Null


tolysz:
I am testing the git.master version and got this error:

    Object (fromList [("some-string",Null)])
    aeson-example-simple: failed to parse field some-string: expected Text, encountered Null

I would expect that `fromJSON` would coerce `Null` to `Nothing :: Maybe Text` but it sees `Null` as a `Type`  ( I tested it on both GHC 7.10.2 and 7.10.3 )

Which is a result of the following code:

    {-# LANGUAGE RecordWildCards #-}
    {-# LANGUAGE OverloadedStrings #-}

    import Data.Aeson
    import Data.Text

    data Some = Some {someString :: Maybe Text}
      deriving Show

    instance FromJSON Some where
      parseJSON = withObject "Some" $ \o -> do
            someString     <- o .:?  "some-string"
            return Some{..}

    a :: Value
    a = object ([("some-string",Null)])

    main = do
       print a
       case fromJSON $ a of
           Error e -> error e
           Success some -> print (some :: Some)

and in `.cabal`

    executable aeson-example-simple
      main-is: Simple.hs
      ghc-options: -Wall
      build-depends:
        aeson,
        base,
        bytestring,
        text

I will investigate it a bit longer but if you have anything to suggest It would be awesome...

bergmark:
This seems to have been fixed. Currently on master:
```
λ> main
Object (fromList [("some-string",Null)])
Some {someString = Nothing}
```

PR #326, opened 2015-12-12
Add missing null handling:

tolysz:
Simple solution; adding just one extra pattern match; almost by definition `Null` maps to Nothing in `Maybe a` situation and only the rest is exposed for further parsing.
resolves: https://github.com/bos/aeson/issues/325


phadej:
https://github.com/bos/aeson/pull/288 related

tolysz:
I re-based it.

phadej:
AFAICS #288 already *fixed* the issue?

bergmark:
Right, I tested #325 and confirmed that this is fixed. Thank you for the pull request and sorry for not merging it.

PR #337, opened 2016-01-21
Define Semigroup instances to silence -Wcompat

hvr:
Since GHC 8.0/base-4.9, Data.Semigroup is available in `base-4.9`, so we
can now finally express that something is (also) a Semigroup.

In order to avoid CPP, this PR adds a conditional dependency on the
`semigroups` package for GHCs prior to 8.0

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -62,7 +62,7 @@ import Data.Char (isLower, isUpper, toLower)
 import Data.Data (Data)
 import Data.HashMap.Strict (HashMap)
 import Data.Hashable (Hashable(..))
-import Data.Monoid ((<>))
+import Data.Semigroup (Semigroup((<>)))
 import Data.Scientific (Scientific)
 import Data.String (IsString(..))
 import Data.Text (Text, pack, unpack)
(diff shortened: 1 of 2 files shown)


RyanGlScott:
While we're at it, should we add `MonadFail` instances for [`IResult`](https://github.com/bos/aeson/blob/63a7a4347d5cb7ffcaa28e165de509fb3163e8cf/Data/Aeson/Types/Internal.hs#L124), [`Result`](https://github.com/bos/aeson/blob/63a7a4347d5cb7ffcaa28e165de509fb3163e8cf/Data/Aeson/Types/Internal.hs#L135), and [`Parser`](https://github.com/bos/aeson/blob/63a7a4347d5cb7ffcaa28e165de509fb3163e8cf/Data/Aeson/Types/Internal.hs#L238) to complete the GHC 8.0 experience?

hvr:
I'm working on that too... expect a PR rsn!

RyanGlScott:
Ah, you're way ahead of me :)

I guess it's worth noting that since `Semigroup` isn't a superclass of `Monoid` yet (and therefore `mappend` != `(<>)`), this PR would break `aeson` code that relies on `mappend`. But I suppose it's up to the `aeson` maintainers to determine if that's OK.

hvr:
@RyanGlScott that's actually a good point; I'm actually not totally sure what the intention is for those `Monoid` instances. If they are sensible, we can easily re-add `Monoid` instances.

bos:
Am I correct in reading that you've deleted the `Monoid` instances? Please don't :-) I've already had enough fun accidentally changing semantics on people; I don't want to break their builds entirely for no particular reason.

phadej:
maybe just `{-# DEPRECATED #-}` instances would be ok for some time?

hvr:
@phadej Does GHC support deprecating instances?

hvr:
@bos Fair enough, I'll update this PR soon. At least the underlying types have a way to express `mempty` without bottoms, so they do have some level of justification at least :-)

hvr:
@bos I've updated the PR to retain all previously existing `Monoid` instances

PS: It would be great if you could flip on the switch at https://travis-ci.org/bos for the `aeson` repo, so that the `.travis.yml` file that's checked into the Git repo would help to pre-validate pull-requests...

bos:
Applied, thanks. I turned on Travis, will turn it back off again if it starts mailing me :-)

hvr:
@bos see #343 :-)

PR #345, opened 2016-01-26
Fix breakage by a65be52b36300dcf7f17273c303903b795581037

hvr:
Explicitly enabling Haskell2010 instead of GHC's default mode
enables a stricter layout rule...


bos:
:-)

PR #347, opened 2016-01-28
Add Lift Value instance

phadej:
I'm not sure in which module this should belong though.

### DIFF
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -1,6 +1,13 @@
 {-# LANGUAGE CPP, FlexibleInstances, NamedFieldPuns,
-    NoImplicitPrelude, TemplateHaskell,
-    UndecidableInstances #-}
+    NoImplicitPrelude, UndecidableInstances #-}
+#if __GLASGOW_HASKELL >= 800
+-- a) THQ works on cross-compilers and unregisterised GHCs
+-- b) may make compilation faster as no dynamic loading is ever needed (not sure about this)
+-- c) removes one hindrance to have code inferred as SafeHaskell safe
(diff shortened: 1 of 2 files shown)


phadej:
@hvr, moved it to `Value` definition module.

RyanGlScott:
Indeed, quoting makes `aeson` unbuildable on stage-1 compilers, which is kind of unfortunate. Luckily, you can work around this limitation by coming up with the `Name`s yourself. Here's an example of how it's done in `lens`: https://github.com/ekmett/lens/blob/ed3266ec4aaa4720cd0dea5957d383f750cf5dc8/src/Control/Lens/Internal/TH.hs#L95

RyanGlScott:
Actually, never mind. `Data.Aeson.TH` is already using quotes! In that case, my objection is invalid.

(Futhermore, I wasn't aware of `TemplateHaskellQuotes`, so that means you _will_ be able to build `aeson` on stage-1 compilers starting with GHC 8.0, which is super-cool. Thanks for implementing that extension, @hvr!)

bergmark:
:+1:

hvr:
@RyanGlScott it's actually @ezyang who made this possible (see ghc/ghc@f16ddcee0c64a92ab911a7841a8cf64e3ac671fd), but sadly was talked out of making this a separate pragma-token. However, I saw the need for having a explicit way to request a subset of the full `TemplateHaskell` (see ghc/ghc@85fcd035f73679927a0539d5f6c9b919517365e1) as this allows us to be more accurate in specifying language-features needed, and has a few benefits (with the downsides that we need to update packages to make use of this, rather than having them magically work w/o any adaptation)

PR #359, opened 2016-02-07
Track JSONPath in the FromJSON instance for Seq

### LINKED ISSUE #358: JSONPath: The instance for Seq a doesn't track index
~~~
> eitherDecode "[1,2,true]" :: Either String (Seq Int)
Left "Error in $: expected Int, encountered Boolean"

> eitherDecode "[1,2,true]" :: Either String [Int]
Left "Error in $[2]: expected Int, encountered Boolean"
~~~

(I think this is the last incorrect instance; sorry for the noise!)

neongreen:
Fixes #358


bergmark:
Cool :+1:

PR #360, opened 2016-02-08
Track JSONPath in the FromJSON instance for Either

### LINKED ISSUE #356: When parsing Either, JSONPath of the error appears to be incomplete
~~~ haskell
> eitherDecode "[1,2,{\"Left\":[3,4]}]" :: Either String (Int, Int, Either (Int, Bool) Bool)
Left "Error in $[2][1]: expected Bool, encountered Number"
~~~

It should be `$[2].Left[1]`, I think.

neongreen:
Fixes #356


bergmark:
Nice, thank you!

PR #363, opened 2016-02-09
Use nats package to provided Natural instances with old bases

phadej:


phadej:
Depends on https://github.com/nick8325/quickcheck/pull/76

aeson.cabal
```diff
@@ -110,6 +110,10 @@ library
     -- `Data.Semigroup` is available in base only since GHC 8.0
     build-depends: semigroups >= 0.16.1
 
+  if !impl(ghc >= 7.10)
+    -- `Numeric.Natural` is available in base only since GHC 7.8
+    build-depends: nats>=1
```
hvr: why no upper bound? Any major version bumps in `nats` from now on will be due to non-backward compatible changes. If there a reason we can leave off the upper bound it should be described in a  comment

aeson.cabal
```diff
@@ -110,6 +110,10 @@ library
     -- `Data.Semigroup` is available in base only since GHC 8.0
     build-depends: semigroups >= 0.16.1
 
+  if !impl(ghc >= 7.10)
+    -- `Numeric.Natural` is available in base only since GHC 7.8
+    build-depends: nats >=1 && <1.1
+
```
bergmark: * Isn't it simpler to say `impl(ghc < 7.10)`?
* The comment says "since GHC 7.8" but the conditional applies to `< 7.10`, is there a typo?

hvr: `impl(ghc < 7.10)` is subtly different from `!impl(ghc >= 7.10)` in its semantics

one implies `impl(ghc)` the other one does not

phadej: Fixed comment.

PR #364, opened 2016-02-09
Add NonEmpty instances

phadej:


tests/Instances.hs
```diff
@@ -136,6 +137,9 @@ instance (ApproxEq a) => ApproxEq [a] where
 instance Arbitrary Version where
     arbitrary = makeVersion . fmap getNonNegative <$> resize 4 (listOf1 arbitrary)
 
+instance Arbitrary a => Arbitrary (NonEmpty a) where
```
phadej: I restricted `QuickCheck` to `<2.8.3` because of this.
I'll make a PR to `QuickCheck` to include this and `Version` instance (though different)

bergmark:
Can you please add some tests that specify the format as well? Most of the roundtrip tests don't have a corresponding test verifying that the format doesn't silently change. See https://github.com/bos/aeson/blob/master/tests/UnitTests.hs#L216

bitemyapp:
@bergmark this has the NonEmpty unit tests now, can this go forward? I need this in like 4 different projects ^_^

phadej:
@bergmark I could fixup `nats` and ´tagged` PRs to be included in the next release (0.11.1.0 I guess), with the changes being only added instances, it would be quite conservative?

bitemyapp:
I personally just need the NonEmpty instance so that I don't have to newtype in my libraries. I'm trying to get Bloodhound back in Stackage and would like to move on this PR and a release now with any further things coming in another later release so it doesn't hold back what should be an obvious change.

bergmark:
@phadej that'd be great, thanks! I added a 0.11.1 milestone, is it this PR and #363 only?

aeson.cabal
```diff
@@ -162,6 +162,9 @@ test-suite tests
     vector,
     quickcheck-instances >=0.3.12
 
+  if !impl(ghc >= 8.0)
+    build-depends: semigroups >= 0.16.1
```
hvr: why no upper bound? maybe a comment would useful to explain why an upper bound may not be needed

bitemyapp: @hvr is this a roundabout way of asking for an upper bound? If so, I'd happily add one and open a new PR so this can get in soon.

phadej: @hvr https://github.com/bos/aeson/commit/2027fb84c60c08af70fbed1da7f88ee9d9a9cfb8 You are the best one to answer? :)

phadej:
Added upperbound

phadej:
rebased

bitemyapp:
@bergmark not to be a pest, but when could this go up for release? I'd like to get Bloodhound out of stackage-nonexistence :)

bitemyapp:
@bergmark :heart: Thank you very much :)

PR #366, opened 2016-02-13
Fix a bug with generics and single-field records

neongreen:
Here's the code that reproduces the bug:

    import Data.Aeson
    import Data.Aeson.TH
    import GHC.Generics

    newtype Wrap a = Wrap {unwrap :: a}
      deriving Generic

    return []

    thToEncodingUnwrap :: Wrap String -> Encoding
    thToEncodingUnwrap =
      $(mkToEncoding defaultOptions{unwrapUnaryRecords=True} ''Wrap)

    gToEncodingUnwrap :: Wrap String -> Encoding
    gToEncodingUnwrap =
      genericToEncoding defaultOptions{unwrapUnaryRecords=True}

The results of thToEncodingUnwrap and gToEncodingUnwrap differ:

    > thToEncodingUnwrap (Wrap "")
    "\"\""
    λ> gToEncodingUnwrap (Wrap "")
    "\"unwrap\":\"\""

This was because the `toEncoding` code couldn't inspect the generated `Builder` and break it into key:value, but when the record has only 1 field the proper way to encode it (when `unwrapUnaryRecords` is turned on) is taking the value without the key.

### DIFF
--- a/Data/Aeson/Types/Generic.hs
+++ b/Data/Aeson/Types/Generic.hs
@@ -247,7 +247,7 @@ instance OVERLAPPING_ TaggedObjectEnc' U1 False where
     taggedObjectEnc' _ _ _ = Tagged mempty
 
 instance (RecordToEncoding f) => TaggedObjectEnc' f True where
-    taggedObjectEnc' opts _ = Tagged . (\z -> B.char7 ',' <> recordToEncoding opts z)
+    taggedObjectEnc' opts _ = Tagged . (B.char7 ',' <>) . fst . recordToEncoding opts
 
 instance (GToEncoding f) => TaggedObjectEnc' f False where
     taggedObjectEnc' opts contentsFieldName =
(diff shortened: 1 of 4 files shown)


neongreen:
The build fails with GHC 7.4.2:

~~~
tests/DataFamilies/Types.hs:15:15:
    Couldn't match type `Rep (Nullary Int)'
                   with `M1 t42 t43 (M1 t44 t45 U1 :+: (M1 t46 t47 U1 :+: M1 t48 t49 U1))'
    Expected type: Rep (Nullary Int) x
      Actual type: M1 t42 t43 (M1 t44 t45 U1 :+: (M1 t46 t47 U1 :+: M1 t48 t49 U1)) x
    In the pattern: M1 (L1 (M1 U1))
    In an equation for `to': to (M1 (L1 (M1 U1))) = C1
    In the instance declaration for `Generic (Nullary Int)'
~~~

Googling suggests it's a known bug in GHC 7.4. Moreover, the bug isn't in Aeson – if somebody tries to write `deriving Generic` for a data family, ne's going to get this error message regardless of whether ne uses Aeson or not. However, it still breaks the tests (and thus the Travis build), so I don't know what to do about it.

bergmark:
In that case I think we should add CPP for that test case

neongreen:
I added CPP.

bergmark:
Thanks @neongreen, looks great!

Since this is a breaking change it will be included in 0.12. I haven't decided if I should merge 0.12 changes into master or if I should have a separate branch for that...

RyanGlScott:
If you wish, you can use the [`generic-deriving`](http://hackage.haskell.org/package/generic-deriving-1.10.4/docs/Generics-Deriving-TH.html) library to get around old GHC bugs involving `deriving Generic` on data family instances. For example, you can replace:

```haskell
deriving instance Generic (Nullary Int)
```

with

```haskell
import Generics.Deriving.TH

$(deriveAll0 'C1) -- where C1 is a constructor for Nullary Int
```

bergmark:
I'm not sure how to merge this now that #307 is on master. @neongreen do you have time to take a look? Would appreciate it!

neongreen:
Sure, in a couple of hours.

neongreen:
I've rebased the pull request. It passes tests locally – but I have to say that I haven't reviewed the changes in #307 and so I don't guarantee that passes tests = works correctly. I'd appreciate it if someone who -has looked at aeson's generics implementation recently- could check whether my rebasing is valid.

bergmark:
Thank you for the quick turnaround @neongreen!

I added some extra test cases for #307 so I'm not that worried that this conflict would cause bugs, but if someone else wants to take a look that is of course always appreciated :-)

I also followed @RyanGlScott's suggestion to use generic-deriving since it only adds the dependency to the test-suite.

Issue #361, opened 2016-02-09
instances for Proxy, Tagged and Const


phadej:
As `tagged` is package providing `Proxy` for old bases, providing instances for `Tagged` is also ok?

`Proxy` could be serialised as `null`. It's there but not carrying any information.
`Tagged` and `Const` are virtually the same as `Identity`.

bergmark:
I would have expected `encode () = "null"`... but I'm not suggesting we should change it now :)

PR #367, opened 2016-03-01
Add Proxy, Tagged and Const instances

phadej:
Resolves https://github.com/bos/aeson/issues/361

This could be included in `0.11.1.0`?


aeson.cabal
```diff
@@ -149,9 +150,11 @@ test-suite tests
     aeson,
     attoparsec,
     base,
+    base-orphans >=0.5.3 && <0.6,
```
phadej: for `Eq a => Eq (Const a b)`

tests/Instances.hs
```diff
@@ -145,3 +149,12 @@ instance Arbitrary Version where
...
 #endif
+
+instance Arbitrary (Proxy a) where
+    arbitrary = pure Proxy
+
+instance Arbitrary b => Arbitrary (Tagged a b) where
+    arbitrary = Tagged <$> arbitrary
+
+instance Arbitrary a => Arbitrary (Const a b) where
+    arbitrary = Const <$> arbitrary
```
bergmark: Have these instances been proposed to QuickCheck? Do we need CPP?

phadej: Not yet. I'll make a PR soonish. Not sure whether Nick would like to depend on tagged though.

phadej: Also some of my other PR's restricts `QuickCheck` to `2.8.3` because of others orphan instances

PR #371, opened 2016-03-09
Make sure the hashes of two equal Values are always the same

### LINKED ISSUE #370: The hashes of two equal Values may be different
`Hashable` instance of `Value` doesn't obey the law that the hashes of two equal `Value`s must always be the same.

```haskell
>>> let a = object ["223" .= False, "807882556" .= True]
>>> let b = object ["807882556" .= True, "223" .= False]
>>> a
Object (fromList [("223",Bool False),("807882556",Bool True)])
>>> b
Object (fromList [("807882556",Bool True),("223",Bool False)])
>>> a == b
True
>>> hash a
-408214619
>>> hash b
229481997
>>> hash a == hash b
False
```

watashi:
Resolves #370

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -90,6 +90,10 @@ import Data.Monoid (Monoid(..))
 import Data.Traversable (Traversable(..))
 #endif
 
+#if !MIN_VERSION_unordered_containers(0,2,6)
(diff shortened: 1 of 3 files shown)


bergmark:
Thank you!

bergmark:
I released this in [v0.11.1.1](http://hackage.haskell.org/package/aeson-0.11.1.1)

PR #378, opened 2016-03-22
Fix typo in 125b641785e1ce3f9facfaf7199de2917d4

hvr:
The `__GLASGOW_HASKELL__` identifier was written incorrectly
therefore causing 125b641785e1ce3f9facfaf7199de2917d4 to
have no effect, and hence making `aeson` fail on TH-less GHCs


bergmark:
Thanks!

PR #380, opened 2016-03-23
Small typo

reygoch:
It seems like there was a typo which prevented package from compiling on Windows. Now it works.


bergmark:
This is not a typo but a missing import, I'll release a fix shortly.

bergmark:
I just released aeson-0.11.1.4 with a fix.

reygoch:
Excellent, thanks!

PR #388, opened 2016-04-11
Added unsafeToEncoding

eskimor:
I needed something like that, because I have an encoded JSON from a library (servant-server) and want to include it in another JSON object. I want to skip an additional decoding/encoding. 

If there is another way to accomplish this - I am happy too! If not: Here is a PR ;-)

### DIFF
--- a/Data/Aeson/Types.hs
+++ b/Data/Aeson/Types.hs
@@ -14,6 +14,7 @@ module Data.Aeson.Types
     -- * Core JSON types
       Value(..)
     , Encoding
+    , unsafeToEncoding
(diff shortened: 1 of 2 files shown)


bergmark:
Sounds good to me! Sorry for not reacting earlier, I must have accidentally archived the notification.

eskimor:
Very cool - thank you for merging!

bergmark:
Shipped https://hackage.haskell.org/package/aeson-0.11.2.0

eskimor:
Wow even better! Thank you!

PR #390, opened 2016-04-26
Enabled PolyKinds in Data.Aeson.Types.Instances

duairc:
This is needed to make the instances for `Proxy`, `Tagged` and `Const` fully general.

### DIFF
--- a/Data/Aeson/Types/Instances.hs
+++ b/Data/Aeson/Types/Instances.hs
@@ -4,6 +4,11 @@
     ViewPatterns #-}
 {-# LANGUAGE DefaultSignatures #-}
 
+-- Needed for Tagged, Const and Proxy instances
+#if __GLASGOW_HASKELL__ >= 706
+{-# LANGUAGE PolyKinds #-}
+#endif
+
 #define NEEDS_INCOHERENT
 #include "overlapping-compat.h"
 


bergmark:
Thank you for your contribution!

It is unclear to me what the impact of this is, can you help? Is this a breaking change or would it only break orphan instances patched to allow other cases, or not even that?

And is the reason GHC 7.4 is excluded because "PolyKinds yields broken interface files in 7.4! This was fixed in 7.6" as mentioned in the [language pragma history](https://ghc.haskell.org/trac/ghc/wiki/LanguagePragmaHistory)?

duairc:
Hey, sorry for not getting back to you.

Basically, the current instances only work for (e.g.) `FromJSON a => FromJSON (Tagged (s :: *)) a`, while with `PolyKinds` it's `FromJSON a => FromJSON (Tagged (s :: k)) a`, where `k` can be any kind (including `*`). So it won't break any code that relied on the `*`-specific instances. It *will* break polykinded orphan instances, and also non-overlapping "monokinded" orphan instances for non-`*` kinds if they exist.

TL;DR: Yes, it will break orphan instances, but is otherwise not a breaking change.

As for not including GHC 7.4, I can't remember the specifics of it, but I know that `PolyKinds` is considered broken on GHC 7.4, possibly because of the issue you describe. For example, the [tagged](https://hackage.haskell.org/package/tagged-0.8.3/docs/src/Data-Tagged.html) package only enables `PolyKinds` on GHC >= 7.6.

bergmark:
Thanks, this is good stuff!

bergmark:
Shipped https://hackage.haskell.org/package/aeson-0.11.2.0

Issue #351, opened 2016-02-02
overlapping instances for [Char]


phadej:
While trying to compile `bower-json-0.7.0.0` with current `bos/master`: I get an error:

```
    bower-json/src/Web/Bower/PackageMeta/Internal.hs:318:26:
        Overlapping instances for Aeson.ToJSON [a]
          arising from a use of ‘.=’
        Matching instances:
          instance [overlappable] Aeson.ToJSON a => Aeson.ToJSON [a]
            -- Defined in ‘aeson-0.11.0.0:Data.Aeson.Types.Instances’
          instance Aeson.ToJSON [Char]
            -- Defined in ‘aeson-0.11.0.0:Data.Aeson.Types.Instances’
```

Used `stack.yaml`:

```yaml
resolver: lts-5.1
packages:
- '.'
- location:
    git: git@github.com:bos/aeson.git
    commit: fdb867286b6b4fd5afc703718b0a30dadc300db2
extra-deps:
- fail-4.9.0.0
```

It compiles fine with `aeson` versions in `LTS-4.2` and `LTS-5.1` though. I have no idea what happened lately in master.

---

Anyway, my proposal is to remove overlapping `FromJSON` and `ToJSON` instances, by the same trick as with `Read` and `Show`, i.e. introducing `parseJSONList` and `listToJSON` functions. They would have default implementations, so this is not breaking change. With https://github.com/bos/aeson/issues/349 fixed, there won't be other overlapping instances.

bergmark:
Bisect shows 8e1dbc7e723c7e20758368559f7093f1cfd5f742 as the bad commit

hvr:
IMHO the hoops one has to jump through to support a  `[Char]` instance is more trouble than it's worth... I can't wait for the day we can exterminate `type String = [Char]`, Haskell's biggest design flaw :-/

phadej:
I'm happy to do the PR to whatever is the agreement on the final solution. #354 is just the smallest possible fix, I guess.

bergmark:
I vote that we ship 0.11 with #354 for the least suprise. The Read/Show strategy seems like a good enhancement after that. 'll reopen this issue

andrewthad:
I'm also in favor of the `Read`/`Show` strategy to get rid of `OverlappingInstances`.

andrewthad:
[...]

PR #391, opened 2016-05-02
Higherkinded special strings

andrewthad:
This PR was formed by:

- Start from master branch
- Pull https://github.com/bos/aeson/pull/351 to stop using overlapping instances for strings
- Pull https://github.com/bos/aeson/pull/381 to get higher kinded typeclasses (`ToJSON1`, etc.)
- Modify higher kinded typeclass definitions and instances to conform to the pattern outlined by @RyanGlScott in [a comment](https://github.com/bos/aeson/issues/351#issuecomment-215855423).

I would appreciate any feedback if anyone sees anything that looks wrong. In particular, it's possible that I ended up with a non-list container that serializes to a json string instead of an array. All existing tests are still passing.

### DIFF
--- a/Data/Aeson/Types/Class.hs
+++ b/Data/Aeson/Types/Class.hs
@@ -39,8 +39,9 @@ module Data.Aeson.Types.Class
     -- * Functions needed for documentation
     , typeMismatch
     -- * Encoding functions
-    , list
+    , listEncoding
     , listValue
+    , listParser
     ) where
 
 import Data.Aeson.Encode.Builder
@@ -329,9 +330,7 @@ typeMismatch expected actual =
 class FromJSON1 f where
     liftParseJSON :: (Value -> Parser a) -> (Value -> Parser [a]) -> Value -> Parser (f a)
     liftParseJSONList :: (Value -> Parser a) -> (Value -> Parser [a]) -> Value -> Parser [f a]
-    liftParseJSONList f g v = case v of
-        Array vals -> fmap V.toList (V.mapM (liftParseJSON f g) vals)
-        _ -> typeMismatch "[a]" v
+    liftParseJSONList f g v = listParser (liftParseJSON f g) v 
 
 -- | Lift the standard 'parseJSON' function through the type constructor.
 parseJSON1 :: (FromJSON1 f, FromJSON a) => Value -> Parser (f a)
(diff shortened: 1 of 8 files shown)


Data/Aeson/Types/Class.hs
```diff
@@ -272,3 +320,95 @@ typeMismatch expected actual =
...
+
+list :: (a -> Encoding) -> [a] -> Encoding
+list _  []     = emptyArray_
+list to (x:xs) = Encoding $
+                B.char7 '[' <> fromEncoding (to x) <> commas xs <> B.char7 ']'
+      where commas = foldr (\v vs -> B.char7 ',' <> fromEncoding (to v) <> vs) mempty
+{-# INLINE list #-}
+
+listValue :: (a -> Value) -> [a] -> Value
+listValue f = Array . V.fromList . map f
```
RyanGlScott: Perhaps `list` should be renamed to `listEncoding`?

Data/Aeson/Types/Class.hs
```diff
@@ -272,3 +320,95 @@ typeMismatch expected actual =
...
+-- Lifings of FromJSON and ToJSON to unary and binary type constructors
+-------------------------------------------------------------------------------
+
+-- | Lifting of the 'FromJSON' class to unary type constructors.
+class FromJSON1 f where
+    liftParseJSON :: (Value -> Parser a) -> (Value -> Parser [a]) -> Value -> Parser (f a)
+    liftParseJSONList :: (Value -> Parser a) -> (Value -> Parser [a]) -> Value -> Parser [f a]
+    liftParseJSONList f g v = case v of
+        Array vals -> fmap V.toList (V.mapM (liftParseJSON f g) vals)
+        _ -> typeMismatch "[a]" v
```
RyanGlScott: We should really put this functionality into a combinator of type `Parser a -> Parser [a]` like we did for `listValue`.

Data/Aeson/Types/Class.hs
andrewthad: I agree. I had only left it named `list` here because I stole it out of `Data.Aeson.Encode.Functions`, but `listEncoding` is a much better name.

Data/Aeson/Types/Class.hs
andrewthad: Agree. I'll pull this into a `listParser` function.

andrewthad:
Pushed some more changes that incorporate the suggestions from @RyanGlScott.

phadej:
@andrewthad could you rebase it. Can you write lifted instances for `Map` and `HashMap` also, or should I continue on this. I have few cycles this Thursday.

andrewthad:
@phadej I've merged changes from master into this branch. I've also added the higher kinded instances for Map and HashMap. Feel free to take over this branch and add any more instances. I won't have any time to look at it for the next few days.

Also, there's one thing that could use some review. I left the `ToJSON` and `FromJSON` instances for `Map` and `HashMap` alone instead of redefining them with `toJSON1` and friends. It would be better to do it that way, but I have a slight worry that it may hurt performance. If anyone could benchmark it and show that it doesn't hurt performance, it would be great to see the duplication there eliminated.

andrewthad:
Closing in favor of https://github.com/bos/aeson/pull/407

PR #393, opened 2016-05-08
Alternate key strategy

andrewthad:
This is an alternative approach that accomplishes the same thing as https://github.com/bos/aeson/pull/341. A lot of the helper functions in this branch were taken from that one. Here is an example on the JSON encodings that are produced:

    > let encodePrint = BC8.putStrLn . L.toStrict . encode
    > encodePrint $ Map.singleton ("hey" :: Text) (4 :: Int)
    {"hey":4}
    > encodePrint $ Map.singleton ("hey" :: String) (4 :: Int)
    {"hey":4}
    > encodePrint $ Map.singleton (2 :: Int) ("dog" :: Text)
    {"2":"dog"}
    > encodePrint $ Map.singleton True ("dog" :: Text)
    {"true":"dog"}
    > encodePrint $ Map.singleton (2 :: Int, 3 :: Int) ("dog" :: Text)
    [[[2,3],"dog"]]
    > encodePrint $ Map.singleton (2 :: Int, "bob" :: String) ("dog" :: Text)
    [[[2,"bob"],"dog"]]
    > encodePrint $ Map.singleton (Identity (4 :: Int)) ("dog" :: Text)
    {"4":"dog"}
    > encodePrint $ Map.singleton (Identity ("bobert" :: String)) ("dog" :: Text)
    {"bobert":"dog"}
    > encodePrint $ Map.singleton (Prelude.map Identity ("bobert" :: String)) ("dog" :: Text)
    [[["b","o","b","e","r","t"],"dog"]]

This implementation is not totally complete, and I have not benchmarked it. I do think that the implementation ends up being simpler than the one in the PR that inspired this approach.

### DIFF
--- a/Data/Aeson/Types/Class.hs
+++ b/Data/Aeson/Types/Class.hs
@@ -258,13 +258,13 @@ class KeyValue kv where
     infixr 8 .=
 
 data ToJSONKeyFunction a
-  = ToJSONKeyText (a -> Text, a -> Encoding)
-  | ToJSONKeyValue (a -> Value, a -> Encoding)
+    = ToJSONKeyText (a -> Text, a -> Encoding)
+    | ToJSONKeyValue (a -> Value, a -> Encoding)
 
 data FromJSONKeyFunction a
-  = FromJSONKeyText (Text -> a)
-  | FromJSONKeyTextParser (Text -> Parser a)
-  | FromJSONKeyValue (Value -> Parser a)
+    = FromJSONKeyText (Text -> a)
+    | FromJSONKeyTextParser (Text -> Parser a)
+    | FromJSONKeyValue (Value -> Parser a)
 
 -- | Fail parsing due to a type mismatch, with a descriptive message.
 --
--- a/Data/Aeson/Types/Class.hs
+++ b/Data/Aeson/Types/Class.hs
@@ -37,7 +37,11 @@ module Data.Aeson.Types.Class
 import Data.Aeson.Types.Internal
 import Data.Text (Text)
 import GHC.Generics (Generic, Rep, from, to)
+import Data.Monoid ((<>))
+import Data.Aeson.Encode.Builder (emptyArray_)
+import qualified Data.ByteString.Builder as B
 import qualified Data.Aeson.Encode.Builder as E
+import qualified Data.Vector as V
 
 -- | Class of generic representation types ('Rep') that can be converted to
 -- JSON.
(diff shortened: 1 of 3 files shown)


bergmark:
style nitpick: please use 4 spaces for indentation.

bergmark:
I played a bit with this, but I got stuck on trying to use the previously working rest-stringmap, over at https://github.com/bos/aeson/tree/stringmap

```haskell
λ> import Data.Map
λ> eitherDecode "{\"a\":\"b\"}" :: Either String (Map Text Text)
Right (fromList [("a","b")])

λ> eitherDecode "{\"a\":\"b\"}" :: Either String (HashMap Text Text)
Right (fromList [("a","b")])

λ> eitherDecode "{\"a\":\"b\"}" :: Either String (Rest.StringMap.HashMap.Strict.StringHashMap Text Text)
Left "Error in $: expected Map k v, encountered Object"

λ> eitherDecode "{\"a\":\"b\"}" :: Either String (Rest.StringMap.Map.Strict.StringMap String String)
Left "Error in $: expected Map k v, encountered Object"
```

Did i do something wrong?

andrewthad:
Sorry it took me so long to look at this. I should have made `String` key parsing work before I put this up. I think the issue you ran into is because:

1. The `FromJSONKey` [instance for `Char`](https://github.com/bos/aeson/blob/db78b7da7c36a818875f34541367936047662689/Data/Aeson/Types/Instances.hs#L1642) should have `parseJSONKeyList` defined.
2. The `fromJSONKey` definition for the [list instance](https://github.com/bos/aeson/blob/db78b7da7c36a818875f34541367936047662689/Data/Aeson/Types/Instances.hs#L1662) should call `fromJSONKeyList`.

I'll try to clean this up more this weekend.

bergmark:
Thanks! Good thing my mistake was in aeson instead of a silent breakage in rest-stringmap.

andrewthad:
Parsing of string keys should now be working correctly.

bergmark:
There are some overlapping instances, e.g. `FromJSON (M.Map String v)` wrt the ToJSONKey instance

phadej:
I'd like to see benchmark comparison with #341 (and have that benchmark included)!

Data/Aeson/Types/Class.hs
```diff
@@ -250,6 +257,15 @@ class KeyValue kv where
...
     infixr 8 .=
 
+data ToJSONKeyFunction a
+    = ToJSONKeyText (a -> Text, a -> Encoding)
+    | ToJSONKeyValue (a -> Value, a -> Encoding)
+
+data FromJSONKeyFunction a
+    = FromJSONKeyText (Text -> a)
+    | FromJSONKeyTextParser (Text -> Parser a)
+    | FromJSONKeyValue (Value -> Parser a)
```
phadej: `FromJSONKeyFunction` can be `Functor`

Data/Aeson/Types/Class.hs
```diff
@@ -250,6 +257,15 @@ class KeyValue kv where
     (.=) :: ToJSON v => Text -> v -> kv
     infixr 8 .=
 
+data ToJSONKeyFunction a
+    = ToJSONKeyText (a -> Text, a -> Encoding)
+    | ToJSONKeyValue (a -> Value, a -> Encoding)
```
phadej: `ToJSONKeyFunction` can be [`Contravariant`](http://hackage.haskell.org/package/contravariant-1.4/docs/Data-Functor-Contravariant.html#t:Contravariant), IMHO that is quite light dependency, but might make writing instances handy (vs. providing `contramap` with new name specialised for `ToJSONKeyFunction`)

Data/Aeson/Types/Instances.hs
```diff
@@ -1584,6 +1608,94 @@ instance FromJSON a => FromJSON (Const a b) where
...
+    default toJSONKeyList :: ToJSON a => ToJSONKeyFunction [a]
+    toJSONKeyList = ToJSONKeyValue (toJSON, toEncoding)
+
+class FromJSONKey a where
+    fromJSONKey :: FromJSONKeyFunction a
+    default fromJSONKey :: FromJSON a => FromJSONKeyFunction a
+    fromJSONKey = FromJSONKeyValue parseJSON
+    fromJSONKeyList :: FromJSONKeyFunction [a]
+    default fromJSONKeyList :: FromJSON a => FromJSONKeyFunction [a]
+    fromJSONKeyList = FromJSONKeyValue parseJSON
```
phadej: This default implementation is frustrating.

andrewthad:
Closing this in favor of a branch where @phadej has made further improvements: https://github.com/bos/aeson/pull/399. Any discussion should continue there.

Data/Aeson/Types/Class.hs
andrewthad: Yeah, this would be a good improvement.

Data/Aeson/Types/Class.hs
andrewthad: I thought about this, but I feel like the dependency on `contravariant` probably not worth it when we could just provide a specialized version of the function instead.

Issue #387, opened 2016-04-06
Invalid JSON generated by genericToEncoding


alexanderkjeldaas:
`omitNothingFields` does not work for `genericToEncoding`
```haskell
> data Test = Test { x :: Maybe String, y :: Maybe String }  deriving (Generic)
> instance ToJSON Test where toEncoding =
      genericToEncoding defaultOptions {omitNothingFields = True}
> encode $ Test Nothing Nothing
"{,}"
```

bergmark:
Thank you for the report!

This is similar to #366 but is not fixed by that PR.

tolysz:
I do not want to say it; but we miss a constructor of `VALUE` which represents the values we want to omit. (as it is the container which needs to remove the value; not just value to emit nothing).

Without the additional constructor ( https://skillsmatter.com/skillscasts/6587-programming-with-universal-properties ) there is no way to have a function which maps 
`Null` value to 2 states at the same time.

Tritlo:
I'm having a similar issue where when I use `omitNothingFields`, I get an empty `,`, i.e.

    data GraphResponse = GraphResponse { data :: [Value]
                                       , errors :: Maybe [Value]
                                       , info :: Maybe Value
                                       } deriving (Show, Generic, FromJSON)



    instance ToJSON GraphResponse where
      toJSON = genericToJSON defaultOptions {omitNothingFields = True}
      toEncoding = genericToEncoding defaultOptions {omitNothingFields = True}


where `GraphResponse [] Nothing Nothing` results in the `JSON` `{"data":[],,}`, i.e. with extra commas.

Issue #392, opened 2016-05-06
v1.0.0.0 planning


bergmark:
in suggested merge order:
- [x] #300 via #307 Removing contents field for nullary constructors
- [x] Investigate whether #307 is/can be made a non-breaking change (no since the encode output changes, but `"contents" : []` is luckily still accepted when parsing)
- [x] #366 Fix a bug with generics and single-field records
- [x] #352 via #381 Lifted type classes
- [x] #351 via #382 stop using overlapping instances for strings
- [x] #289 via #341 Generalized map serialization
- [x] #391 Modify higher kinded typeclasses from #381
- [x] #387 Fix omitNothingFields + genericToEncoding via #394
- [x] #357 JSONPath isn't tracked in autoderived instances (TH & generics) via #398
- [x] #375 Improve haddocks by not using orphan instances
- [x] Generics for Higher kinded type classes via #414 and #415
- [x] Add documentation on how to use/write `ToJSONKey` (etc) instances. @andrewthad
- [x] Finalise encoding refactoring @phadej
- [x] Add notes on all these checklist items to the changelog @bergmark
- [x] Write tests for lifted typeclasses and their generic/th derivations @phadej
- [x] Do some stackage/Silk testing @bergmark
- [x] Fix recently added TODOs
- [x] #317 Do not accept and truncate floats to integrals

Please give feedback if I missed something!

bergmark:
@andrewthad @phadej should we start by merge the lifted type classes next?

andrewthad:
Yeah, I think that that one is the best place to start since  it was rebased on top of master like a week ago.

RyanGlScott:
We should probably add generics/TH support for the lifted typeclasses to this list. I haven't made an issue for it, since I was hoping it could be a part of #352, but I dropped the ball.

andrewthad:
@RyanGlScott If I understand correctly, the generics/TH support is a non-breaking change. Even if it were added after the `0.12` release, I think it would only require bumping minor version. If it's something that you can do in under a week, I don't think that anyone would object to you adding it for the `0.12` release though.

andrewthad:
Also @bergmark, I'd still like to redo the module merge from https://github.com/bos/aeson/issues/375 to eliminate orphan instances. I think that now is probably a good time to do that. Let me know if you agree and I'll do it.

andrewthad:
Nevermind, I just noticed https://github.com/bos/aeson/pull/412

andrewthad:
[...]

PR #394, opened 2016-05-10
Fix issue #387 with encoding using omitNothingFields

Lysxia:


bergmark:
Thank you! This looks good to me. I've added it to the list over at #392 so it will get merged after the stuff that was already on there.

bergmark:
I decided to merge #396 instead of this. Still, thank you for your contribution!

Lysxia:
No problem!

PR #395, opened 2016-05-13
Build also with stack on travis (benchmarks)

phadej:


hvr:
sadly this results in less readable travis log-output as you can't collapse the sections anymore

phadej:
@hvr, that's true :(

phadej:
Alternatively we could merge benchmarks into `aeson` package, if there aren't really good reason why they are separate package?

bergmark:
Can't, the benchmarks have dependencies like statistics that depend on aeson

phadej:
finally green

hvr:
@bergmark re cyclic deps, see https://github.com/haskell/cabal/pull/3422

mgsloan:
Stack has actually supported cyclic deps for quite some time, though it gets a little dicey with older Cabals due to https://github.com/haskell/cabal/issues/2870

PR #396, opened 2016-05-14
use mconcat and intersperse to avoid extra commas. Fixes #387

### LINKED ISSUE #387: Invalid JSON generated by genericToEncoding

Tritlo:
This is a fix for #387. When an item is nothing, the result of encoding it is empty, but the concatenation code does not take this into account. This uses intersperse and mconcat for the desired behaviour. I am not sure about the performance characteristics of the way to filter out the empty values, and would love some input. Possible improvement is using `toLazyByteStringWith` from `Data.ByteString.Builder.Extra`, though I am unsure of the internals of `ByteString`. It seems as though there should be an easy way to check if a `Builder` has only the `empty` bytestring, though I did not find it.

Edit: I hadn't seen #394 when I created this, since it's not linked on #387 for some reason. That fixes #387 specifically, but not my issue, i.e. that there are extra commas added for empty field.

### DIFF
--- a/Data/Aeson/Types/Generic.hs
+++ b/Data/Aeson/Types/Generic.hs
@@ -30,9 +30,11 @@ import Data.Aeson.Types.Instances
 import Data.Aeson.Types.Internal
 import Data.Bits (unsafeShiftR)
 import Data.ByteString.Builder as B
+import qualified Data.ByteString.Lazy as BSL (null)
 import Data.DList (DList, toList, empty)
 import Data.Maybe (fromMaybe)
 import Data.Monoid ((<>))
+import Data.List (intersperse)
 import Data.Text (Text, pack, unpack)
 import GHC.Generics
 import qualified Data.HashMap.Strict as H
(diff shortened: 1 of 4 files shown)


Tritlo:
The build seems to be failing due to some cabal configure issue? And only on ghc 8?

Tritlo:
I've added an updated version, which only jumps through the comparing hoops when the `omitNothingFields` option is on.

bergmark:
There is already #394, but this looks like a cleaner patch. Did you write any tests / run the tests in #394?

Tritlo:
I did not write any tests other than testing it on some objects. I ran the tests from #394, but I was unsure whether I should include them (seeing as I did not write them myself). I have now cherry-picked the tests commit from Lysxia's branch and added them here, so that the credit for them goes where the credit is due.

PR #398, opened 2016-05-20
Track JSONPath in autoderived parsers. Fixes #357

### LINKED ISSUE #357: JSONPath isn't tracked in autoderived instances
Let's create a type and derive an instance for it.

With TH:

~~~ haskell

import Data.Aeson
import Data.Aeson.TH

data X = X {a :: (Int, Int, Int)} 
  deriving Show

deriveJSON defaultOptions ''X
~~~

Or with generics:

~~~ haskell

import Data.Aeson

data X = X {a :: (Int, Int, Int)}
  deriving (Show, Generic)

instance FromJSON X where
  parseJSON = genericParseJSON defaultOptions 
~~~

Then try to decode an invalid value:

~~~
> eitherDecode "{\"a\": [1, 2, true]}" :: Either String X
Left "Error in $[2]: expected Int, encountered Boolean"
~~~

I believe it should be `$.a[2]`.

(I checked with latest Aeson from the repository, commit 2f0b989.)

Lysxia:
The first commit is not a direct fix, but gets rid of some redundancy in error messages in nested records.

In the [definition of `(.:)`](https://github.com/bos/aeson/blob/803188803d402090ed515e5738710d542e29a244/Data/Aeson/Types/Instances.hs#L1658-L1659) (as well as its siblings `(.:?)`, etc.) the location information is provided twice. 

```haskell
    data X = {a :: (Int,Int,Int)}
    data Y = {x :: X}
    instance FromJSON X where
        parseJSON (Object obj) = X <$> obj .: "a"
    instance FromJSON Y where
        parseJSON (Object obj) = Y <$> obj .: "x"
```

This results in this kind of error when using `decodeEither`:

```haskell
    > eitherDecode "{\"x\":{\"a\": [1,2,true]}}" :: Either String Y
    Left "Error in $.x.a[2]: failed to parse field x: failed to parse field a: expected Int, encountered Boolean"
```

The location is given in two different forms; the second one is extremely verbose and also incomplete, missing the `[2]`.

This commit simply removes the use of `modifyFailure` in the combinators, so that only the compact path description is left.

    Left "Error in $.x.a[2]: expected Int, encountered Boolean"

One thing that might be problematic with this solution is that the function `parse :: (a -> Parser b) -> a -> Result b` no longer has any location information by default since it discards the `JSONPath` parameter.  OTOH some users might prefer to insert `modifyFailure` themselves to customize their error messages, which previously prevented the use of `(.:)`.

### DIFF
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -107,7 +107,7 @@ import Data.Aeson.Types ( Value(..), Parser
                         , defaultOptions
                         , defaultTaggedObject
                         )
-import Data.Aeson.Types.Internal (Encoding(..))
+import Data.Aeson.Types.Internal (Encoding(..), (<?>), JSONPathElement(Key))
 import Control.Monad       ( liftM2, return, mapM, fail )
 import Data.Bool           ( Bool(False, True), otherwise, (&&), not )
 import Data.Either         ( Either(Left, Right) )
@@ -1010,7 +1010,7 @@ instance OVERLAPPABLE_ (FromJSON a) => LookupField a where
     lookupField tName rec obj key =
         case H.lookup key obj of
           Nothing -> unknownFieldFail tName rec (T.unpack key)
-          Just v  -> parseJSON v
+          Just v  -> parseJSON v <?> Key key
 
 instance (FromJSON a) => LookupField (Maybe a) where
     lookupField _ _ = (.:?)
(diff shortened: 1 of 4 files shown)


Lysxia:
Oh wait, the Generic is still wrong.

Lysxia:
This is why I should test before submitting the PR.

Lysxia:
Fixes #357

bergmark:
Thank you for this! Looks very nice.

PR #399, opened 2016-05-25
Alternate key strategy oleg

phadej:
Follow-up to https://github.com/bos/aeson/pull/393

- [Original](http://oleg.fi/aeson-bench/datakinds.html)
- [without coerce](http://oleg.fi/aeson-bench/without-coerce.html)
- [with coerce](http://oleg.fi/aeson-bench/with-coerce.html)

Looks like that alternative approach is as fast, and is simpler. Good job @andrewthad !
Coerce approach gives a little speedup, don't know if it's worth the safety (common case in our codebase though, where the keys are `newtypes` of `Text`).

Didn't add dependency on `contravariant.

### DIFF
--- a/Data/Aeson/Types.hs
+++ b/Data/Aeson/Types.hs
@@ -37,6 +37,12 @@ module Data.Aeson.Types
     , KeyValue(..)
     , modifyFailure
 
+    -- ** Keys for maps
+    , ToJSONKey(..)
+    , ToJSONKeyFunction(..)
+    , FromJSONKey(..)
+    , FromJSONKeyFunction(..)
+
     -- ** Generic JSON classes
     , GFromJSON(..)
     , GToJSON(..)
--- a/Data/Aeson/Types.hs
+++ b/Data/Aeson/Types.hs
@@ -42,6 +42,8 @@ module Data.Aeson.Types
     , ToJSONKeyFunction(..)
     , FromJSONKey(..)
     , FromJSONKeyFunction(..)
+    , fromJSONKeyCoerce
+    , coerceFromJSONKeyFunction
 
     -- ** Generic JSON classes
     , GFromJSON(..)
--- a/Data/Aeson/Types/Class.hs
+++ b/Data/Aeson/Types/Class.hs
@@ -262,7 +262,8 @@ data ToJSONKeyFunction a
     | ToJSONKeyValue (a -> Value, a -> Encoding)
 
 data FromJSONKeyFunction a
-    = FromJSONKeyText (Text -> a)
+    = FromJSONKeyCoerce
+    | FromJSONKeyText (Text -> a)
     | FromJSONKeyTextParser (Text -> Parser a)
     | FromJSONKeyValue (Value -> Parser a)
   deriving (Functor)
--- a/Data/Aeson/Types/Class.hs
+++ b/Data/Aeson/Types/Class.hs
@@ -258,13 +258,13 @@ class KeyValue kv where
     infixr 8 .=
 
 data ToJSONKeyFunction a
-  = ToJSONKeyText (a -> Text, a -> Encoding)
-  | ToJSONKeyValue (a -> Value, a -> Encoding)
+    = ToJSONKeyText (a -> Text, a -> Encoding)
+    | ToJSONKeyValue (a -> Value, a -> Encoding)
 
 data FromJSONKeyFunction a
-  = FromJSONKeyText (Text -> a)
-  | FromJSONKeyTextParser (Text -> Parser a)
-  | FromJSONKeyValue (Value -> Parser a)
+    = FromJSONKeyText (Text -> a)
+    | FromJSONKeyTextParser (Text -> Parser a)
+    | FromJSONKeyValue (Value -> Parser a)
 
 -- | Fail parsing due to a type mismatch, with a descriptive message.
 --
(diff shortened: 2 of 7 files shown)


phadej:
@andrewthad 
> I thought about this, but I feel like the dependency on `contravariant` probably not worth it when we could just provide a specialized version of the function instead.

`contravariant` is very light dependency, doesn't bring anything else except itself. But I leave it to maintainers to decide

andrewthad:
@phadej I'm assuming you mean that `aeson` already transitively depends on everything that `contravariant` depends on. I haven't checked to see if that is the case, but if it is, then that would make me feel more comfortable with depending on it.

phadej:
```diff
 packages:
 - '.'
 extra-deps:
+- StateVar-1.1.0.4
 - attoparsec-0.13.0.2
+- contravariant-1.4
 - dlist-0.7.1.2
 - fail-4.9.0.0
 - hashable-1.2.4.0
 - mtl-2.2.1
 - primitive-0.6.1.0
 - scientific-0.3.4.6
+- semigroups-0.18.1
+- stm-2.4.4.1
 - syb-0.6
 - tagged-0.8.4
 - text-1.2.2.1
+- transformers-compat-0.5.1.4
 - unordered-containers-0.2.7.0
 - vector-0.11.0.0
+- void-0.7.1
```

And with https://github.com/ekmett/contravariant/pull/27 the `void` and `semigroups` aren't pulled in.

andrewthad:
I've opened up a PR against this PR that fixes two small things: https://github.com/phadej/aeson/pull/2

Thanks for doing the benchmarks to look into the performance issues. After spending a lot of time yesterday thinking about `unsafeCoerce` and inconsistent `Eq` and `Hashable` instances, I have ended up in favor of the `FromJSONKeyCoerce` data constructor. I think it is the best solution we have available. So I'm +1 on merging this in to master.

As a side note, in another branch, I played around with [using rewrite rules instead of unsafeCoerce](https://github.com/andrewthad/aeson/commit/7e8a740fd95d5791a457307740d421ad4d9e0540) to make decoding `HashMap`s with `Text` keys faster, but it ends up only working for `Text` and not for `newtype`s around `Text`, so the `unsafeCoerce` code ends up being necessary no matter regardless.

andrewthad:
Also, @bergmark it would be good to hear your thoughts on the `contravariant` dependency.

phadej:
- http://oleg.fi/aeson-bench/safer-coercible.html
- http://oleg.fi/aeson-bench/unsafe.html 

I'm not sure what happening with `Map` in benchmark in top version

Data/Aeson/Types.hs
```diff
@@ -37,6 +37,14 @@ module Data.Aeson.Types
     , KeyValue(..)
     , modifyFailure
 
+    -- ** Keys for maps
+    , ToJSONKey(..)
+    , ToJSONKeyFunction(..)
+    , FromJSONKey(..)
+    , FromJSONKeyFunction(..)
+    , fromJSONKeyCoerce
+    , coerceFromJSONKeyFunction
```
phadej: These two functions aren't exported from `Data.Aeson` as they are quite special.

Data/Aeson/Types/Class.hs
```diff
@@ -261,13 +276,50 @@ data ToJSONKeyFunction a
...
+
 data FromJSONKeyFunction a
-    = FromJSONKeyCoerce
+    = FromJSONKeyCoerce (CoerceText a)
     | FromJSONKeyText (Text -> a)
     | FromJSONKeyTextParser (Text -> Parser a)
     | FromJSONKeyValue (Value -> Parser a)
-  deriving (Functor)
 
+instance Functor FromJSONKeyFunction where
```
andrewthad: Not a valid `Functor` instance since the following does not hold:

    fmap id x = x

phadej: I'd say it's valid enough. Especially if we add `fmap id = coerceFromJSONKeyFunction` and `fmap coerce = coerceFromJSONKeyFunction` rewrite rules!

andrewthad:
@phadej Concerning your two most recent benchmarks, my interpretation of them is that the safer one is faster than the unsafe one. Intuitively, that seems wrong. Am I reading these correctly?

phadej:
They both take about of the same time. My machine isn't the best one to run the benchmarks (mail client etc. running). I posted them just to show that safer approach doesn't hurt the performance, if it's even faster - that's only good thing :) (might be that GHC can do something with `coerce`, which it cannot with `unsafeCoerce` - dunno about that).

phadej:
Seems to be transient failure on travis

bergmark:
> Also, @bergmark it would be good to hear your thoughts on the contravariant dependency.


Adding a dependency only for one instance sounds a bit excessive to me.

phadej:
So is it good as it is now?

andrewthad:
It looks good to me. The quirkiness of the `Functor` instance should probably be documented, but I'm planning on adding a more docs to these things in this PR after it's merged in, so that can wait for later.

bergmark:
Just to make sure, do you guys think this is ready to be merged?

phadej:
@bergmark yes

andrewthad:
yes

bergmark:
👏 Great work to both of you!

Issue #402, opened 2016-05-29
Encoding performance - toList Data.Sequence vs. foldable Data.Sequence


ondrap:
I was testing something slightly different and found out that doing `AE.encode seqVar` is about 15% slower than doing `AE.encode (toList seqVar)`. I am not sure how it works for other `Foldable` containers, however it seems to me it would make sense to change at least the `Seq` instance to:

```
toEncoding = list . toList
```

phadej:
We should write a benchmark for this,
`list` and `foldable` encoding are implemented differently, we probably can tweak them to have same "best" performance. 

Make sense to do after https://github.com/bos/aeson/pull/391

mgsloan:
The poor vector performance is likely due to https://github.com/haskell/vector/issues/112

bergmark:
Marked as "need-info" as we want benchmark results.

PR #404, opened 2016-06-01
Add encode/foldable benchmark

phadej:
Benchmark for https://github.com/bos/aeson/issues/402

screen shot 2016-06-01 at 22 43 18

Indeed, using `foldable` is suboptimal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 dist
 .cabal-sandbox/
 cabal.sandbox.config
+.stack-work/
 
 *.o
 *.hi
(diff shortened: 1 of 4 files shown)


phadej:
And resolved #402 as it's trivial:

screen shot 2016-06-01 at 22 59 00

I have no idea why `Vector` serialising is slower as it is now, but won't like to touch it now. Have to check other benchmarks as well.

phadej:
@mgsloan vice versa, using `toEncoding . toList` is faster then current specially implemented `toEncoding`. But indeed, with previous `foldable` implementation, it was pretty slow.

phadej:
Done in #407

PR #407, opened 2016-06-02
High kind strings

phadej:
Continuation of https://github.com/bos/aeson/pull/391

@andrewthad: I run benchmarks for `Map` serialising: there is no change before this PR, before of after *	Use Higher order *JSON classes for Map/HashMap* commit. Which is great!

TODO
- [x] rest of instances
- [x] cleanup


phadej:
- added rest of `ToJSON1` / `FromJSON1` instances
- didn't add more `FromJSONKey` and `ToJSONKey` instances as that can be done in follow-up PR
- organised stuff a bit

AFAICS this is ready for final review / merge. ping @andrewthad @bergmark

andrewthad:
This looks good to me. I think we should go ahead and get this into master and then add any other instances later as they are needed.

bergmark:
Let's do it!

bergmark:
With this `Data.Aeson.Types.Instances.Tuple` takes a very long time to compile, anything you've noticed?

phadej:
@bergmark I guess lots of instances + lots of inlining makes it quite bad

andrewthad:
@bergmark Yeah, I've noticed that too.

bergmark:
```
$ stack clean --full; and time stack build aeson --fast
[...]
       17.12 real        15.20 user         1.78 sys
$ stack clean --full; and time stack build aeson
[...]
      220.03 real       208.53 user         5.99 sys
```

:-(

Issue #375, opened 2016-03-16
FromJSON/ToJSON instances in haddocks


andrewthad:
I'm not sure why all of the instances for `FromJSON` and `ToJSON` were moved into an internal module (I assume there was a good reason), but it's now impossible to find these instances from the haddocks.

bergmark:
That's unfortunate. I don't know either as the offending commit (260a7e839bfda660bb7e0b715567b35d76cc4693) doesn't explain.

bergmark:
An alternative solution to merge the modules again is expose this module and make use of https://github.com/haskell/haddock/pull/449

phadej:
I'd guess that's somehow related to the haddock (bug or feature?), because at least haddock used to bundle orphan instances if class was exported too:
- http://hackage.haskell.org/package/binary-orphans-0.1.0.0/docs/Data-Binary-Orphans.html has `Binary Value` under `Binary` list, but
- http://hackage.haskell.org/package/binary-orphans-0.1.4.0/docs/Data-Binary-Orphans.html doesn't.

andrewthad:
If it's a bug, it's been this way for at least a year. Because `Day` from [Data.Time.Calendar](http://hackage.haskell.org/package/time-1.6/docs/Data-Time-Calendar.html) has the exact same problem. There's a `Show` instance orphaned somewhere else, and it's exported, but not documented.

andrewthad:
If I understand correctly, the situation we are in is summed up by the following two statements:

- We don't know why all the instances were orphaned
- Haddock currently does not show orphan instances

I would have assumed that the instances were orphaned so that the `FromJSON`/`ToJSON` combinators that they use (like `.:` and `.:?`) wouldn't need to be defined in the same module. Because without orphan instances, a class and its instances and any combinators (that mention the class) used by the instances all have to be defined in the same module.

But, I looked into it a little more and found that this wasn't the reason why this was done. All of the combinators are [still defined in the `Data.Aeson.Types.Instances` module](https://github.com/bos/aeson/blob/7da420a73632b8e3b24536da73ad7020be584849/Data/Aeson/Types/Instances.hs#L1677). So the only benefit of the split that I can see was that 1700 lines of instances (and combinators) were able to be moved out the the `Class` module.

I checked in on `Data.Aeson.Types.Generic` as well (because it provides the orphan instances for `GFromJSON` and `GToJSON`), and it seems like in that case the benefit is that 600 more lines of instances are separated from the class definitions.

[...]

Issue #301, opened 2015-10-12
Generating nested objects with new toEncoding


ondrap:
I have a type where the Json instance should be slightly more complex:

    instance ToJSON EsBulkDescr where
       toJSON (EsBulkDescr {esdIndex=idx, esdMapping=emap, esdId=msgId}) =
           object [ "index" .= object [ "_index" .= idx, "_type" .= emap, "_id" .= msgId ]]

I would expect, the toJSON should translate into something like:

       toEncoding (EsBulkDescr {esdIndex=idx, esdMapping=emap, esdId=msgId}) =
           pairs ( "index" .= pairs ("_index" .= idx <> "_type" .= emap <> "_id" .= msgId ))

However, I just didn't find a way to do it without creating new datatype for the embedded substructure. Would there be a way to make Encoding (or Series??) a ToJSON instance, so this could be supported? I have provisionally implemented this and it works, but maybe there is some better way:

    instance ToJSON Encoding where
       toJSON = error "Unimplemented"
       toEncoding = id

ondrap:
I did some more hacking and it seems to me more is needed:

* I think `Encoding` should not be instance of `Monoid`. I don't think there is a valid use outside of internals of aeson - I guess you can probably easily create invalid JSON by concatenating 2 encodings
* `Encoding` should have `ToJSON` instance (I don't like the `toJSON`error, but if it is supposed to be deprecated in the future, it seems to me relatively OK)
* It would be great to have some 'unsafe' functions to access the underlying bytestring builder; e.g. to implement the base64 encoding for bytestrings I have to convert to `Text` and the `encodeUtf8` it and escape it, wheras all 3 steps could be skipped if there was an unsafe access function to the builder.

dylex:
I think this can all be done much better by just exposing a single function that is a specialized variant of `(.=)`:
```
somethingKeyEncodingSeries :: Text -> Encoding -> Series
somethingKeyEncodingSeries name value = Value $ Encoding $ E.text name <> B.char7 ',' <> fromEncoding value
```
Obviously it needs a better name, but once you had this, you could just replace the first `.=` with it to fix everything, without needing any extra broken instances or removing the `Monoid`.

dylex:
It would also be nice to have a version of `foldableEncoding :: Foldable t => t Encoding -> Encoding` to do the equivalent for arrays.

ondrap:
Is there any legal use for the Monoid instance of Encoding? It seems to me that the only use is for generating invalida json output.

[...]

PR #412, opened 2016-06-03
Abstract `Encoding`

phadej:
This PR makes everything use `Data.Aeson.Encoding` -module, except `.Instances`.
I run some benchmarks, cannot notice any difference (which is good).

If this idea is good, we could merge this before we reorganise instances (https://github.com/bos/aeson/issues/375)

### DIFF
--- a/Data/Aeson/Encode/Functions.hs
+++ b/Data/Aeson/Encode/Functions.hs
@@ -3,12 +3,11 @@
 module Data.Aeson.Encode.Functions
     (
       builder
-    , char7 -- TODO: used by TH module
     ) where
 
 import Data.Aeson.Encoding
 import Data.Aeson.Types.Class
-import Data.ByteString.Builder (Builder, char7)
+import Data.ByteString.Builder (Builder)
 import Data.Monoid ((<>))
 
 builder :: ToJSON a => a -> Builder
--- a/Data/Aeson/Encode/Functions.hs
+++ b/Data/Aeson/Encode/Functions.hs
@@ -2,22 +2,14 @@
 
 module Data.Aeson.Encode.Functions
     (
-      brackets
-    , builder
-    , builder'
-    , char7
+      builder
+    , char7 -- TODO: used by TH module
     , encode
-    , list
-    , encodeMap
-    , encodeWithKey
     ) where
 
-import Data.Aeson.Encode.Builder
+import Data.Aeson.Encoding
 import Data.Aeson.Types.Class
-import Data.Aeson.Types.Internal
 import Data.ByteString.Builder (Builder, char7)
-import Data.ByteString.Builder.Prim (primBounded)
-import Data.Foldable (toList)
 import Data.Monoid ((<>))
 import qualified Data.ByteString.Builder as B
 import qualified Data.ByteString.Lazy as L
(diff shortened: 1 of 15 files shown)


Data/Aeson/Encode/Functions.hs
```diff
@@ -2,78 +2,15 @@
 
 module Data.Aeson.Encode.Functions
```
phadej: After `.Instances` is refactored also, this module becomes obsolete.

bergmark:
This does look a lot cleaner, thanks!

Some notes:

* It would be nice to also have some examples on how to write encoding-based instances. Are we exposing enough for users to use this? #301 is related.
* #301 also mentions that we should remove the Semigroup/Monoid instances.
* There are still TODOs in the code (mostly a reminder to myself before merging)
* Having both `Data.Aeson.Encode` and `Data.Aeson.Encoding` is confusing. Do we need both?

phadej:
@bergmark 
1. I'd plan to rewrite `.Instances` without using `Builder` to act as an examples, would also update the documentation as well
1. No Semigroup/Monoid is valid concern. OTOH e.g. cbor's [`Encoding is also Monoid`](https://github.com/well-typed/binary-serialise-cbor/blob/master/Data/Binary/Serialise/CBOR/Encoding.hs`). My gut feeling that you'd occasionally need to concat `Encoding`s, and then you may or may not break it. Point of this PR is to make it much harder to produce `Encoding` producing invalid token stream, this would be an improvement.
1. yes, I added three TODOs myself. Two export related will be gone after follow-up PR, and third one (about date/time hack) will vanish around that time as well: we'll need to do something smarter to write `ToJSONKey`/`FromJSONKey` instances for those datatypes.
1. [`Data.Aeson.Encode`](https://hackage.haskell.org/package/aeson-0.11.2.0/docs/Data-Aeson-Encode.html) has
- `encode` and `encodeToBuilder` could be moved to `Data.Aeson.Encoding`
- `encodeToTextBuilder` could be then moved to either `Data.Aeson.Encoding.Text` or `Data.Aeson.Text`, I'd like the latter better, as we could add decoding from `Text` functions too.

My proposal for merging:
- merge this
- fix https://github.com/bos/aeson/issues/375
- continue to polish rough edges.

phadej:
ping @andrewthad, above comment ^

andrewthad:
First of all, nice work! I think that this is definitely a safer approach to dealing with `Encoding`.

At some point in the future (not necessarily now), it would be neat to make it impossible to construct an invalid encoding with the functions from `Data.Aeson.Encoding`. As mentioned in multiple comments above, this would require eliminating the `Semigroup` and `Monoid` instances. It would also mean not exporting [comma, colon, openBracket, etc.](https://github.com/phadej/aeson/blob/5a0faf747a4e38ad8b51474c775f350d5f91f166/Data/Aeson/Encoding.hs#L23) and not exporting `>*<`. It seems like this might be a fairly involved process though since the TH stuff uses these. Maybe a less invasive approach would be to split `Data.Aeson.Encoding` into two modules, one that is completely safe and another that allows you to build invalid `Encoding`s. Sorry I ended up rambling there. It's definitely not a pressing issue, but I just wanted to record my thoughts somewhere.

I'm completely in favor of the merge proposal. Concerning https://github.com/bos/aeson/issues/375, my preference would be to have two modules (one for the `ToJSON` stuff and one for the `FromJSON` stuff) with a third one that simply re-exports both. If @bergmark agrees, I would say to go ahead and do the module merging.

phadej:
I'm :+1: on splitting out `Data.Aeson.Encoding.Internal`, but would like to do it later?

bergmark:
Alright let's start by merging this.

PR #413, opened 2016-06-06
Un-orphan instances

phadej:
Resolves https://github.com/bos/aeson/issues/375


phadej:
@andrewthad @bergmark this is ready for review

Data/Aeson/Types/Generic.hs
```diff
@@ -1,8 +1,13 @@
...
+{-# LANGUAGE DataKinds              #-}
+{-# LANGUAGE EmptyDataDecls         #-}
+{-# LANGUAGE FlexibleInstances      #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE KindSignatures         #-}
+{-# LANGUAGE MultiParamTypeClasses  #-}
+{-# LANGUAGE ScopedTypeVariables    #-}
+{-# LANGUAGE TypeOperators          #-}
+{-# LANGUAGE UndecidableInstances   #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
```
andrewthad: We can get rid of `-fno-warn-orphans`.

phadej: Done

andrewthad:
I've given it a cursory glance, and this looks how I would expect it to. I'm +1 on merging into master. There's one place where you left an `-fno-warn-orphans` flag on that I think can be removed though.

PR #414, opened 2016-06-07
Generics and TH support for lifted typeclasses

RyanGlScott:
Because boilerplate makes me sad.

This adds the capability to automatically derive:

* `ToJSON1`, `ToJSON2`, `FromJSON1`, and `FromJSON2` with Template Haskell
* `ToJSON1` and `FromJSON1` via GHC generics (sadly, generics isn't expressive enough to cover `ToJSON2` and `FromJSON2`)

This code is somewhat old and had to rebased several times recently, so @andrewthad, @phadej, please make sure I didn't accidentally trample something you wrote.

### DIFF
--- a/Data/Aeson.hs
+++ b/Data/Aeson.hs
@@ -77,9 +77,14 @@ module Data.Aeson
     , GFromJSON(..)
     , GToJSON(..)
     , GToEncoding(..)
+    , Zero
+    , One
     , genericToJSON
+    , genericLiftToJSON
     , genericToEncoding
+    , genericLiftToEncoding
     , genericParseJSON
+    , genericLiftParseJSON
     , defaultOptions
 
     -- * Inspecting @'Value's@
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -1,5 +1,9 @@
-{-# LANGUAGE CPP, FlexibleInstances, NamedFieldPuns,
-    NoImplicitPrelude, UndecidableInstances #-}
+{-# LANGUAGE BangPatterns #-}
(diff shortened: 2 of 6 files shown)


phadej:
Unfortunately https://github.com/bos/aeson/pull/413 moves everything around, but I can rebase this after moving is merged; could review then on the way too.

andrewthad:
I'm not really qualified to review it, but this looks good from a cursory glance. I think it may be easier to rebase (or more accurately, redo by means of coping and pasting) https://github.com/bos/aeson/pull/413 on top of this branch than it would be to rebase this branch on top of https://github.com/bos/aeson/pull/413.

Data/Aeson.hs
```diff
@@ -77,9 +77,14 @@ module Data.Aeson
     , GFromJSON(..)
     , GToJSON(..)
     , GToEncoding(..)
+    , Zero
```
phadej: Does one need to export these from `Data.Aeson` ?

RyanGlScott: "Need" is a strong word, but if `GFromJSON`/`GToJSON`/`GToEncoding` are exported from `Data.Aeson`, then I feel it makes sense to export `Zero`/`One` too, since it would now be pretty much impossible to write useful type signatures involving the former without using the latter (since you need things like `GToJSON Zero (Rep a)`).

phadej: Makes sense

phadej:
@andrewthad I disagree: https://github.com/bos/aeson/pull/415

andrewthad:
@phadej Dang. You did that way faster than I thought was possible.

Issue #401, opened 2016-05-25
Instances for DList


phadej:
`aeson` depends on it, so it should provide instances as well.

PR #417, opened 2016-06-07
Add DList instances and some roundtrip key tests

phadej:
Resolve https://github.com/bos/aeson/issues/401

These should be straightforward


andrewthad:
Quick question. In the `FromJSONKey` instance for `Bool`:

    fromJSONKey = FromJSONKeyTextParser $ \t -> case t of
        "true"  -> pure True
        "false" -> pure False
        _       -> fail $ "Cannot parse key into Bool: " ++ T.unpack t

How are you pattern matching on `Text`? I've never seen this before.

phadej:
@andrewthad the same way as e.g. with `Num`, using `Eq` instance:
```
λ Prelude > :set -XOverloadedStrings 
λ Prelude > :t \x -> case x of { "" -> True; _ -> False }
\x -> case x of { "" -> True; _ -> False }
  :: (Data.String.IsString a, Eq a) => a -> Bool
```

bergmark:
@andrewthad -XOverloadedStrings allows this.

andrewthad:
Good to know.

PR #418, opened 2016-06-08
Document json key

andrewthad:
Still a work in progress (I haven't added docs for `FromJSONKey` yet), but I thought I would go ahead and put it up. Any feedback is welcome.


Data/Aeson/Types/ToJSON.hs
```diff
@@ -318,17 +318,102 @@ instance KeyValue Pair where
...
+--   It is possible to get the 'ToJSONKey' instance for free as we did
+--   with 'Foo'. However, in this case, we have a natural way to go to
+--   and from 'Text' that does not require any escape sequences. So, in
+--   this example, 'ToJSONKeyText' will be used instead of 'ToJSONKeyValue'.
+--   At the moment, there are no conveniences for this situation. The
+--   'Show' instance can be used to help write 'ToJSONKey', but this will 
+--   be inefficient since it uses 'String' as an intermediate step:
+--
+--   > instance ToJSONKey Color where
+--   >   toJSONKey = ToJSONKeyText f g
```
phadej: In https://github.com/bos/aeson/pull/419/files#diff-69a0e3f150624a735a43f70eb5ae9c00R1827 I define helper to define `toJSONKey` for textual keys using `toJSONKeyTextEnc`, we should provide similar helper `toJSONKeyText :: (a -> Text) -> ToJSONKeyFunction a` and export this.

i.e.

> be ineffiecient since it uses 'String' as an intermediate step:
```hs
instance ToJSONKey Color where
    toJSONKey = ToJSONKeyText f g
      where f = Text.pack . show
            g = text . Text.pack . show 
            -- text function is from Data.Aeson.Encoding
```
> or more conveniently
```hs
instance ToJSONKey Color where
    toJSONKey = toJSONKeyText (Text.pack . show)
```

andrewthad: I agree. That's a common enough enough pattern that it warrants a function to help with it. Additionally, I think I'll point people to using something like [the deriving machinery for TextShow](http://hackage.haskell.org/package/text-show-3.2.2/docs/TextShow-TH.html) to get a function of type `a -> Text`. I suspect `showt` should outperform `Text.pack . show`, but I need to benchmark it to be sure. (Just to be clear, I'm only talking about recommending the `text-show` package in `aeson`'s docs, not bringing it in as a dependency). @RyanGlScott Since you wrote `text-show`, you might already know about what kind of performance difference exists.

RyanGlScott: I do have a [small benchmark](https://github.com/RyanGlScott/text-show/blob/b49cd9444bf8c057517f8ea7057bbc5bd6954615/benchmarks/Bench.hs) in `text-show` which compares the performance of calling `Text.pack . show` versus using `text-show`'s functionality. On a [recent benchmark run](https://travis-ci.org/RyanGlScott/text-show/jobs/136144077#L1537), I obtained the following results:

```
benchmarking String Show, then Text.pack/Small sample
time                 4.235 μs   (4.215 μs .. 4.257 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 4.245 μs   (4.221 μs .. 4.284 μs)
std dev              102.3 ns   (70.67 ns .. 135.3 ns)
variance introduced by outliers: 28% (moderately inflated)

benchmarking String Show, then Text.pack/Medium sample
time                 8.068 ms   (8.028 ms .. 8.130 ms)
                     0.999 R²   (0.996 R² .. 1.000 R²)
mean                 8.107 ms   (8.053 ms .. 8.187 ms)
std dev              185.4 μs   (96.48 μs .. 288.5 μs)

benchmarking String Show, then Text.pack/Large sample
time                 1.003 s    (736.8 ms .. 1.291 s)
                     0.990 R²   (0.964 R² .. 1.000 R²)
mean                 1.041 s    (1.009 s .. 1.072 s)
std dev              53.56 ms   (0.0 s .. 54.26 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking Text Show (TH)/Small sample
time                 1.775 μs   (1.768 μs .. 1.785 μs)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 1.827 μs   (1.801 μs .. 1.876 μs)
std dev              119.4 ns   (64.69 ns .. 194.0 ns)
variance introduced by outliers: 76% (severely inflated)

benchmarking Text Show (TH)/Medium sample
time                 3.094 ms   (3.037 ms .. 3.172 ms)
                     0.997 R²   (0.996 R² .. 0.999 R²)
mean                 3.102 ms   (3.071 ms .. 3.136 ms)
std dev              107.5 μs   (87.66 μs .. 129.3 μs)
variance introduced by outliers: 17% (moderately inflated)

benchmarking Text Show (TH)/Large sample
time                 360.0 ms   (346.2 ms .. 376.6 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 373.3 ms   (368.0 ms .. 378.0 ms)
std dev              7.535 ms   (67.99 as .. 8.053 ms)
variance introduced by outliers: 19% (moderately inflated)
```

So your hunch was correct: `text-show` is indeed more performant (at least in this scenario in the benchmark).

phadej:
@andrewthad see *Refactor ToJSONKeyFunction, add some docs* commit in #422

PR #421, opened 2016-06-08
Data.Time keys, and almost done Encoding refactoring

phadej:
Sorry for PR spam

### DIFF
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -174,6 +174,9 @@ test-suite tests
   if !impl(ghc >= 7.10)
     build-depends: nats >=1 && <1.2
 
+  if impl(ghc >= 7.8)
+    build-depends: hashable-time >= 0.2 && <0.3
+
 source-repository head
   type:     git
   location: git://github.com/bos/aeson.git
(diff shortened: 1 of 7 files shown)


bergmark:
Should this be closed in favor of #422?

phadej:
There is a cascade of PRs, sorry for that. I guess it's easier to review in steps, as #422 is quite large at this point. Also it seems to conflict, when this doesn't.

aeson.cabal
```diff
@@ -174,6 +174,9 @@ test-suite tests
   if !impl(ghc >= 7.10)
     build-depends: nats >=1 && <1.2
 
+  if impl(ghc >= 7.8)
+    build-depends: hashable-time >= 0.2 && <0.3
+
```
bergmark: What does this mean for older GHC's, and why is it only needed in the test-suite?

phadej: `hashable-time` hss restrictive bounds atm, so some tests are run only when it's available. I could pit Ot behind a automatic flag too, but then it will be harder to spot that tests aren't run at all

bergmark: Ok thanks, sounds good to leave it as is then.

Issue #416, opened 2016-06-07
Change definition of ToJSONKeyFunction


phadej:
```hs
data ToJSONKeyFunction a
    = ToJSONKeyText (a -> Text, a -> Encoding)
    | ToJSONKeyValue (a -> Value, a -> Encoding)
```
to
```hs
data ToJSONKeyFunction a
    = ToJSONKeyText !(a -> Text) !(a -> Encoding)
    | ToJSONKeyValue !(a -> Value) !(a -> Encoding)
```

i.e. unwrap tuple and have strict fields

cc @andrewthad

andrewthad:
Yeah, that makes way more sense. I have no idea why I used a tuple there.

phadej:
Ok, let's change the definition after currently-open PRs are merged/closed.

phadej:
This is done

PR #422, opened 2016-06-08
Encoding refactoring final

phadej:
The last commit particularly makes encoding a tagged type (Maybe we could have `TaggedEncoding tag` and `type Encoding = TaggedEncoding Value`)

There are only few tags (`Value`, `Text` and `T`), but more can be introduced if we feel so, none of `Instances` non-generic code need anything else (in particular tags could help to track what kind of Encoding we are assembling in generic code - dunno if that helps, cc @RyanGlScott ).


phadej:
*Refactor ToJSONKeyFunction, add some docs* resolves #416

bergmark:
Looks like an actual test failure on travis

phadej:
force pushed, let's see.

bergmark:
This needs a rebase now

phadej:
rebased

PR #423, opened 2016-06-08
Tell which key failed to parse in maps

phadej:
### Before

```hs
> eitherDecode "[{\"x\": 2}]" :: Either String [Map Int Int]
Left "Error in $[0]: input does not start with a digit"
```

### After

```hs
Left "Error in $[0].x: input does not start with a digit"
```


bergmark:
Great thanks :-)

PR #426, opened 2016-06-09
Allow building with GHC 7.4/7.6's bundled bytestring using bytestring-builder

RyanGlScott:
Currently, `aeson` requires `bytestring >= 0.10.4` due to its use of `Data.ByteString.Builder`. This poses a problem for GHC 7.4 and 7.6, which are shipped with older versions of `bytestring`, and as a result must manually install more recent `bytestring`s to work with `aeson`.

Doing so can cause other problems, though. For instance, it can force the `unix` package to be reinstalled (since the `unix` that's shipped with GHC 7.4/7.6 isn't compatible with newer `bytestring`, but if a `Cabal` script uses the version of `Cabal` that's bundled with GHC and tries to link against the reinstalled `unix`, it'll produce linker errors due to `Cabal` itself being linked against the older `unix`. This isn't a hypothetical scenario: this has been observed [here](https://travis-ci.org/lambda-llama/bitset/jobs/136297705#L312).

Anyway, the fix is simple. We allow `aeson` to build with older `bytestring`s by conditionally depending on the `bytestring-builder` package, which backports `Data.ByteString.Builder` when necessary. This has the result of making `aeson`-related build plans much saner with GHC 7.4 and 7.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,12 @@ before_cache:
 
 matrix:
   include:
-    - env: BUILD=cabal CABALVER=1.16 GHCVER=7.4.2
+    - env: BUILD=cabal CABALVER=1.18 GHCVER=7.4.2
       compiler: ": #GHC 7.4.2"
-      addons: {apt: {packages: [cabal-install-1.16,ghc-7.4.2], sources: [hvr-ghc]}}
-    - env: BUILD=cabal CABALVER=1.16 GHCVER=7.6.3
+      addons: {apt: {packages: [cabal-install-1.18,ghc-7.4.2], sources: [hvr-ghc]}}
+    - env: BUILD=cabal CABALVER=1.18 GHCVER=7.6.3
       compiler: ": #GHC 7.6.3"
-      addons: {apt: {packages: [cabal-install-1.16,ghc-7.6.3], sources: [hvr-ghc]}}
+      addons: {apt: {packages: [cabal-install-1.18,ghc-7.6.3], sources: [hvr-ghc]}}
     - env: BUILD=cabal CABALVER=1.18 GHCVER=7.8.4
       compiler: ": #GHC 7.8.4"
       addons: {apt: {packages: [cabal-install-1.18,ghc-7.8.4], sources: [hvr-ghc]}}
--- a/Data/Aeson/Encoding/Builder.hs
+++ b/Data/Aeson/Encoding/Builder.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE BangPatterns, OverloadedStrings #-}
+{-# LANGUAGE BangPatterns, CPP, OverloadedStrings #-}
 
 -- |
 -- Module:      Data.Aeson.Encoding.Builder
@@ -52,8 +52,21 @@ import Data.Time.LocalTime
 import Data.Word (Word8)
 import qualified Data.HashMap.Strict as HMS
 import qualified Data.Text as T
-import qualified Data.Text.Encoding as TE
 import qualified Data.Vector as V
+
+#if MIN_VERSION_bytestring(0,10,4)
+import Data.Text.Encoding (encodeUtf8BuilderEscaped)
+#else
+import Data.Bits ((.&.))
+import Data.Text.Internal (Text(..))
+import Data.Text.Internal.Unsafe.Shift (shiftR)
+import Foreign.Ptr (minusPtr, plusPtr)
+import Foreign.Storable (poke)
+import qualified Data.ByteString.Builder.Internal as B
+import qualified Data.ByteString.Builder.Prim.Internal as BP
+import qualified Data.Text.Array as A
+import qualified Data.Text.Internal.Encoding.Utf16 as U16
+#endif
 
 -- | Encode a JSON value to a "Data.ByteString" 'B.Builder'.
 --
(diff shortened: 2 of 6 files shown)


RyanGlScott:
Ugh, unfortunately I don't think the Travis build will succeed until https://github.com/basvandijk/scientific/pull/41 is merged and released.

phadej:
I had this idea too, would be great!

RyanGlScott:
Well, crud. This isn't as straightforward as I originally thought, since I didn't realize that the `text` library doesn't define the [`encodeUtf8BuilderEscaped`](http://hackage.haskell.org/package/text-1.2.2.1/docs/Data-Text-Encoding.html#g:6) function unless built against `bytestring >= 0.10.4.0`.

The way I see it, we have three options:

1. Make a PR for `text` to request that it use `bytestring-builder`. This might be difficult, since `text` is such a widely used library that has minimal dependencies.
2. Make a PR for `bytestring-builder` to backport `encodeUtf8BuilderEscaped`. This might be considered scope creep, but it would address the dependency issue.
3. Conditionally redefine `encodeUtf8BuilderEscaped` in `aeson` itself. This would require making sure the two definitions remain in sync, but it wouldn't require changing any other libraries.

Thoughts?

phadej:
`text` will probably be dependency on `Cabal`, thru `parsec` so option 1 is no go. How `bytestring-builder` is `bytestring` compat package; yet very specific: there aren't `BSL.toStrict` and `IsString` instances to mention a few. So the option 2 is a bit problematic too. I don't think we have to track/copy definition `encodeUtf8EscapedBuilder` as far as it's not much slower and produces same results, so I'd go with third option. Another question, is if `text` becomes core package and it's API vary a lot across releases / dependency versions, it might make sense to have `text-compat`, yet it would need a maintainer.

RyanGlScott:
Very well, I'll proceed with option 3.

Unfortunately, there's another blocker: there's no `Semigroup` instance for the `Builder` type from `bytestring-builder`. I've opened https://github.com/ekmett/semigroups/pull/68 to fix this.

lpsmith:
Incidentally,  I made a PR against `text` once upon a time to do exactly this,  for probably very similar reasons.

RyanGlScott:
(For reference: https://github.com/bos/text/pull/67)

So option 1 looks very unlikely :)

lpsmith:
Well,  I think it's worth another attempt :)

RyanGlScott:
`semigroups-0.18.2` has been released, which supports `bytestring-builder`, so I think this PR should work now. Let me close and re-open to trigger another Travis build.

RyanGlScott:
Hm, a couple of oddities with the Travis build.

One thing is that the `stack` build refuses to proceed. I suspect that this just may be due to whatever resolver it's using (nightly?) not having `semigroup-0.18.2`. The [error message](https://travis-ci.org/bos/aeson/jobs/139775781#L233) would seem to indicate as such, anyway, so I don't think there's anything we can do about that but wait.

The more annoying problem is that the GHC 7.4 and 7.6 builds fail. This is especially aggravating since I've replicated the exact steps that Travis is doing locally on my machine, and it works fine. I can only guess that Travis has cached some dependencies from previous builds that are sneaking in where they shouldn't. For example, in the GHC 7.4 build `cabal configure` [shows that `bytestring-0.10.8.1` is being used](https://travis-ci.org/bos/aeson/jobs/139775775#L309) but `bytestring-builder` isn't, which is most definitely not what we want! Can someone with the right permissions delete the Travis cache for my PR so that I can try again from scratch?

bergmark:
i cleared the cache and restarted the build

RyanGlScott:
Sorry, I think I misdiagnosed the issue. Upon taking a closer look at the Travis logs, it looks like `cabal-install-1.16` (a relatively old version) is incorrectly picking `-f-bytestring-builder` for the GHC 7.4 and 7.6 builds. This might just be an old `cabal-install` bug, so I'll try bumping the `cabal-install` versions in the `.travis.yml` file to 1.18 to see if that helps things.

RyanGlScott:
Yep, using `cabal-install-1.18` does the trick.

There was one last hiccup I had to solve in that `aeson` uses the function `Data.ByteString.Lazy.toStrict`, which didn't exist prior to `bytestring-0.10.0.0`. I took the most direct route and defined a `lazyToStrictByteString` function, which is defined as `toStrict` if a recent-enough `bytestring` is used, and otherwise backports a definition that I took from [an internal function](http://hackage.haskell.org/package/blaze-builder-0.4.0.1/docs/src/Blaze-ByteString-Builder.html#line-125) from the `blaze-builder` library.

Alternatively, I could define an `encodingToStrictByteString` function to use in place of `toStrict . encodingToLazyByteString` that's defined in terms of [`toByteString`](http://hackage.haskell.org/package/blaze-builder-0.4.0.1/docs/src/Blaze-ByteString-Builder.html#toByteString) from `blaze-builder`, but that would require incurring an extra dependency... plus, I'm not sure if new packages should be using `blaze-builder` anyway.

RyanGlScott:
Also, the `stack` failure is due to `semigroups-0.18.2` [being held back on Stackage](https://github.com/fpco/stackage/issues/1645) since `binary-orphans` depends on `semigroups < 0.18.2`... @phadej? :)

phadej:
I'm on holiday till Wednesday (only on mobile, it is). I hope this can wait till I'm back :)

RyanGlScott:
Thanks, @phadej!

I went ahead and opened https://github.com/bos/text/pull/155 to try and make `text` depend on `bytestring-builder`... we'll see where it goes. :)

PR #427, opened 2016-06-09
Issue 406 base compat

### LINKED ISSUE #406: Use base-compat and time-locale-compat to reduce CPP
Opinions?

phadej:
Resolve #406

### DIFF
--- a/Data/Aeson.hs
+++ b/Data/Aeson.hs
@@ -108,6 +108,9 @@ module Data.Aeson
     , json'
     ) where
 
+import Prelude ()
+import Prelude.Compat (Maybe, either, Either(..), uncurry, String, (.))
+
 import Data.Aeson.Types.FromJSON (ifromJSON)
 import Data.Aeson.Encoding (encodingToLazyByteString)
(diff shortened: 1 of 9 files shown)


bergmark:
I think we should have the Prelude.Compat import in every module, then we never have to think about this again.

phadej:
@bergmark what about having `NoImplicitPrelude` in `aeson.cabal` ?

bergmark:
If we do that you can no longer use plain `ghci` to load the project. `cabal repl` used to not take this into account as well IIRC (has it been fixed?). `stack ghci` does take this into account, but I have not tested how it interacts with loading multiple packages.

phadej:
force pushed, `Prelude.Compat` is in every module now

RyanGlScott:
BTW, I don't think you need to enable `LANGUAGE CPP` in `Data.Aeson.Parser.Time` anymore.

phadej:
@RyanGlScott noted

phadej:
rebased

PR #435, opened 2016-06-23
Remove use of Proxy and undefined in generics code

RyanGlScott:
In order to facilitate the use of `FromJSON1` and `ToJSON1`, I introduced some infelicitous code in #414. Namely, `gToJSON`/`gToEncoding`/`gFromJSON` now requires passing around a `Proxy` argument and two `undefined` values everywhere. To make this a little nicer, I introduce two new data types, `ToArgs` and `FromArgs`, which subsume the need for `Proxy`s, and only contain function arguments if necessary.


bergmark:
LGTM Thanks! 🎉

PR #442, opened 2016-07-04
Fix #317: Do not truncate or under/overflow non-integral numbers

### LINKED ISSUE #317: Do not `truncate` `Integral` values
aeson-0.10.0.0 `truncate`s `Integral` values like `Int`, `Integer`, etc. when converting from JSON:
```
instance FromJSON Integer where
    parseJSON = withScientific "Integral" $ pure . truncate

[...]
 
parseIntegral :: Integral a => String -> Value -> Parser a
parseIntegral expected = withScientific expected $ pure . truncate
```

I cannot test the latest version right now, but here's 0.9.0.1, which uses `floor`:
```
Prelude Data.Aeson> decode "3.14" :: Maybe Int
Just 3
```

The expected result is `Nothing` (or maybe an error message warning about the extra `.14` bit).

phadej:


Data/Aeson/Types/FromJSON.hs
```diff
@@ -178,18 +178,46 @@ parseRealFloat _        Null       = pure (0/0)
...
+
 parseIntegral :: Integral a => String -> Value -> Parser a
-parseIntegral expected = withScientific expected $ pure . truncate
+parseIntegral expected = withScientific expected $ truncateIntegral expected
 {-# INLINE parseIntegral #-}
 
+truncateBoundedIntegral :: Integral a => String -> Scientific -> Parser a
+truncateBoundedIntegral expected s
+    | Scientific.isInteger s =
+        let x = truncate s
```
basvandijk: Why use `truncate` here? If we've already determined `s` is integer truncating will not have an effect anymore.

I would recommend using [toBoundedIntege](http://hackage.haskell.org/package/scientific-0.3.4.8/docs/Data-Scientific.html#v:toBoundedInteger) here.

phadej: TIL about `toBoundedInteger`. Thanks!

phadej:
Yeah, let's use only `toBoundedInteger`, the reason should be more-or-less obvious from the value itself.

Will amend ASAP

Data/Aeson/Types/FromJSON.hs
```diff
@@ -178,18 +178,42 @@ parseRealFloat _        Null       = pure (0/0)
 parseRealFloat expected v          = typeMismatch expected v
 {-# INLINE parseRealFloat #-}
 
+truncateIntegral :: Integral a => String -> Scientific -> Parser a
+truncateIntegral expected s
+    | Scientific.isInteger s = pure (truncate s)
```
basvandijk: Here a similar comment applies as my previous comment. Instead of using both `isInteger` and `truncate` you can use the single [floatingOrInteger](http://hackage.haskell.org/package/scientific-0.3.4.8/docs/Data-Scientific.html#v:floatingOrInteger) which should be slightly more efficient and shorter.

Data/Aeson/Types/FromJSON.hs
```diff
@@ -178,11 +178,11 @@ parseRealFloat _        Null       = pure (0/0)
 parseRealFloat expected v          = typeMismatch expected v
 {-# INLINE parseRealFloat #-}
 
-truncateIntegral :: Integral a => String -> Scientific -> Parser a
-truncateIntegral expected s
-    | Scientific.isInteger s = pure (truncate s)
-    | otherwise               =
-        fail $ "Floating number specified for " ++ expected ++ ": " ++ show s
+truncateIntegral :: forall a. Integral a => String -> Scientific -> Parser a
```
basvandijk: Thanks for the changes! Now, I would just rename this function to something else since it's not really truncating anymore.

phadej: It's not exposed, so it's not that big deal. But will do if there are good versions. `parseIntegralFromScientific` (and `parseBoundedIntegralFromScientific`)?

basvandijk: Sounds good!

bergmark:
Sweet, thanks @phadej! And thank you for the review @basvandijk!

Issue #437, opened 2016-06-29
Can't decode to records with multiple value constructors


bgwines:
Hi, there! Let's say you have a record

```{.haskell}
data DataType
    = A { a :: Int }
    | B { b :: Bool }
```

I haven't been able to successfully decode to such a data type using Aeson. Not sure if this a bug or a feature request.

I've put up an isolated 100% reproducible test case for you at https://github.com/bgwines/aeson-repro

Thanks for your attention to the matter! :)

phadej:
sum types are encoded (by default) with `tag` field to *tag the constructor*:
```hs
λ > A.encode $ User "foo" "bar" Nothing
"{\"email\":\"bar\",\"username\":null,\"name\":\"foo\"}"
λ > A.encode $ SmallUser "foo" "bar" Nothing
"{\"email\":\"bar\",\"tag\":\"SmallUser\",\"username\":null,\"name\":\"foo\"}"
```

Currently there isn't a way to deriving untagged serialisation generically, though it could be possible to add `UntaggedObject` to [`SumEncoding`](http://hackage.haskell.org/package/aeson-0.11.2.0/docs/Data-Aeson-TH.html#t:SumEncoding). We can do it for `aeson-1` (the next major release). Opinions @bergmark ? I remember I needed similar functionality in the past too. 

So the the derived instance would be something like
```hs
instance FromJSON DataType where
    parseJSON v = parseA v <|> parseB v
      where
      	parseA = withObject "A" $ \obj -> A <$> obj .: "a"
      	parseB = withObject "B" $ \obj -> B <$> obj .: "b" 
```

bergmark:
Sure this is a reasonable thing to have, I've used similar instances.

PR #443, opened 2016-07-04
Add UntaggedValue to SumEncoding

phadej:
Resolves https://github.com/bos/aeson/issues/437

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -549,6 +549,20 @@ data SumEncoding =
     -- by the encoded value of that field! If the constructor is not a
     -- record the encoded constructor contents will be stored under
     -- the 'contentsFieldName' field.
+  | UntaggedValue
+    -- ^ Constructor names won't be encoded. Instead only the contents of the
+    -- constructor will be encoded as if the type had single constructor. JSON
+    -- encodings have to be disjoint for decoding to work properly.
+    --
+    -- When decoding, constructors are tried in the order of definition. If some
+    -- encodings overlap, the first one defined will succeed.
+    --
+    -- /Note:/ Nullary constructors are encoded as the string (using
+    -- 'constructorTagModifier'). Having a nullary constructor alongside a
+    -- single field constructor that encodes to a string leads to ambiguity.
+    --
+    -- /Note:/ Only the last error is kept when decoding, so in the case of
+    -- mailformed JSON, only an error for the last constructor will be reported.
   | ObjectWithSingleField
     -- ^ A constructor will be encoded to an object with a single
     -- field named after the constructor tag (modified by the
--- a/tests/Properties.hs
+++ b/tests/Properties.hs
@@ -172,6 +172,16 @@ isObjectWithSingleField :: Value -> Bool
 isObjectWithSingleField (Object obj) = H.size obj == 1
 isObjectWithSingleField _            = False
 
+-- | is untaggedValue of EitherTextInt
+isUntaggedValueETI :: Value -> Bool
+isUntaggedValueETI (String s)
+    | s == "nonenullary"   = True
+isUntaggedValueETI (Bool _)   = True
+isUntaggedValueETI (Number _) = True
+isUntaggedValueETI (Array a)  = length a == 2
+isUntaggedValueETI _          = False
+
+
 --------------------------------------------------------------------------------
 
 
@@ -293,6 +303,10 @@ tests = testGroup "properties" [
             , testProperty "ObjectWithSingleField" (toParseJSON gNullaryParseJSONObjectWithSingleField gNullaryToJSONObjectWithSingleField)
             ]
         ]
+      , testGroup "EitherTextInt" [
+          testProperty "UntaggedValue" (isUntaggedValueETI . gEitherTextIntToJSONUntaggedValue)
+        , testProperty "roundtrip" (toParseJSON gEitherTextIntParseJSONUntaggedValue gEitherTextIntToJSONUntaggedValue)
+        ]
       , testGroup "SomeType" [
           testProperty "2ElemArray" (is2ElemArray . gSomeTypeToJSON2ElemArray)
         , testProperty "TaggedObject" (isTaggedObject . gSomeTypeToJSONTaggedObject)
@@ -324,6 +338,9 @@ tests = testGroup "properties" [
       --   gApproxToJSONUnwrap `sameAs` gApproxToEncodingUnwrap
       , testProperty "ApproxDefault" $
         gApproxToJSONDefault `sameAs` gApproxToEncodingDefault
+
+      , testProperty "EitherTextInt UntaggedValue" $
+        gEitherTextIntToJSONUntaggedValue `sameAs` gEitherTextIntToEncodingUntaggedValue
 
       , testProperty "SomeType2ElemArray" $
         gSomeTypeToJSON2ElemArray `sameAs` gSomeTypeToEncoding2ElemArray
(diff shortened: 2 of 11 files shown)


phadej:
ping @RyanGlScott could you review this?

tests/Properties.hs
```diff
@@ -293,6 +301,10 @@ tests = testGroup "properties" [
             , testProperty "ObjectWithSingleField" (toParseJSON gNullaryParseJSONObjectWithSingleField gNullaryToJSONObjectWithSingleField)
             ]
         ]
+      , testGroup "EitherTextInt" [
+        --  testProperty "UntaggedValue" (isUntaggedValue . gEitherTextIntToJSONUntaggedValue)
+        --, testProperty "roundtrip" (toParseJSON gEitherTextIntParseJSONUntaggedValue gEitherTextIntToJSONUntaggedValue)
```
RyanGlScott: Why are some of these tests commented out?

phadej: generics aren't yet implemented

RyanGlScott: Oh, of course. I don't think adding generics support would be too complicated, especially since `UntaggedValue` is conceptually very simple. I don't have time to concoct something right now, but if you want to take a shot at it I can review later.

phadej:
Handling nullary constructors is a bit tricky, but seems to work:

```hs
λ > :t $(mkParseJSON defaultOptions { sumEncoding = UntaggedValue } ''Maybe)
<interactive>:1:3-68: Splicing expression
    mkParseJSON (defaultOptions {sumEncoding = UntaggedValue}) ''Maybe
  ======>
    \ value_aFnV
      -> ((case value_aFnV of {
             String str_aFnW | (str_aFnW == (pack "Nothing")) -> pure Nothing
             other_aFnX
               -> Data.Aeson.TH.parseTypeMismatch'
                    "Nothing"
                    "GHC.Base.Maybe"
                    "String"
                    (Data.Aeson.TH.valueConName other_aFnX) })
          GHC.Base.<|>
            (case value_aFnV of {
               arg_aFnY -> (Just <$> (parseJSON arg_aFnY)) }))
$(mkParseJSON defaultOptions { sumEncoding = UntaggedValue } ''Maybe)
  :: FromJSON a => Value -> Parser (Maybe a)
λ > :t $(mkToJSON defaultOptions { sumEncoding = UntaggedValue } ''Maybe)
<interactive>:1:3-65: Splicing expression
    mkToJSON (defaultOptions {sumEncoding = UntaggedValue}) ''Maybe
  ======>
    \ value_aFp0
      -> case value_aFp0 of {
           Nothing -> String (pack "Nothing")
           Just arg1_aFp1 -> toJSON arg1_aFp1 }
$(mkToJSON defaultOptions { sumEncoding = UntaggedValue } ''Maybe)
  :: ToJSON a => Maybe a -> Value
```

phadej:
with *Better mkParseJSON...* commit:

```hs
 > :t $(mkParseJSON defaultOptions { sumEncoding = UntaggedValue } ''Maybe)
<interactive>:1:3-68: Splicing expression
    mkParseJSON (defaultOptions {sumEncoding = UntaggedValue}) ''Maybe
  ======>
    \ value_aS3z
      -> ((case value_aS3z of {
             String str_aS3A | (str_aS3A == (pack "Nothing")) -> pure Nothing })
          GHC.Base.<|>
            ((case value_aS3z of {
                arg_aS3B -> (Just <$> (parseJSON arg_aS3B)) })
             GHC.Base.<|> (typeMismatch "GHC.Base.Maybe" value_aS3z)))
$(mkParseJSON defaultOptions { sumEncoding = UntaggedValue } ''Maybe)
  :: FromJSON a => Value -> Parser (Maybe a)
```

phadej:
Ready for review.

tests/UnitTests.hs
```diff
@@ -108,6 +108,18 @@ camelFrom c s = let (p:ps) = split c s
...
+
+data MyEither a b = MyLeft a | MyRight b
+  deriving (Generic, Show, Eq)
+
+instance (ToJSON a, ToJSON b) => ToJSON (MyEither a b) where
+    toJSON = genericToJSON defaultOptions { sumEncoding = UntaggedValue }
+    toEncoding = genericToEncoding defaultOptions { sumEncoding = UntaggedValue }
+
+instance (FromJSON a, FromJSON b) => FromJSON (MyEither a b) where
+    parseJSON = genericParseJSON defaultOptions { sumEncoding = UntaggedValue }
```
RyanGlScott: Using `optsUntaggedValue` above would save some characters.

tests/Properties.hs
```diff
@@ -172,6 +172,16 @@ isObjectWithSingleField :: Value -> Bool
...
 isObjectWithSingleField _            = False
 
+-- | EitherTextInt: String, Number or Array
+isUntaggedValue :: Value -> Bool
+isUntaggedValue (String s)
+    | s == "nonenullary"   = True
+isUntaggedValue (Bool _)   = True
+isUntaggedValue (Number _) = True
+isUntaggedValue (Array a)  = length a == 2
+isUntaggedValue _          = False
```
RyanGlScott: Does this test work for any datatype, or only `EitherTextInt`?

phadej: only `EitherTextInt`

RyanGlScott: Ah, OK. I was confused since the implementation doesn't quite match the documentation (it says `String, Number or Array`, but it also checks for `Bool`).

Data/Aeson/Types/Internal.hs
```diff
@@ -549,6 +549,10 @@ data SumEncoding =
     -- by the encoded value of that field! If the constructor is not a
     -- record the encoded constructor contents will be stored under
     -- the 'contentsFieldName' field.
+  | UntaggedValue
+    -- ^ Constructor won't be encoded. The contents of the constructor will be
+    -- encoded as if the type had single constructor.  JSON encodings have to
```
bergmark: missing "be" at eol

Data/Aeson/Types/Internal.hs
```diff
@@ -549,6 +549,10 @@ data SumEncoding =
     -- by the encoded value of that field! If the constructor is not a
     -- record the encoded constructor contents will be stored under
     -- the 'contentsFieldName' field.
+  | UntaggedValue
+    -- ^ Constructor won't be encoded. The contents of the constructor will be
+    -- encoded as if the type had single constructor.  JSON encodings have to be
+    -- disjoint for decoding to work properly.
```
bergmark: I think this can be clarified a bit, how about "Constructor **names** won't be encoded. **Instead only** the contents of the constructor will be encoded, as if the type had a single constructor"?

bergmark:
Looking at usage with nullary constructors; We might want to warn about using them with UntaggedValue since the constructor name is not removed in that case. It will also lead to ambiguity if a nullary constructor is used with a one field constructor encoding to a string.

phadej:
```hs
    -- ^ Constructor names won't be encoded. Instead only the contents of the
    -- constructor will be encoded as if the type had single constructor. JSON
    -- encodings have to be disjoint for decoding to work properly.
    --
    -- When decoding, constructors are tried in the order of definition. If some
    -- encodings overlap, the first one defined will succeed.
    --
    -- /Note:/ Nullary constructors are encoded as the string (using
    -- 'constructorTagModifier'). Make sure that other constructor encodings
    -- are disjoint.
    --
    -- /Note:/ only the last error is kept when decoding, so in the case of
    -- mailformed JSON, only error for the last constructor will be reported.
```
?

bergmark:
How about just spelling it out: "Make sure that other constructor encodings are disjoint." -> "Having a nullary constructor alongside a single field constructor that encodes to a string leads to ambiguity"?

bergmark:
Looking great, thanks!

PR #446, opened 2016-07-08
Fix #445: correct .:? haddock

### LINKED ISSUE #445: Error in comment on `.:?`
The comment [currently reads](https://github.com/bos/aeson/blob/9fcb4a023a240d7baf5df11058d04cc13aa7b6a4/Data/Aeson/Types/FromJSON.hs#L663):
```
-- | Retrieve the value associated with the given key of an 'Object'.
-- The result is 'Nothing' if the key is not present, or 'empty' if
-- the value cannot be converted to the desired type.
```

If the key is present with the value `Null` and cannot be converted to the desired type this suggests that the result will be `empty`. But in practice the result is a successful parse:
```
> let f = (\a -> a .:? "type") :: HashMap Text Value -> Parser (Maybe Bool)
> parse f (HM.singleton "type" Null)
Success Nothing
```

Also if the key is present and the value can be converted to the desired type the comment suggests that it will. But this isn't the case if the value is `Null`:
```
> data Foo = Foo deriving Show
> instance FromJSON Foo where parseJSON Null = pure Foo
> let g = (\a -> a .:? "type") :: HashMap Text Value -> Parser (Maybe Foo)
> parse g (HM.singleton "type" Null) -- Expecting `Success Foo`
Success Nothing
```

phadej:


bergmark:
@seagreen What do you think about this change?

seagreen:
LGTM

bergmark:
Cool!

PR #448, opened 2016-07-11
Support QuickCheck-2.9

phadej:
Hopefully we had all CPP checks right.


phadej:
We shouldn't merge this before `quickcheck-instances` support `QuickCheck-2.9` as well.

PR #449, opened 2016-07-11
Support QuickCheck-2.9 for 0.11

phadej:


phadej:
ah my bad. The tests don't actually compile if I use `--alow-newer`

PR #453, opened 2016-07-15
Make aeson-benchmark-typed compare to old aeson

phadej:
screen shot 2016-07-15 at 20 10 15

Slight slowdown since 0.11.0.2, but my machine is quite noisy atm.

ping @winterland1989, you can rebase on that and try

```
STACK_YAML=stack-bench.yaml stack build
cd benchmark
STACK_YAML=../stack-bench.yaml stack exec aeson-benchmark-typed -- -o bench.htm
```

### DIFF
--- a/benchmarks/AesonParse.hs
+++ b/benchmarks/AesonParse.hs
@@ -1,8 +1,9 @@
 {-# LANGUAGE BangPatterns, OverloadedStrings #-}
+{-# LANGUAGE PackageImports #-}
 
 import Control.Exception
 import Control.Monad
-import Data.Aeson
+import "aeson-benchmarks" Data.Aeson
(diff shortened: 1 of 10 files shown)


winterland1989:
`aeson-benchmark-typed` looks like it's all about encoding speed, so i don't think my pr will change that figures.

phadej:
Ah true. We could (should?) add decoding there as well.

phadej:
screen shot 2016-07-18 at 01 22 39

@winterland1989 I added decode benchmarks too. At least on my machine the difference with `aeson-0.11.2.0` is quite small, so your changes should be visible well with this setup too.

winterland1989:
ok, I'll merge from upstream once this get merged, and update the benchmark result.

PR #455, opened 2016-07-16
Fix TH bug when deriving ToJSON1/2 or FromJSON1/2 for types that don't mention the parameter

RyanGlScott:
Before, this code:

```haskell
newtype Foo a = Foo (Foo Int)

$(deriveToJSON  defaultOptions ''Foo)
$(deriveToJSON1 defaultOptions ''Foo)
```

Would generate the following `ToJSON1` instance:

```haskell
    deriveToJSON1 defaultOptions ''Foo
  ======>
    instance ToJSON1 Foo where
      liftToJSON
        = \ _tj1_aFns _tjl1_aFnt value_aFnr
            -> case value_aFnr of {
                 Foo arg1_aFnu -> liftToJSON toJSON toJSONList arg1_aFnu }
      liftToEncoding
        = \ _te1_aFnw _tel1_aFnx value_aFnv
            -> case value_aFnv of {
                 Foo arg1_aFny
                   -> liftToEncoding toEncoding toEncodingList arg1_aFny }
```

Although this typechecks, actually contradicts the strategy that GHC adapts when deriving `* -> *`-kinded classes (e.g., `Functor`). Namely, when an argument type doesn't mention the last type parameter at all, GHC doesn't attempt to "tunnel into" the type expression like the above code is doing with the expression `liftToJSON toJSON toJSONList`.

With the changes in this PR, the generated code would instead be:

```haskell
    deriveToJSON1 defaultOptions ''Foo
  ======>
    instance ToJSON1 Foo where
      liftToJSON
        = \ _tj1_aFo0 _tjl1_aFo1 value_aFnZ
            -> case value_aFnZ of { Foo arg1_aFo2 -> toJSON arg1_aFo2 }
      liftToEncoding
        = \ _te1_aFo4 _tel1_aFo5 value_aFo3
            -> case value_aFo3 of { Foo arg1_aFo6 -> toEncoding arg1_aFo6 }
```

### DIFF
--- a/tests/UnitTests.hs
+++ b/tests/UnitTests.hs
@@ -15,10 +15,10 @@ import Instances ()
 
 import Control.Applicative (Const(..))
 import Control.Monad (forM)
-import Data.Aeson (FromJSONKeyFunction(..), FromJSONKey(..), decode, eitherDecode, encode, genericParseJSON, genericToJSON, genericToEncoding, object, FromJSON(..), withObject, (.=), (.:), (.:?), (.:!))
+import Data.Aeson (FromJSONKeyFunction(..), FromJSONKey(..), decode, eitherDecode, encode, genericParseJSON, genericToJSON, genericToEncoding, object, FromJSON(..), withObject, ToJSON1(..), (.=), (.:), (.:?), (.:!))
 import Data.Aeson.Text (encodeToTextBuilder)
 import Data.Aeson.Internal (JSONPathElement(..), formatError)
-import Data.Aeson.TH (deriveJSON)
+import Data.Aeson.TH (deriveJSON, deriveToJSON, deriveToJSON1)
 import Data.Aeson.Types (ToJSON(..), Value, camelTo, camelTo2, defaultOptions, omitNothingFields, Options(..), SumEncoding(..))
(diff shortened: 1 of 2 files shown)


phadej:
Could you add the test for this? If i understood correctly, `liftToJSON undefined undefined (Foo 1)` should work, but wasn't working previously?

RyanGlScott:
Good point - I've added a regression test for this property.

tests/UnitTests.hs
```diff
@@ -93,6 +93,9 @@ tests = testGroup "unit" [
   , testGroup "Issue #351" $ fmap (testCase "-") issue351
   , testGroup "Nullary constructors" $ fmap (testCase "-") nullaryConstructors
   , testGroup "FromJSONKey" $ fmap (testCase "-") fromJSONKeyAssertions
+  , testGroup "PR #455" [
```
phadej: it could be directly `testCase "PR #445" pr445`, no need to wrap in `testGroup`

phadej:
Except from a minor style issue, LGTM.

RyanGlScott:
Good point, updated.

bergmark:
Cool, thanks!

Issue #461, opened 2016-08-13
instances for Data.Semigroup.*


lspitzner:
As of base-4.9/ghc-8, `base` contains `Data.Semigroup` (and aeson already depends on semigroups anyway), so the instances belong here.

One question though: I looked at the instance for `Identity` and noticed it indirects to a manual `FromJSON1` instance. What is the benefit of this? Wouldn't it completely suffice to implement FromJSON directly using `fmap Identity` and `runIdentity`? (Most stuff in semigroups is similar newtype wrappers.)

bergmark:
Which Semigroup instances would you like to add? As far as I can tell all monoids have semigroup instances already.

bergmark:
Oh, sorry, I realize you mean types in the semigroups package, not semigroup instances.

bergmark:
@phadej can you answer the question about Identity?

I agree that we should add the missing instances. They can be placed next to the NonEmpty instances in `Data.Aeson.Types.ToJSON/FromJSON`. Another thing to note is that we only conditionally depend on the semigroups package so we need to make sure all these types are in both `base` and `semigroups`. I think so but I haven't checked.

 @lspitzner are you willing to work on this? It's a non-breaking addition so we can release a v1.0.1.0 whenever this is ready.

lspitzner:
I am mildly certain that the conditionally dependency is exactly as necessary - `Data.Semigroup` will always be available, from `base` for >=ghc8 and `semigroups` for <ghc8.

Sure, i can write a PR. Would include instances for `First`, `Last`, `Option` and `WrappedMonoid`. All newtype wrappers, so the instances are trivial. Will wait for an answer from @phadej.

phadej:
Every `FromJSON` is implemented using `FromJSON1` if there is  one, just to have consistent instances and DRY.

phadej:
IIRC `Min` and `Max` are in base as well

lspitzner:
ah of course, the FromToJson1 are exposed, too. Derp.

Right, will add Min/Max too.

Issue #365, opened 2016-02-09
Add regression tests for the mapping of types to their json format


bergmark:
We have a lot of roundtrip tests for the formats but they don't verify that the format doesn't accidentally change (as could be seen by the format for Either accidentally changing in v0.10).

Would someone be interested in adding a comprehensive set of tests for all instances? First priority is to check the format on successful encoding/decoding. An extension of this is to also verify that bad input isn't allowed, but that is more open-ended.

See https://github.com/bos/aeson/blob/master/tests/UnitTests.hs#L216

bergmark:
I'm re-opening this since I'm working on making it more comprehensive.

phadej:
I think this in a good shape, if you think otherwise let's re-open *with an explicit list of types*, otherwise this is not actionable.

PR #462, opened 2016-08-13
Add [From|To]JSON[1] instances for Data.Semigroup.* newtypes

lspitzner:


lspitzner:
Fixes #461

phadej:
I'd add at least example encodings to `UnitTests.hs`

bergmark:
Yes please re tests. I'm reworking a specification for the serialization formats as mentioned in #365 so if that gets completed these tests can be added there (vacation and spotty hack time at the moment).

lspitzner:
eh, i had some `FromJSON1`s without the corresponding `FromJSON`s.

Monoid.* had no tests, so i added those.

Just noticed the note about lossy nested Maybes. This extends to Option too then. Should that be pointed out somewhere?

Should i also write changelog entry/bump version?

bergmark:
> Just noticed the note about lossy nested Maybes. This extends to Option too then. Should that be pointed out somewhere?

That sounds like a good thing to do.

> Should i also write changelog entry/bump version?

Changelog entries are always welcome and they can state whether the changes are breaking, but let's wait with version bumps until i do the release (I'll add this to CONTRIBUTING.md).

Thanks!

lspitzner:
Added changelog entry.

Don't really know where to add the "lossy" documentation, as it currently seems to only be mentioned in the unittests and on the tracker. Perhaps at the module haddocks at `Data.Aeson`? (And maybe this should be handled as a separate issue, along the lines of "document how specific instances are currently implemented".)

bergmark:
Thanks!

Short descriptions can be placed directly on the instance http://stackoverflow.com/a/21871914/182603

lspitzner:
But [From|To]JSON[1] ~ 4 instances per type. I'll copy the same note to those instancecs; i guess that is fine if i keep the doc short.

bergmark:
I'm not saying we should document every instance :) just unexpected things like this. The test suite is a better place to specify everything in detail.

PR #464, opened 2016-09-08
Support QuickCheck < 2.9.2

phadej:


phadej:
tested locally. Hopefully travis picks up `QuickCheck-2.9.1`

phadej:
I made also
- https://github.com/nick8325/quickcheck/pull/115 for `Proxy`
- https://github.com/aslatter/qc-instances/pull/25 for `Tagged`

First may be soon added, so if there's no hurry with `QuickCheck-2.9` support otherwise, we could wait to have those instances CPP guarded as well.

PR #465, opened 2016-09-13
Make the comment on `.:!` more accurate.

seagreen:
`.:!` won't necessarily fail of the value its given is `Null`,
rather it will attempt to parse it like any other value.

Also: touch up the comment on `.:?` slightly.


seagreen:
This resolves the doc issue I added as a side note [here](https://github.com/bos/aeson/issues/445#issuecomment-231452697).

Since this is a (slight) doc change it's pretty subjective, let me know if there's anything I can improve.

bergmark:
Thanks!

PR #470, opened 2016-09-29
Add incoherent instances to LookupField in Data.Aeson.TH

### LINKED ISSUE #468: Overlapping instances for Data.Aeson.TH.LookupField with Maybe
I'm currently upgrading a package from `aeson-0.8.1.1` to `aeson-0.11.2.1` (and GHC 7.10.2 to 8.0.1), and I've run into an issue with TH that works in the former but not in the latter.

Consider the following code (assuming the appropriate language extensions):

```haskell
module A where
type family F :: * -> *
data T = T { f1 :: F (), f2 :: Maybe (F ()) }

module B where
import A
import Data.Aeson
import Data.Aeson.TH
instance FromJSON (F ()) => FromJSON T where
  parseJSON = $(mkParseJSON defaultOptions ''T)
```

I get the following error when compiling:

```
        • Overlapping instances for Data.Aeson.TH.LookupField (F ())
            arising from a use of ‘Data.Aeson.TH.lookupField’
          Matching instances:
            instance [overlappable] FromJSON a => Data.Aeson.TH.LookupField a
              -- Defined in ‘Data.Aeson.TH’
            instance FromJSON a => Data.Aeson.TH.LookupField (Maybe a)
              -- Defined in ‘Data.Aeson.TH’
          (The choice depends on the instantiation of ‘’
           To pick the first instance above, use IncoherentInstances
           when compiling the other instance declarations)
        • In the second argument of ‘(<$>)’, namely
            ‘(Data.Aeson.TH.lookupField
                "A.T"
                "T"
                recObj_aauM
                (text-1.2.2.1:Data.Text.pack "f1"))’
          In the first argument of ‘(<*>)’, namely
            ‘(S.T
              <$>
                (Data.Aeson.TH.lookupField
                   "A.T"
                   "T"
                   recObj_aauM
                   (text-1.2.2.1:Data.Text.pack "f1")))’
          In the expression:
            ((S.T
              <$>
                (Data.Aeson.TH.lookupField
                   "A.T"
                   "T"
                   recObj_aauM
                   (text-1.2.2.1:Data.Text.pack "f1")))
             <*>
               (Data.Aeson.TH.lookupField
                  "A.T"
                  "T"
                  recObj_aauM
                  (text-1.2.2.1:Data.Text.pack "f2")))
```

The error appears to be specific to having at least two field names, one of them a `Maybe` and one of them not. It doesn't happen for the following definitions of `T`:

```haskell
data T = T (F ()) (Maybe (F ()))
data T = T { f1 :: F () }
data T = T { f1 :: F (), f2 :: F () }
data T = T { f1 :: Maybe (F ()) }
data T = T { f1 :: Maybe (F ()), f2 :: Maybe (F ()) }
data T = T { f1 :: F (), f2 :: Either () (F ()) }
```

The error appears to be specific to both type and data families. (It still occurs if you change `F` to a `data family`.) It does not happen for `F` defined as follows:

```haskell
data F a
newtype F a = F ()
```

spl:
This confirms that `IncoherentInstances` does indeed fix #468.

### DIFF
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -13,6 +13,7 @@
 {-# LANGUAGE TemplateHaskell #-}
 #endif
 
+#include "incoherent-compat.h"
 #include "overlapping-compat.h"
 
 {-|
@@ -1222,7 +1223,7 @@ class LookupField a where
 instance OVERLAPPABLE_ LookupField a where
     lookupField = lookupFieldWith
 
-instance LookupField (Maybe a) where
+instance INCOHERENT_ LookupField (Maybe a) where
     lookupField pj _ _ = parseOptionalFieldWith pj
 
 lookupFieldWith :: (Value -> Parser a) -> String -> String
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -18,6 +18,7 @@
 {-# LANGUAGE PolyKinds #-}
 #endif
 
+#include "incoherent-compat.h"
 #include "overlapping-compat.h"
 
 -- TODO: Drop this when we remove support for Data.Attoparsec.Number
@@ -942,9 +943,8 @@ instance ( FromRecord arity a
       (:*:) <$> parseRecord opts fargs Nothing obj
             <*> parseRecord opts fargs Nothing obj
 
-instance ( Selector s
-         , GFromJSON arity a
-         ) => FromRecord arity (S1 s a) where
+instance OVERLAPPABLE_ (Selector s, GFromJSON arity a) =>
+  FromRecord arity (S1 s a) where
     parseRecord opts fargs lab =
       (<?> Key label) . gParseJSON opts fargs <=< (.: label)
         where
@@ -952,7 +952,7 @@ instance ( Selector s
           defLabel = pack . fieldLabelModifier opts $
                        selName (undefined :: t s a p)
 
-instance OVERLAPPING_ (Selector s, FromJSON a) =>
+instance INCOHERENT_ (Selector s, FromJSON a) =>
   FromRecord arity (S1 s (K1 i (Maybe a))) where
     parseRecord _ _ (Just lab) obj = (M1 . K1) <$> obj .:? lab
     parseRecord opts _ Nothing obj = (M1 . K1) <$> obj .:? pack label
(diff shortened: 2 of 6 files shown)


RyanGlScott:
Although I can't see a better solution than this, would it also be possible to apply a similar trick for the [generic implementation of `FromJSON`](https://github.com/bos/aeson/blob/8044d1a54d4dc619a9e740c4c651795807b0abc0/Data/Aeson/Types/FromJSON.hs#L945-L961)?

spl:
@RyanGlScott: Done (3273f02). I'm not sure if `IncoherentInstances` will affect the other instances in `Data.Aeson.Types.FromJSON`. Might be beneficial to extract `FromRecord` into its own module.

spl:
Hmm, fails on [GHC 7.8.4](https://travis-ci.org/bos/aeson/jobs/163738091).

RyanGlScott:
Interesting. It looks like `aeson-0.8.1.1` didn't use `IncoherentInstances` in [its file](https://hackage.haskell.org/package/aeson-0.8.1.1/src/Data/Aeson/Types/Generic.hs) that defined the GHC generics code, which is why it successfully built on GHC 7.8 (so `aeson-0.8.1.1` exhibited this problem only in GHC generics–based code, not in TH-based code!) I suppose we really have no choice but to isolate these incoherent instances to their own modules to avoid infecting other instances.

spl:
@RyanGlScott:

> I suppose we really have no choice but to isolate these incoherent instances to their own modules to avoid infecting other instances.

Ugh. That's problematic because `FromRecord` and `GFromJSON` are mutually recursive.

Possibly solutions:

* `.hs-boot` file
* Only use `{-# INCOHERENT #-}` pragma on instance and don't use `{-# LANGUAGE IncoherentInstances #-}`: supports only GHC >= 7.10

bergmark:
I'm haming a hard time picking a favorite...

RyanGlScott:
Actually, I have a third idea for eliminating the use of `IncoherentInstances` (and `OverlappingInstances` as well) in this scenario. See https://github.com/bos/aeson/issues/468#issuecomment-250777315

spl:
I took the easy way out and updated the PR to annotate with `INCOHERENT` only the instances that are needed to be incoherent (2nd option in https://github.com/bos/aeson/pull/470#issuecomment-250674709), which means `aeson` would support more types in GHC >= 7.10 than < 7.10.

If you do find a better way to do this, that would be great. Otherwise, I hope you'll accept this.

bergmark:
Thanks! I think I do prefer the solution you chose. I'll try to release this tomorrow.

bergmark:
Shipped!

spl:
Thanks!

Issue #476, opened 2016-10-19
"Invalid UTF-8 stream" error on valid characters


vshabanov:
Aeson can't decode some characters, for example [U+1F3FF](https://en.wikipedia.org/wiki/Fitzpatrick_scale)

```
> JSON.eitherDecode "\"\\ud83c\\udfff\"" :: Either String String
Left "Error in $: Failed reading: Cannot decode input: Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream"
```

I've hacked a test to check out what other valid characters are affected
https://gist.github.com/vshabanov/4653f07311fc61bc397cc53db98f2407
here are results:
```
10000;LINEAR B SYLLABLE B008 A;Lo;0;L;;;;;N;;;;;
10400;DESERET CAPITAL LETTER LONG I;Lu;0;L;;;;;N;;;;10428;
10800;CYPRIOT SYLLABLE A;Lo;0;R;;;;;N;;;;;
10C00;OLD TURKIC LETTER ORKHON A;Lo;0;R;;;;;N;;;;;
11000;BRAHMI SIGN CANDRABINDU;Mc;0;L;;;;;N;;;;;
11400;NEWA LETTER A;Lo;0;L;;;;;N;;;;;
11C00;BHAIKSUKI LETTER A;Lo;0;L;;;;;N;;;;;
12000;CUNEIFORM SIGN A;Lo;0;L;;;;;N;;;;;
12400;CUNEIFORM NUMERIC SIGN TWO ASH;Nl;0;L;;;;2;N;;;;;
13000;EGYPTIAN HIEROGLYPH A001;Lo;0;L;;;;;N;;;;;
133FF;EGYPTIAN HIEROGLYPH Z015E;Lo;0;L;;;;;N;;;;;
13400;EGYPTIAN HIEROGLYPH Z015F;Lo;0;L;;;;;N;;;;;
14400;ANATOLIAN HIEROGLYPH A001;Lo;0;L;;;;;N;;;;;
16800;BAMUM LETTER PHASE-A NGKUE MFON;Lo;0;L;;;;;N;;;;;
17000;<Tangut Ideograph, First>;Lo;0;L;;;;;N;;;;;
18800;TANGUT COMPONENT-001;Lo;0;L;;;;;N;;;;;
1B000;KATAKANA LETTER ARCHAIC E;Lo;0;L;;;;;N;;;;;
[...]

PR #477, opened 2016-10-20
Fix for "Invalid UTF-8 stream" error on valid characters (#476).

vshabanov:


phadej:
We definitely should have tests for this, could you add tests from the issue into `aeson`s test suite?

vshabanov:
Added a test. Optimized it a bit so it could check all characters in a second instead of half a minute and added basic escapement (like `\n`) test besides UTF-16 `\uXXXX` characters test.

bergmark:
Looks great, thanks!

PR #479, opened 2016-10-27
Specialize ToJSON for common types

niteria:
Here's a motivating example:

```
mkArray :: [Value] -> Array
mkArray = Vector.fromList

mkObject :: [(Text, Value)] -> Object
mkObject = HashMap.fromList

someArray :: Array
someArray =
  mkArray
     [toJSON
        (mkObject
           ["dummy" .=
              mkObject
                ["dummy" .=
                   "dummy",
                 "dummy" .= "dummy",
                 "dummy" .= 42,
                 "dummy" .= 42],
            "dummy" .=
              mkArray
                [toJSON "dummy", toJSON "dummy"],
            "dummy" .=
              mkObject
                ["dummy" .= 42,
                 "dummy" .= 42],
            "dummy" .= mkArray [toJSON "dummy"]])]
```

If `aeson` doesn't pre-specialize, every time this
example is compiled you pay the cost of specialization.

In this case this blows up the number of terms from
181 to 2464 making compile time significantly slower than
necessary.

### DIFF
--- a/Data/Aeson/Types/ToJSON.hs
+++ b/Data/Aeson/Types/ToJSON.hs
@@ -615,6 +615,11 @@ instance ToJSON1 [] where
     {-# INLINE liftToEncoding #-}
 
 instance (ToJSON a) => ToJSON [a] where
+    {-# SPECIALIZE instance ToJSON String #-}
+    {-# SPECIALIZE instance ToJSON [String] #-}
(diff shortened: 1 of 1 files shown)


bergmark:
Hi @niteria,
I'm sorry it took me so long to reply, I must have accidentally archived the notification...

Compile time has taken a hit with the 1.0 release so this is great, thanks! 🎉

PR #482, opened 2016-11-20
Add ToJSON1 Proxy, and ToJSON2 Tagged

phadej:


phadej:
I actually have: I have `data CreateFoo f` where `f` might be `Identity` or `Proxy`, dependending on the stage, to indicate that some fields have been populated by the other party

PR #483, opened 2016-11-21
Add MonadError Parser

phadej:


phadej:
Alternatively, if `mtl` dependency is unwanted, specialised `catchError` would be enough to my usecase (amending error messages).

phadej:
Ah, there is `modifyFailure`.

phadej:
@bergmark opinions on this?

bergmark:
This seems a bit overkill to me since we have modifyFailure. I suppose you could also add this instance locally with a newtype over Parser, or would that be terribly clunky?

phadej:
I don't insist on this. Can we leave this open if more people will stumble on this and ask for?

bergmark:
I think we can close this, it's the top result when searching for "mtl" or "MonadError" so hopefully people will find it!

PR #485, opened 2016-12-11
Add UUID instances

phadej:


bergmark:
Thanks! Would using to/fromASCIIBytes be an improvement?

phadej:
Ah true, have to write `toEncoding`, there toASCIIBytes would make sense (and tests, obviously)

phadej:
Tests added, travis is green

Issue #484, opened 2016-12-03
Runtime serialization error with MySQL `0000-00-00` `Day`s


parsonsmatt:
[The `day` function in `Data.Aeson.Encoding.Builder`](https://github.com/bos/aeson/blob/d8dafaea947f50c890a1b7327a058b4d08359d91/Data/Aeson/Encoding/Builder.hs#L188-L201) is partial, failing where the year is 0 or lesser. I'm running into this with MySQL `date` values being stored as `0000-00-00`.

There's a fairly trivial fix for this by including support for year `0000`, which is `BCE 1` according to [the wikipedia article](https://en.wikipedia.org/wiki/ISO_8601#Years). Ideally, the function could be extended to support all BCE years. At least, I'd hope that we could throw a more specific exception type than just `error`.

I'm happy to put together a patch for this issue.

andrewthad:
I'm in favor of this change. Encoding functions should not be partial. Just to clarify, the wikipedia article indicates that way to encode years would be:

- 1 CE -> 0001
- -1 BCE -> 0000
- -2 BCE -> -0001

It would be nice (and pretty easy I think) to be a little more generous when decoding by allowing a leading plus sign so that they year could also be written as `+2016` or `2016`. @bergmark Your thoughts?

bergmark:
I'm confused re. what the `Day` type is meant to represent. [The time docs](https://hackage.haskell.org/package/time-1.7/docs/Data-Time-Calendar.html#t:Day) say "The Modified Julian Day is a standard count of days, with zero being the day 1858-11-17." Is that a documentation error?

bergmark:
n/m I see that this is a comment on the implementation of Day.

bergmark:
Fixed by #486

PR #486, opened 2016-12-11
Add BCE day support

parsonsmatt:
This PR adds support for the encoding of BCE `Day`s, fixing #484.

### DIFF
--- a/Data/Aeson/Parser/Time.hs
+++ b/Data/Aeson/Parser/Time.hs
@@ -25,6 +25,7 @@ module Data.Aeson.Parser.Time
 import Prelude ()
 import Prelude.Compat
 
+import Control.Applicative ((<|>))
 import Control.Monad (void, when)
 import Data.Aeson.Internal.Time (toPico)
 import Data.Attoparsec.Text as A
@@ -49,11 +50,7 @@ run p t = case A.parseOnly (p <* endOfInput) t of
 -- | Parse a date of the form @[+,-]YYYY-MM-DD@.
 day :: Parser Day
 day = do
-  absOrNeg <- choice
-    [ char '-' *> pure negate
-    , char '+' *> pure id
-    , pure id
-    ]
+  absOrNeg <- negate <$ char '-' <|> id <$ char '+' <|> pure id
   y <- decimal <* char '-'
   m <- twoDigits <* char '-'
   d <- twoDigits
--- a/Data/Aeson/Parser/Time.hs
+++ b/Data/Aeson/Parser/Time.hs
@@ -46,13 +46,18 @@ run p t = case A.parseOnly (p <* endOfInput) t of
             Left err -> fail $ "could not parse date: " ++ err
             Right r  -> return r
 
--- | Parse a date of the form @YYYY-MM-DD@.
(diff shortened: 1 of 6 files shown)


phadej:
@parsonsmatt verify (cherry-pick into your branch) that tests #487 in pass.

parsonsmatt:
Thanks for the test @phadej, I've fixed a bug and now everything is passing :smile:

Data/Aeson/Parser/Time.hs
```diff
@@ -46,13 +46,18 @@ run p t = case A.parseOnly (p <* endOfInput) t of
             Left err -> fail $ "could not parse date: " ++ err
             Right r  -> return r
 
--- | Parse a date of the form @YYYY-MM-DD@.
+-- | Parse a date of the form @[+,-]YYYY-MM-DD@.
 day :: Parser Day
 day = do
+  absOrNeg <- choice
+    [ char '-' *> pure negate
```
phadej: I prefer `negate <$ char '-'` way, but probably there isn't any difference

phadej: (in fact `negate <$ char '-' <|> negate <$ char '+" <|> pure id` may inline better, as `choice` is recursive, but again, probably there are no difference.

parsonsmatt: That's an easy enough change to make. It even fits on one line, too!

phadej:
Thanks, this looks great!

parsonsmatt:
Not at all. I'll put that together soon and push it up.

parsonsmatt:
Ok, unit tests that describe the current encodings are up. Would you like to see any other cases?

phadej:
@parsonsmatt sorry for not spotting that earlier, but there's a small framework for specifying examples around https://github.com/bos/aeson/blob/66a5b73182ae5670ecd088301b4e18d93540bfba/tests/UnitTests.hs#L337

phadej:
Also would be nice to have `> 0 && <1000` example, though it's redundant: zero year looks too much as a special case

phadej:
And, now as I think about, what's `"-0000-01-01"`, I guess the same as without the minus; but good to give it an example (`jsonDecodingExamples` IIRC)

parsonsmatt:
Thanks for the pointer! Writing it up as an `Example` fails, since the test for `jsonDecodingExample` asserts that the `encode (fromGregorian 0 1 1) == "\"-0000-01-01\""`. I'll add it as a non-Example test case.

phadej:
@parsonsmatt:

```diff
   , MaybeExample "Word8 -1"   "-1"   (Nothing :: Maybe Word8)
   , MaybeExample "Word8 300"  "300"  (Nothing :: Maybe Word8)
+  -- Negative zero year, encoding never produces such:
+  , MaybeExample "negative zero year" "\"-0000-01-01\"" (Just $ fromGregorian 0 1 1)
   ]
 
 jsonExamples :: [Example]
````

phadej:
You found it yourself. Great!

parsonsmatt:
Let me know if there's anything else I can do for this PR :smile:

phadej:
LGTM

bergmark:
Great, thanks!

Issue #491, opened 2016-12-13
More general pair


andrewthad:
The type signature of pair is:

    pair :: Text -> Encoding -> Series

However, it would make sense to provide another pair function:

    pair' :: Encoding' Text -> Encoding -> Series

This function would be help me improve the performance (very mildly) of something I'm working on.

phadej:
This make sense, please make a PR!

andrewthad:
PR is up https://github.com/bos/aeson/pull/492

PR #492, opened 2016-12-15
Add another function for building a key-value pair Encoding

andrewthad:
Resolves https://github.com/bos/aeson/issues/491


phadej:
It has to be exported from `Data.Aeson.Encoding` module as well, it's not that. I'd also re-implement `pair` in terms of `pair'`

andrewthad:
Thanks for the feedback. I've addressed both issues that @phadej brought up.

phadej:
LGTM

PR #493, opened 2016-12-16
Remove unncessary constraints from To/FromJSONkey Identity

phadej:


bergmark:
PVP says "Breaking change. If [...] the types of any entities [...] were changed." so strictly speaking it's a major bump. But is this really breaking?

Ping @hvr

PR #494, opened 2016-12-16
Add non-infix parseField, also explicit liftParseField

phadej:


Data/Aeson/Types/FromJSON.hs
```diff
@@ -59,11 +59,18 @@ module Data.Aeson.Types.FromJSON
     -- * Functions
     , fromJSON
     , ifromJSON
+    , typeMismatch
+    , parseField
+    , parseFieldMaybe
+    , parseFieldMaybe'
+    , liftParseField
```
phadej: I'm not sure `lift`but it kind of make sense here.

phadej:
I'm not sure whether these functions could be exported from `Data.Aeson`

Data/Aeson/Types/FromJSON.hs
bergmark: opaleye uses `explicit` for this

bergmark:
A pro of using operator aliases is that the operators get a pronouncable name. Cons are that there are twice as many operator names to memorize, and that TMTOWTDI. I'm on the fence about this.

bergmark:
... another pro is that partial application becomes more legible

phadej:
This is motivated by the need for `liftParseField`, as the haddock points out the example usage.

phadej:
Renamed to `explicitParseField`

bergmark:
Thanks!

PR #499, opened 2016-12-22
Fix incorrect GitHub repository language detection

yegortimoshenko:
Fixes incorrect repository language detection by ignoring vendored `tests/JSONTestSuite` folder.
See: https://github.com/github/linguist#overrides


bergmark:
Thanks!

PR #509, opened 2017-02-04
Re-export some TH type names in Data.Aeson.TH

xtendo-org:
When writing functions that call `deriveJSON` with different options, type names such as `Name` or `Q` are required for the function's type signature. As they are not exported by this package, the user must separately import `Language.Haskell.TH` and add `template-haskell` to the project `build-depends` section. Creating wrappers of `deriveJSON` is inconvenient because of this.

This Pull Request attempts to relieve the pain in a minimally invasive way.

As far as I know, this is a backwards-compatible change.

### DIFF
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -115,6 +115,10 @@ module Data.Aeson.TH
     , mkParseJSON
     , mkLiftParseJSON
     , mkLiftParseJSON2
+
+    , Dec
+    , Name
+    , Q
     ) where
 
 import Prelude ()


bergmark:
I'm unsure about this; aeson almost never re-exports identifiers from other packages. e.g. if you want to give a type signature for `decode` you need to import `ByteString` from `bytestring`.

 is there a reason to treat template-haskell separately?

bergmark:
For the moment let's stick with the status quo. There is no clear best way to handle this but I'm leaning towards not re-exporting for all my projects. Let me know if you want more details on my reasoning.

Thank you for your contribution even though it didn't end up being merged!

Issue #510, opened 2017-02-04
Best practices for FromJSON


andrewthad:
In the haddocks for `Data.Aeson`, the following example is given:

    {-# LANGUAGE OverloadedStrings #-}
    
    instance FromJSON Person where
        parseJSON (Object v) = Person <$>
                               v .: "name" <*>
                               v .: "age"
        -- A non-Object value is of the wrong type, so fail.
        parseJSON _          = empty

However, this is not the best way to write this instance. In my own code, I would write this:

    {-# LANGUAGE OverloadedStrings #-}
    
    instance FromJSON Person where
        parseJSON = withObject "Person" $ \v -> Person
            <$> v .: "name"
            <*> v .: "age"

This gives better error messages when the JSON is malformed and is more concise. I propose changing the docs to use this as the example.

phadej:
I agree. Using `with*` helpers is much nicer.

bergmark:
I agree!

phadej:
@andrewthad could you make a PR?

andrewthad:
PR is up: https://github.com/bos/aeson/pull/511

PR #511, opened 2017-02-07
improve example FromJSON instance

andrewthad:
Resolves https://github.com/bos/aeson/issues/510

### DIFF
--- a/Data/Aeson.hs
+++ b/Data/Aeson.hs
@@ -274,9 +274,9 @@ eitherDecodeStrict' =
 -- > {-# LANGUAGE OverloadedStrings #-}
 -- >
 -- > instance FromJSON Person where
--- >   parseJSON = withObject "Person" $ \v -> Person
--- >     <$> v .: "name"
--- >     <*> v .: "age"
+-- >     parseJSON = withObject "Person" $ \v -> Person
+-- >         <$> v .: "name"
+-- >         <*> v .: "age"
 --
 -- We can now parse the JSON data like so:
 --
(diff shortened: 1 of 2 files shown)


Data/Aeson.hs
```diff
@@ -274,11 +274,9 @@ eitherDecodeStrict' =
 -- > {-# LANGUAGE OverloadedStrings #-}
 -- >
 -- > instance FromJSON Person where
--- >     parseJSON (Object v) = Person <$>
--- >                            v .: "name" <*>
--- >                            v .: "age"
--- >     -- A non-Object value is of the wrong type, so fail.
--- >     parseJSON _          = empty
+-- >   parseJSON = withObject "Person" $ \v -> Person
```
phadej: Most code examples use 4 space indents. It's worth to keep the consistent style.

phadej:
Style issue, otherwise LGTM, thanks!

RyanGlScott:
@andrewthad: There is at least one more comment in `aeson` that needs a similar fix. A quick search reveals [this one](https://github.com/bos/aeson/blob/d8dafaea947f50c890a1b7327a058b4d08359d91/Data/Aeson/Types/FromJSON.hs#L288).

andrewthad:
@RyanGlScott Good catch. I'll get that fixed.

andrewthad:
I ended up needing to leave the existing example since the docs use the existing example as a demonstration of `typeMismatch`. I've provided in the `withObject` approach as an alternative with the mention that it is "to be preferred". Let me know what you think.

bergmark:
LGTM

Issue #450, opened 2016-07-13
Faster parser using FFI?


winterland1989:
Recently we found a bottleneck in our data pipeline which read json from upstream and routing them into downstream, during benchmark aeson show the expected throughtput(50~70MB/s), but a nodejs guy in our team showed that v8 can do even better(about 3~5 time faster on parsing).

I tried many optimization including use multiple ghc thread and `Chan`, but still can't defeat a single threaded v8 on parsing, on the other hand, encoding seems on pair with v8. it makes me wondering if we can do better by use some fast c code, for example like what [json-stream](https://github.com/ondrap/json-stream) does. Any idea welcomed! 

N.B. In benchmark aeson still beat go version, this's not a complaint thread, just want to start a discussion on further optimization possibilities : )

phadej:
How do you parse the data with `aeson`. Do you use `Generic`, `TemplateHaskell` or manual deriving? Parsing into ADT (with all checks), costs some CPU. Or just parsing into `Value`? Do you data has a lot of ints, or some other specific characteristics (I'm quite sure, that going thru `Scientific` slows parsing a bit, but we care about correctness).

I doubt that with current overall architecture/feature set we can get constant factors much lower.

It would be nice to know how `json-stream` compares to `aeson`. Do you want to add benchmark (I guess `Typed.hs` is most interesting, using `twitter` data)

winterland1989:
I first design the ADT version which is replaced by nodejs now，during benchmark we directly parse into `Value` because I think it's not fair to Haskell version if we count addtional ADT construction, but benchmark showed that the ADT construction time is negligible.

I will try to make a bench on json-stream, we did use Twitter sample json actually, the one around 60KB.

phadej:
If you can share the benchmarks, it would be great!

winterland1989:
The benchmark code is [here](https://github.com/winterland1989/aeson-benchmark), result:

```
benchmarking twitter10.json/aeson parser
time                 137.3 μs   (135.7 μs .. 138.8 μs)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 138.0 μs   (136.8 μs .. 139.2 μs)
std dev              4.272 μs   (3.694 μs .. 5.045 μs)
variance introduced by outliers: 28% (moderately inflated)

benchmarking twitter10.json/json-stream parser
time                 86.81 μs   (86.16 μs .. 87.51 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
[...]

PR #513, opened 2017-02-13
Pure Haskell UTF parser for `unescapeText`.

jprider63:
#450

### DIFF
--- a/Data/Aeson/Parser/Unescape.hs
+++ b/Data/Aeson/Parser/Unescape.hs
@@ -7,6 +7,258 @@ module Data.Aeson.Parser.Unescape (
 ) where
 
 import           Control.Exception          (evaluate, throw, try)
+import           Control.Monad              (when)
+import           Data.ByteString            as B
+import           Data.Bits                  (shiftL, shiftR, (.&.), (.|.))
+import           Data.Text                  (Text)
+import qualified Data.Text.Array            as A
+import           Data.Text.Encoding.Error   (UnicodeException (..))
+import           Data.Text.Internal.Private (runText)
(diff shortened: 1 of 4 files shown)


bergmark:
Good stuff, thanks!

Like i mentioned in #450 it'd be nice to have this in separate module.
That way both unit tests and benchmarks can be run against both implementations without needing a rebuild. 

I can look into that this week.

jprider63:
Sorry, I missed that request initially. I've split them into separate modules. I think you'll still need to rebuild to test and benchmark though.

bergmark:
I rebased this on top of the `unescape` branch along with other changes but a test fails on travis. I wasn't able to reproduce it locally: https://travis-ci.org/bos/aeson/jobs/203164303#L859

jprider63:
This looks like FromJSONKey? This test fails for me on the main branch (d470ac6), so I assumed it was fine.

bergmark:
Very strange, I haven't seen that error before

jprider63:
It is odd. I've uploaded the [failed logs](http://lpaste.net/3696006614487138304) from master in case you want to see it. How should we proceed?

bergmark:
Thanks for following up!

Are you able to debug this locally? It would be very helpful since I haven't been able to reproduce it.

Even if we don't figure out how to fix this it would be good to know the severity; Any particular things to avoid / any workarounds? Will it cause bugs?

Since no one has mentioned being affected by this when using the library maybe we could release this with a disclaimer, but I think we should make sure that we get the test suite passing on travis again.

bergmark:
Resolved thanks to @Lysxia!

Rebased @ f16a7f5. I'm releasing now.

PR #519, opened 2017-03-17
Disable rewrite rules for tests with flag(fast)

Lysxia:
Lets test "MyText'" pass


bergmark:
Awesome @Lysxia, thanks! This was blocking #513 so now we can finally make a release.

Cheers

PR #520, opened 2017-03-18
Refactor TH derivation and fix #517

### LINKED ISSUE #517: deriveJSON breaks serialized isomorphism for unary types
I haven't dug into the dumped splices yet, but here are the results I'm seeing:

```haskell
data Foo = Foo

$(deriveJSON defaultOptions ''Foo)
```

```haskell
decode (encode Foo) == Just Foo
```
> False

I think the `FromJSON` produced expects an array, while `ToJSON` creates an Object. Sorry about this :)

Lysxia:
#517


bergmark:
Great stuff as usual, thank you!

Issue #473, opened 2016-10-10
Generic encoding of data type with single nullary constructor


linearray:
`data Foo = Foo deriving (Show,Generic,ToJSON,FromJSON)`

> > encode Foo
> "[]"

Is this behaviour intentional? When reading through the documentation for defaultOptions I saw nothing that would indicate this sort of "shortcut".

hvr:
FWIW, it's consistent with the encoding for the canonical single-nullary-constructor case `encode ()` :-)

phadej:
hmm, https://github.com/bos/aeson/blob/master/tests/UnitTests/NullaryConstructors.hs show something rather different, or are missing some case?

linearray:
In the test cases C1 is one of three type constructors for the data type Nullary.

To trigger the reported behaviour you need a data type with a single nullary type constructor.

bergmark:
Is this solved by the `allNullaryToStringTag` flag of the deriving options?

linearray:
No. Regardless of its setting, with the default being `True`, the output is the same.

bergmark:
Changing defaults is scary since this isn't caught by the type checker. Too bad the default is what we'd like to change here :(

What are our other options?

1. Adding a new option only targeting single constructors: seems ad hoc
2. Adding a new option is a "fixed" allNullaryToStringTag: confusing in docs and inconsistent if used with allNullaryToStringTag
3. Adding a new option and removing the old one: Breaks code but in a good way, could be backwards compatible
4. Changing the type of the current option to cover all three cases: Also breaks in a good way + backwards compat

I'm leaning towards number 3.

linearray:
`allNullaryToStringTag` is `True` by default, so I do not see Option 3 being much different from changing the default right away (assuming the new option will also be `True` by default).

People with old, generically produced JSON data stored away will only be able to read it back in manually.
People, who rely on the current output (JSON schema, etc.) will be surprised.

Lysxia:
What are the three cases?

Single-constructor types are unwrapped as a list, a new option to disable this unwrapping would actually make for a less ad-hoc behavior, and allow `allNullaryToStringTag` and `sumEncoding` options to apply.

bergmark:
Does #522 resolve this in a good way for you @linearray?

linearray:
I did not have any time to test this, but from a quick glance I think this is the best solution one can hope for. Thanks for implementing!

phaazon:
[...]

Issue #206, opened 2014-05-19
Irregular Aeson.TH results with ObjectWithSingleField 


pjonsson:
This might be a documentation issue. The documentation for ObjectWithSingleField says:

A constructor will be encoded to an object with a single field named after the constructor tag (modified by the constructorTagModifier) which maps to the encoded contents of the constructor.

Here's my test program:

``` Haskell
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE CPP #-}
module Main where

import Language.Haskell.Exts.Annotated
import Data.Aeson
import Data.Aeson.TH

$(deriveToJSON defaultOptions { sumEncoding = ObjectWithSingleField } ''ExportSpecList)
$(deriveToJSON defaultOptions { sumEncoding = ObjectWithSingleField } ''SrcSpanInfo)
$(deriveToJSON defaultOptions { sumEncoding = ObjectWithSingleField } ''SrcSpan)
```

Running `ghc -ddump-splices --make Main.hs` gives (1/3):

``` Haskell
Main.hs:1:1: Splicing declarations
    deriveToJSON
      (defaultOptions {sumEncoding = ObjectWithSingleField}) ''SrcSpan
  ======>
    Main.hs:25:3-79
    instance ToJSON SrcSpan where
      toJSON
        = \ value_a81J
            -> case value_a81J of {
                 SrcSpan arg1_a81K arg2_a81L arg3_a81M arg4_a81N arg5_a81O
                   -> object
                        [((Data.Text.pack "srcSpanFilename") .= arg1_a81K),
                         ((Data.Text.pack "srcSpanStartLine") .= arg2_a81L),
                         ((Data.Text.pack "srcSpanStartColumn") .= arg3_a81M),
                         ((Data.Text.pack "srcSpanEndLine") .= arg4_a81N),
                         ((Data.Text.pack "srcSpanEndColumn") .= arg5_a81O)] }
```

An object, and a set of values that are of primitive types so no need to `toJSON` them. Ok, SrcSpanInfo next:

``` Haskell
Main.hs:1:1: Splicing declarations
    deriveToJSON
      (defaultOptions {sumEncoding = ObjectWithSingleField})
      ''SrcSpanInfo
  ======>
    Main.hs:24:3-83
    instance ToJSON SrcSpanInfo where
      toJSON
        = \ value_a80U
            -> case value_a80U of {
[...]

PR #522, opened 2017-03-18
Option to encode single-constructor types as tagged sums

Lysxia:
As suggested in #206  and possibly #473

### DIFF
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -341,7 +341,7 @@ consToValue jc opts cons = do
     matches tjs = case cons of
       -- A single constructor is directly encoded. The constructor itself may be
       -- forgotten.
-      [con] -> [argsToValue jc tjs opts False con]
+      [con] | not (tagSingleConstructors opts) -> [argsToValue jc tjs opts False con]
       _ | allNullaryToStringTag opts && all isNullary cons ->
               [ match (conP conName []) (normalB $ conStr opts conName) []
               | con <- cons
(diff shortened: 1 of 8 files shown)


bergmark:
Github showed the already merged PR's in the history so i force pushed a rebase.

bergmark:
Thanks again, This looks good as well!

Unavoidably this is a breaking change due to the `Options` constructor being exposed. I'll file an issue for un-exposing in the next major along with this to avoid this in the future.

I'll cut a release with the other PR's tomorrow and then merge this.

bergmark:
Looks like i also confused myself and travis by doing "git push origin Lysxia/feature-474" so don't mind the test failure!

bergmark:
I noticed that the new field needs to be added to the `Show Options` instance. I added a regression test so rebasing this PR should break. This should be rebased on top of #526, but let's save some effort and wait for that to get merged!

bergmark:
#526 was merged so this can be rebased.

bergmark:
Same here, I didn't get notified of the recent changes.

Thanks a lot for this!

Lysxia:
Okay! You're welcome!

Actually this patch may lead to some serious regressions because I've introduced an actual loop in the generic implementation. We can probably avoid it by tracking the new option at the type level, as a parameter of the `GToJSON` type class, or by splitting that class in some way.

Lysxia:
This is now just two simple overlapping instances instead of the previous recursive mess.

Ready to merge.

bergmark:
Great stuff!

Issue #424, opened 2016-06-08
Are INLINE pragmas worth the compilation time?


phadej:
### Without (after removal of pragmas)

INLINE pragmas in `FromJSON.hs` and `ToJSON.hs`

Second line ais nother clean build with aeson, criterion, statistics and aeson-benchmarks:
It seems that INLINE pragmas slowdown the downstream package compilation as well?

```
stack build aeson  69.23s user 2.65s system 99% cpu 1:12.35 total
stack build       313.90s user 12.86s system 100% cpu 5:26.45 total
```

```
benchmarking generic/direct/twitter100
time                 405.4 μs   (402.9 μs .. 407.9 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 405.0 μs   (402.6 μs .. 407.8 μs)
std dev              8.039 μs   (6.568 μs .. 10.82 μs)
variance introduced by outliers: 12% (moderately inflated)

benchmarking manual/direct/twitter100
time                 387.0 μs   (383.2 μs .. 391.5 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 385.2 μs   (383.0 μs .. 387.6 μs)
std dev              8.185 μs   (7.162 μs .. 9.457 μs)
variance introduced by outliers: 14% (moderately inflated)

benchmarking th/direct/twitter100
time                 406.7 μs   (402.7 μs .. 409.9 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 404.4 μs   (401.7 μs .. 407.7 μs)
std dev              9.895 μs   (8.084 μs .. 12.91 μs)
variance introduced by outliers: 16% (moderately inflated)
```
### With:

```
stack build aeson  179.39s user 4.51s system 97% cpu 3:09.30 total
stack build        567.26s user 16.09s system 100% cpu 9:41.51 total
```

```
benchmarking generic/direct/twitter100
time                 399.7 μs   (395.2 μs .. 405.4 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 400.3 μs   (397.5 μs .. 403.5 μs)
std dev              9.911 μs   (7.580 μs .. 12.47 μs)
variance introduced by outliers: 17% (moderately inflated)

benchmarking manual/direct/twitter100
time                 385.4 μs   (383.1 μs .. 387.9 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
[...]

PR #335, opened 2016-01-19
Speed up default FromJSON/ToJSON instances

RyanGlScott:
Fixes #296 and #309 (I believe).

The `DefaultInstances`-based mechanism for deriving `FromJSON` and `ToJSON` instances via GHC generics currently consumes way more memory than it should. I believe this is the result of two things:

1. [GHC Trac #9630](https://ghc.haskell.org/trac/ghc/ticket/9630). It was discovered that having generic classes that have more than one method [such as](https://github.com/bos/aeson/blob/4667ef1029a373cf4510f7deca147c357c6d8947/Data/Aeson/Types/Class.hs#L38):

   ```haskell
   class GToJSON f where
     gToJSON :: Options -> f a -> Value
     gToEncoding :: Options -> f a -> Encoding
   ```

   hurt the optimizer badly. To compensate for this bug, I split `GToEncoding` off from `GToJSON` (as well as the many internal typeclasses that `GToJSON` uses).
2. `Data.Aeson.Types.Generic` puts `INLINE` pragmas on all generic class instance methods. This results in an explosion of inlined code, to the point that [`pandoc-types` takes ~7 GB of memory](https://github.com/fpco/stackage/issues/845#issuecomment-171342178) to compile. In my experience, inlining generic methods has turned out to be a bad idea, so I've removed the `INLINE` pragmas.

To test out the improvments, I did a very fast-and-loose profiling of the time and memory it takes to compile `pandoc-types` (a package known to be [affected badly](https://github.com/bos/aeson/issues/296#issuecomment-169826563) by the `aeson-0.10` compilation regressions). I did these tests on a 64-bit Linux laptop with 4 GB of RAM. Here are the results:

* `pandoc-types` (vanilla `aeson-0.10`):

```
$ /usr/bin/time -v cabal install pandoc-types
Resolving dependencies...
Downloading pandoc-types-1.16.0.1...
Configuring pandoc-types-1.16.0.1...
Building pandoc-types-1.16.0.1...
Installed pandoc-types-1.16.0.1
        Command being timed: "cabal install pandoc-types"
        User time (seconds): 251.96
        System time (seconds): 5.90
        Percent of CPU this job got: 41%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 10:20.85
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 3048536
[...]

PR #524, opened 2017-03-19
Refactor Generic ToJSON

Lysxia:
This merges the code for `GToEncoding` and `GToJSON` which look awfully similar.

I decided to keep exposing `GToEncoding` and `GToJSON` with the original kind as constraint synonyms for backwards compatibility, as they seem to be used in some places for type annotations. I don't think their methods are in use externally however.

This will conflict with #522. I guess I'll resolve it depending on whichever gets merged first.

### DIFF
--- a/Data/Aeson.hs
+++ b/Data/Aeson.hs
@@ -76,8 +76,8 @@ module Data.Aeson
     -- ** Generic JSON classes and options
     , GFromJSON(..)
     , FromArgs(..)
-    , GToJSON(..)
-    , GToEncoding(..)
+    , GToJSON
+    , GToEncoding
     , ToArgs(..)
     , Zero
     , One
--- a/Data/Aeson/Encoding.hs
+++ b/Data/Aeson/Encoding.hs
@@ -14,6 +14,7 @@ module Data.Aeson.Encoding
     , Series
     , pairs
     , pair
+    , pairStr
     , pair'
     -- * Predicates
     , nullEncoding
--- a/Data/Aeson/Encoding/Internal.hs
+++ b/Data/Aeson/Encoding/Internal.hs
@@ -13,6 +13,7 @@ module Data.Aeson.Encoding.Internal
     , Series (..)
     , pairs
     , pair
+    , pairStr
     , pair'
     -- * Predicates
     , nullEncoding
@@ -124,6 +125,11 @@ data Series = Empty
 
 pair :: Text -> Encoding -> Series
 pair name val = pair' (text name) val
+{-# INLINE pair #-}
+
+pairStr :: String -> Encoding -> Series
+pairStr name val = pair' (string name) val
+{-# INLINE pairStr #-}
 
 pair' :: Encoding' Text -> Encoding -> Series
 pair' name val = Value $ retagEncoding $ retagEncoding name >< colon >< val
--- a/Data/Aeson/Types.hs
+++ b/Data/Aeson/Types.hs
@@ -63,8 +63,8 @@ module Data.Aeson.Types
     -- ** Generic JSON classes
     , GFromJSON(..)
     , FromArgs(..)
-    , GToJSON(..)
-    , GToEncoding(..)
+    , GToJSON
+    , GToEncoding
     , ToArgs(..)
     , Zero
     , One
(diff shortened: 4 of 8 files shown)


phadej:
There is a performance reason they are separate, ping @RyanGlScott

RyanGlScott:
The only thing critical for compilation speed is that `GToJSON` has exactly one class method (see #335), which this PR preserves.

That being said, @Lysxia, can you run this code on a generics-related benchmark (like [this one](https://github.com/bos/aeson/blob/d8dafaea947f50c890a1b7327a058b4d08359d91/benchmarks/AesonCompareAutoInstances.hs)) to see if the compilation time/runtime are about the same? If so, :+1: from me.

phadej:
If I might ask, than would be nice to run benchmark on old GHCs as well, e.g 7.8.4 and 7.10.3 too, would be great that performance doesn't degrade on old compilers dramatically

Lysxia:
Sure!

bergmark:
Looks like there are some build errors on older GHC's

Lysxia:
Here are some benchmarks. One file per GHC version. `master` is on the left, this PR is on the right. There are two parts, the first with the flag `fast` turned on (the default), and the second with it off. I compare the time to compile the library, then the time to compile the benchmark executable, first without optimization, then with. I then run the optimized one.

https://gist.github.com/Lysxia/bb3e6dd68f6121b904363dc48e11da0d

`aeson` takes about the same time to build in any case. To compile the final executable with an optimized `aeson`, this PR seems to slow down a bit, except with `8.0.2`. 

    7.8.4  user	0m12.077s	user	0m12.967s
    7.10.3 user	0m15.960s	user	0m18.560s
    8.0.1  user	0m19.010s	user	0m22.313s
    8.0.2  user	0m19.000s	user	0m18.223s

At runtime there are cases where this PR is slower if aeson is not optimized, but faster if it is optimized. For instance:

     7.8.4
     D/toJSON/generic
     aeson+fast   28.52 μs   32.91 μs
     aeson-fast   9.335 μs   5.926 μs

I found one case where the reverse happens:

    8.0.2
    BigRecord/encode/generic
    aeson+fast   63.22 μs   50.10 μs
    aeson-fast   3.886 μs   10.15 μs

Not sure what's going on yet but I can look into it later.

phadej:
@Lysxia what are the base point times, i.e. before the patch?

phadej:
Great job anyways, tell if you cannot find a reason for BigRecord digression

bergmark:
> the first with the flag fast turned on (the default)

What do you mean by default? when compiling aeson normally fast is off and -02 is used.

> To compile the final executable with an optimized aeson, this PR seems to slow down a bit, except with 8.0.2.

So it actually got faster on 8.0.2!? I'm inclined to say that we can limit performance requirements to the latest ghc, do others agree?

phadej:
@begrmark, I don't care about compilation performance, but run-time performance shouldn't be drastically different on at least three previous major GHC releases (i.e. 7.8.4-8.0.2 atm)

bergmark:
> I don't care about compilation performance, but run-time performance shouldn't be drastically different on at least three previous major GHC releases

Oops, I misread the first portion as stating everything got faster on 8.0.2 :)

Lysxia:
> What do you mean by default? when compiling aeson normally fast is off and -02 is used.

Oops, you are right. I got it confused with the fact that it is turned on by the `stack-lts*.yaml` files.

As for performance there are unfortunately some regressions to investigate, but there is hope that this update will be at least as fast as it is currently.

Lysxia:
`GHC.Generics.from` and `to` are not inlined for large types ([GHC issue](https://ghc.haskell.org/trac/ghc/ticket/11068)), and this causes the inliner to behave somewhat erratically.

Adding `INLINE` annotations in aeson helps, but I think a more effective solution would be for users to somehow derive `Generic` instances with `INLINE` `to` and `from`.

bergmark:
Related to inlining: #424

Lysxia:
It turns out inlining `from` isn't sufficient. But by selectively inlining a few more functions I think I've managed to fix all regressions without affecting compile times too much. At least the one case I mentioned above is now a speed up.

Lysxia:
Latest comparison benchmarks, left is before patch, right is after.

https://gist.github.com/Lysxia/7ea97543e1821ae7f73758e82d35b6a9

Issue #518, opened 2017-03-07
Documentation of minimal complete definition for `class ToJson` when using non-defaultOptions


lspitzner:
(I specified `toEncoding` with special options and discovered that `toJSON` still used defaults.. and now I re-read the docs and I find them unclear:)

The last paragraph reads:

> If DefaultSignatures doesn't give exactly the results you want, you can customize the generic encoding with only a tiny amount of effort, using genericToJSON and genericToEncoding with your preferred Options:
> ~~~~
> instance ToJSON Coord where
>     toJSON     = genericToJSON defaultOptions
>     toEncoding = genericToEncoding defaultOptions
> ~~~~

Should this be s/DefaultSignatures/defaultOptions/ at the start and s/defaultOptions/myFancySpecialOptions/ in the code?

Maybe the "minimal complete definition" can/should be specified clearly for both cases (default and non-default), like it is common in other classes?

Lysxia:
I agree the documentation can be clarified.

> Should this be s/DefaultSignatures/defaultOptions/ at the start and s/defaultOptions/myFancySpecialOptions/ in the code?

I would answer "not quite" to the first, "yes" to the second.

My interpretation is that `DefaultSignatures` refers to the default implementations of the type class methods, though now that you mention it, this isn't worded very well.

By looking at the source of `ToJSON`, you can see that: the default `toJSON` is `genericToJSON defaultOptions`; the default `toEncoding` calls `toJSON`.

https://github.com/bos/aeson/blob/master/Data/Aeson/Types/ToJSON.hs#L274-L296

A minimal complete definition is thus given by `toJSON` (e.g., with `genericToJSON` and custom options). Only redefining `toEncoding` is not sufficient, as you remarked.

bergmark:
Sounds good!

bergmark:
Do you think we are ok to close this now @lspitzner? Please re-open if not.

PR #525, opened 2017-03-19
Reword comments about ToJSON and FromJSON

Lysxia:
Hoping this addresses #518.

@lspitzner What do you think?

### DIFF
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -281,12 +281,12 @@ genericLiftParseJSON opts pj pjl = fmap to1 . gParseJSON opts (From1Args pj pjl)
 --
 -- The basic ways to signal a failed conversion are as follows:
 --
--- * 'empty' and 'mzero' work, but are terse and uninformative
+-- * 'empty' and 'mzero' work, but are terse and uninformative;
 --
--- * 'fail' yields a custom error message
+-- * 'fail' yields a custom error message;
 --
 -- * 'typeMismatch' produces an informative message for cases when the
--- value encountered is not of the expected type
+-- value encountered is not of the expected type.
 --
 -- An example type and instance using 'typeMismatch':
 --
@@ -296,27 +296,27 @@ genericLiftParseJSON opts pj pjl = fmap to1 . gParseJSON opts (From1Args pj pjl)
 --
 -- data Coord = Coord { x :: Double, y :: Double }
 --
--- instance FromJSON Coord where
---     parseJSON ('Object' v) = Coord
---         '<$>' v '.:' \"x\" 
+-- instance 'FromJSON' Coord where
+--     'parseJSON' ('Object' v) = Coord
+--         '<$>' v '.:' \"x\"
 --         '<*>' v '.:' \"y\"
 --
 --     \-- We do not expect a non-'Object' value here.
 --     \-- We could use 'mzero' to fail, but 'typeMismatch'
 --     \-- gives a much more informative error message.
---     parseJSON invalid    = 'typeMismatch' \"Coord\" invalid
+--     'parseJSON' invalid    = 'typeMismatch' \"Coord\" invalid
 -- @
 --
 -- For this common case of only being concerned with a single
--- type of JSON value, the functions @withObject@, @withNumber@, etc.
+-- type of JSON value, the functions 'withObject', 'withNumber', etc.
 -- are provided. Their use is to be preferred when possible, since
--- they are more terse. Using @withObject@, we can rewrite the above instance 
+-- they are more terse. Using 'withObject', we can rewrite the above instance
 -- (assuming the same language extension and data type) as:
 --
 -- @
--- instance FromJSON Coord where
---     parseJSON = withObject \"Coord\" $ \v -> Coord
---         '<$>' v '.:' \"x\" 
+-- instance 'FromJSON' Coord where
+--     'parseJSON' = 'withObject' \"Coord\" $ \v -> Coord
+--         '<$>' v '.:' \"x\"
 --         '<*>' v '.:' \"y\"
 -- @
 --
@@ -325,7 +325,7 @@ genericLiftParseJSON opts pj pjl = fmap to1 . gParseJSON opts (From1Args pj pjl)
 --
 -- * "Data.Aeson.TH" provides Template Haskell functions which will derive an
 -- instance at compile time. The generated instance is optimized for your type
--- so will probably be more efficient than the following two options:
+-- so it will probably be more efficient than the following option.
 --
 -- * The compiler can provide a default generic implementation for
 -- 'parseJSON'.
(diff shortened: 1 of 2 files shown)


Data/Aeson/Types/FromJSON.hs
```diff
@@ -325,7 +325,7 @@ genericLiftParseJSON opts pj pjl = fmap to1 . gParseJSON opts (From1Args pj pjl)
 --
 -- * "Data.Aeson.TH" provides Template Haskell functions which will derive an
 -- instance at compile time. The generated instance is optimized for your type
--- so will probably be more efficient than the following two options:
+-- so will probably be more efficient than the following option.
```
bergmark: so it* will

Data/Aeson/Types/FromJSON.hs
```diff
@@ -343,18 +343,17 @@ genericLiftParseJSON opts pj pjl = fmap to1 . gParseJSON opts (From1Args pj pjl)
...
--- If @DefaultSignatures@ doesn't give exactly the results you want,
+-- If the default implementation doesn't give exactly the results you want,
 -- you can customize the generic decoding with only a tiny amount of
 -- effort, using 'genericParseJSON' with your preferred 'Options':
 --
 -- @
--- instance FromJSON Coord where
---     parseJSON = 'genericParseJSON' 'defaultOptions'
+-- instance 'FromJSON' Coord where
+--     'parseJSON' = 'genericParseJSON' customOptions
```
bergmark: Maybe have an example like `defaultOptions { sumEncoding = ObjectWithSingleField }` instead? Otherwise we don't explain how to construct `Options`. It also discourages usage of the Options constructor.

Data/Aeson/Types/FromJSON.hs
```diff
@@ -516,7 +515,7 @@ typeMismatch expected actual =
 --
 -- * "Data.Aeson.TH" provides Template Haskell functions which will derive an
 -- instance at compile time. The generated instance is optimized for your type
--- so will probably be more efficient than the following two options:
+-- so will probably be more efficient than the following option.
```
bergmark: so it* will, again

Data/Aeson/Types/FromJSON.hs
```diff
@@ -534,16 +533,16 @@ typeMismatch expected actual =
...
--- If @DefaultSignatures@ doesn't give exactly the results you want,
+-- If the default implementation doesn't give exactly the results you want,
 -- you can customize the generic decoding with only a tiny amount of
 -- effort, using 'genericLiftParseJSON' with your preferred 'Options':
 --
 -- @
--- instance FromJSON a => FromJSON1 (Pair a) where
---     liftParseJSON = 'genericLiftParseJSON' 'defaultOptions'
+-- instance 'FromJSON' a => 'FromJSON1' (Pair a) where
+--     'liftParseJSON' = 'genericLiftParseJSON' customOptions
```
bergmark: Same as above.

Data/Aeson/Types/ToJSON.hs
```diff
@@ -213,18 +213,18 @@ genericLiftToEncoding opts te tel = gToEncoding opts (To1Args te tel) . from1
...
+--   'toEncoding' (Coord x y) = 'pairs' (\"x\" '.=' x '<>' \"y\" '.=' y)
 -- @
 --
 -- Instead of manually writing your 'ToJSON' instance, there are two options
 -- to do it automatically:
 --
 -- * "Data.Aeson.TH" provides Template Haskell functions which will derive an
 -- instance at compile time. The generated instance is optimized for your type
--- so will probably be more efficient than the following two options:
+-- so will probably be more efficient than the following option.
```
bergmark: it*

Data/Aeson/Types/ToJSON.hs
```diff
@@ -257,15 +257,39 @@ genericLiftToEncoding opts te tel = gToEncoding opts (To1Args te tel) . from1
...
+--
+-- The following examples ensure that 'toEncoding' represents the same
+-- encoding as 'toJSON'.
+--
+-- - A typical implementation defines both 'toJSON' and 'toEncoding'
+--   with the same options.
+--
+-- @
+-- instance 'ToJSON' Coord where
+--     'toJSON'     = 'genericToJSON' customOptions
```
bergmark: defaultOptions suggestion

Data/Aeson/Types/ToJSON.hs
```diff
@@ -491,7 +515,7 @@ contramapToJSONKeyFunction h x = case x of
 --
 -- * "Data.Aeson.TH" provides Template Haskell functions which will derive an
 -- instance at compile time. The generated instance is optimized for your type
--- so will probably be more efficient than the following two options:
+-- so will probably be more efficient than the following option.
```
bergmark: it*

Data/Aeson/Types/FromJSON.hs
```diff
@@ -343,18 +343,21 @@ genericLiftParseJSON opts pj pjl = fmap to1 . gParseJSON opts (From1Args pj pjl)
...
 -- data Coord = Coord { x :: Double, y :: Double } deriving 'Generic'
 --
--- instance FromJSON Coord
+-- instance 'FromJSON' Coord
 -- @
 --
--- If @DefaultSignatures@ doesn't give exactly the results you want,
+-- If the default implementation doesn't give exactly the results you want,
 -- you can customize the generic decoding with only a tiny amount of
 -- effort, using 'genericParseJSON' with your preferred 'Options':
```
lspitzner: How about clarifying what the default implementation is?

> The default implementation will be equivalent to @parseJSON = 'genericParseJSON' 'defaultOptions'@; If you need different options, you can customize the generic decoding by defining:

Data/Aeson/Types/ToJSON.hs
```diff
@@ -213,18 +213,18 @@ genericLiftToEncoding opts te tel = gToEncoding opts (To1Args te tel) . from1
 --
```
lspitzner: (perhaps right under "A type that can be converted to JSON"; github does not let met add a comment there..)

> Instances in general /must/ specify 'toJSON' and /should/ (but do not need to) specify 'toEncoding'.

Data/Aeson/Types/ToJSON.hs
```diff
@@ -257,15 +257,43 @@ genericLiftToEncoding opts te tel = gToEncoding opts (To1Args te tel) . from1
...
+--   the generic options are not satisfactory.
+--
+-- @
+-- instance 'ToJSON' Coord where
+--      'toJSON' (Coord x y) = 'Array' (toJSON [x, y])
+-- @
+--
+-- - The default implementation of 'toJSON' uses 'defaultOptions'. If you leave
+--   'toJSON' implicit while providing a more efficient explicit implementation
+--   of 'toEncoding', then it *must* be equivalent to using 'defaultOptions'.
```
lspitzner: This adds a second example for a manually written instance when we already have an example of this above. I'd prefer to swap/reword things a bit. I'll make a suggestions as a commit, one sec.

lspitzner:
Now I wish to request pulling [this commit](https://github.com/lspitzner/aeson/commit/d59147294cac5ab4407aa72409459fa4fe394a8e) to this pull request. Or something in this direction.. :)

I have reworded the explanation behind `toEncoding`; please review if what is written is somewhat true.

lspitzner:
@bergmark btw when pushing i noticed that c44ee47542f99a6b24631071fa575042131bb4cb contains relatively large binaries. Is there a missing .gitignore entry?

bergmark:
I added your commit, hope you don't mind @Lysxia!

@lspitzner this was when Bryan did a git subtree import, so these things shouldn't change unless we re-do that.

bergmark:
Once again i pushed the qualified branch name to origin... hope it doesn't cause more issues.

bergmark:
Looks like this is ready to merge! More improvements are always welcome of course.

@Lysxia i don't get notifications when changes are pushed so please add a comment next time so I know!

PR #526, opened 2017-03-19
Refactor and future proof Show Options instance

bergmark:
I added a regression test for this in 88c244e.

Unfortunately pattern matching with record syntax does **not*** warn about left-overs which means adding fields to `Options` would silently make the show instance outdated. I changed it to a normal constructor match to prevent this. The downside is that the field names get lost but I think it's worthwhile.

I also don't like `RecordWildCards` since it causes shadowing. Shadowing is a warning under `-Wall` but not when combined with `RecordWildCards`!


phadej:
LGTM

bergmark:
Thanks!

PR #532, opened 2017-03-28
Fix Generic encoding of records with single Maybe field

Lysxia:
With

    data Single = Single { f :: Maybe Int } deriving Generic

    instance ToJSON Single where
      toJSON = genericToJSON defaultOptions{omitNothingFields=True,unwrapUnaryRecords=True}

We have:

    encode (Nothing :: Maybe Int) = "null"
    encode (Just 1 :: Maybe Int) = "1"

Yet:

    encode (Single Nothing) = "{}"
    encode (Single (Just 1)) = "1"

That is, if the field is `Nothing`, then the record is not being unwrapped and `omitNothingFields` applies instead, producing an empty object.

### DIFF
--- a/tests/UnitTests.hs
+++ b/tests/UnitTests.hs
@@ -18,11 +18,11 @@ import Prelude ()
 import Prelude.Compat
 
 import Control.Monad (forM, forM_)
-import Data.Aeson ((.=), (.:), (.:?), (.:!), FromJSON(..), FromJSONKeyFunction(..), FromJSONKey(..), ToJSON1(..), decode, eitherDecode, encode, genericToEncoding, genericToJSON, object, withObject)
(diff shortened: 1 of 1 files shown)


bergmark:
Looks great, thanks!

To be safe I think I'll include this in 1.2.0.0. Do you have a preferred order for merging your PR's?

Lysxia:
Actually I fixed this in the refactoring at the same time (which is how I came across the issue of this PR), so I would merge the refactoring (#524) before this one which would then only provide the test.

#524, #532, and finally #522 (after one last change to do I mentioned over there).

bergmark:
This has some merge conflicts now

Lysxia:
Rebased. This now just adds the regression test case.

bergmark:
Great!

Issue #535, opened 2017-03-31
Default to using haskell code instead of FFI (was: Security over Performance)


eskimor:
The C code in aeson worries me a bit. Aeson as JSON parser can be expected to sit right in the front of webservers, decoding user data! An ideal position for exploiting buffer overflow  and similar bugs.

I believe the default should be secure not fast, so maybe set cffi to false by default? 

This reduces the risk of a security relevant bug in a critical code section for all Haskell based servers and  performance can still be boosted for those who need it - at their own risk.

eskimor:
Especially this highly optimized C code with all the gotos, looks like it would be pretty easy for a security relevant bug to sneak in. I have to admit, I am concerned.

bergmark:
@eskimor Since 1.1.1.0 there is a flag to disable this code, from the changelog:

```
Added a pure implementation of the C FFI code. If you wish to use the pure haskell version set the cffi flag to False. This should make aeson compile when C isn't available, such as for GHCJS. 
```

Does this address your concerns?

eskimor:
Yes and no. It is good that there is a flag to disable it - so I can at least disable it for my server. What I am saying is, that it might be a good idea to switch the default from enabled to disabled, so aeson is secure by default and faster when needed.

bergmark:
I suggested this earlier but I guess it wasn't important to people at the time, so I'm fine with flipping the flag.

mbj:
+1 for the most secure default. Performance problems show up in a noticable way and users will search for solutions. Security problems only are noticed *after the fact* (if at all).

eskimor:
@mbj great argument! PR coming.

eskimor:
I just checked, the Haskell version isn't necessarily secure either, it uses unsafeWrite from Array. At least the code is more readable, currently checking where position is coming from and whether it can overflow by crafted user data.

eskimor:
I am currently checking the Haskell version for out of bounds problems, it is unsafe but looks ok. pos gets advanced at most by 1 during a fold over the input and the output is set to the same length. So it should not overflow. Except that the above is not true in one case:

https://github.com/bos/aeson/blob/master/pure/Data/Aeson/Parser/UnescapePure.hs#L164

here pos is incremented twice, once in the argument and once in the body of writeAndReturn. Is this fine - and if so  why? (I don't know nothing about unicode encodings - sorry)

eskimor:
I moved my last question to it's own issue.

PR #536, opened 2017-03-31
Remove INLINE pragma on timeOfDay64

RyanGlScott:
`timeOfDay64` is an enormous function, and marking it as `INLINE` causes a simple `aeson` benchmark to exhaust its simplifier ticks.

Thanks to @mpickering for spotting this in https://ghc.haskell.org/trac/ghc/ticket/12967#comment:4.

Fixes #502.

PR #537, opened 2017-04-03
Switch cffi flag default.

eskimor:
Also added on additional check in the pure parser, which should catch
buffer overflows hopefully soon enough to prevent any harm.

I also grepped aeson for other unchecked array accesses and found a few, I will review them too, just in case.

### DIFF
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -74,8 +74,8 @@ flag bytestring-builder
   manual: False
 
 flag cffi
-  description: Controls whether to include c-ffi bits or pure haskell
-  default: True
+  description: Controls whether to include c-ffi bits or pure haskell. Default to False for security.
+  default: False
   manual: True
 
 library
--- a/pure/Data/Aeson/Parser/UnescapePure.hs
+++ b/pure/Data/Aeson/Parser/UnescapePure.hs
@@ -255,8 +255,6 @@ unescapeText' bs = runText $ \done -> do
 
       {-# INLINE f #-}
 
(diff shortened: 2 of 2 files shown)


eskimor:
this fixes #535

eskimor:
Other unchecked array accesses in code are pretty trivial and look good to me.

eskimor:
ok - I have absolutely no idea why this fails for those old ghc versions.

bergmark:
Ouch, out of simplifier ticks. @RyanGlScott's PR #536 was supposed to help with this, could it have made it worse on older GHCs?

RyanGlScott:
I don't have time to look at this in much detail, but that `Data.Aeson.Parser.UnescapePure` module looks highly suspicious in that it marks some positively enormous functions as `INLINE`. I'd recommend removing some `INLINE` pragmas to see if that helps.

eskimor:
Thanks! Looks good now!

bergmark:
Cool, thanks!

bergmark:
I'm planning to release 1.2.0.0 in the beginning of next week. Is it ok if we skip a minor version release and ship this along with 1.2.0.0?

eskimor:
Definitely! This also gives time to resolve #538 first!

jprider63:
I'm also interested in the security of this. Mind if I add myself [here](https://github.com/bos/aeson/pull/537/files#diff-d463353550eaa09194d0c19e29ef48adR3)?

PR #541, opened 2017-04-11
Document ZonedTime FromJSON instance.

phaazon:


Data/Aeson/Types/FromJSON.hs
```diff
@@ -1671,6 +1671,17 @@ instance FromJSONKey LocalTime where
...
+-- @YYYY-MM-DD HH:MM Z@
+-- @YYYY-MM-DD HH:MM:SS Z@
+-- @YYYY-MM-DD HH:MM:SS.SSS Z@
+--
+-- The first space may instead be a @T@, and the second space is
+-- optional.  The @Z@ represents UTC.  The @Z@ may be replaced with a
+-- time zone offset of the form @+0000@ or @-08:00@, where the first
+-- two digits are hours, the @:@ is optional and the second two digits
+-- (also optional) are minutes.
 instance FromJSON ZonedTime where
```
phadej: should we rather document `UTCTime`, and refer to it documenting the differences?

phaazon: We could document both?

phadej: It's not worth repeating the docs, the format IIRC is the same, `ZonedTime` only preserves the offset info.

phaazon: Ah! Worth mentioning, yeah.

phaazon: Looking at what already exists, we document `ZonedTime` and state that `UTCTime` is the same but converts the offset into a UTC time.

phadej: Well, I'd say that `UTCTime` is more imporant one (i.e. more people will look it first), but let @bergmark / maintainers decide.

bergmark: I personally use UTCTime a lot more often, but this is good stuff regardless! If someone wants to move this around feel free to send a new PR!
Thanks!

PR #548, opened 2017-05-08
Weeder linting

bergmark:
The libs are now weed free, things to do:

* Set up travis job
* Fix remaining weeds in test suite

### DIFF
--- a/tests/DataFamilies/Types.hs
+++ b/tests/DataFamilies/Types.hs
@@ -18,7 +18,7 @@ import Types (ApproxEq(..))
 
 data family Nullary a
 data instance Nullary Int  = C1 | C2 | C3 deriving (Eq, Show)
-data instance Nullary Char = C4           deriving (Eq, Show)
+-- data instance Nullary Char = C4           deriving (Eq, Show)
 
 data family SomeType a b c
 data instance SomeType c () a = Nullary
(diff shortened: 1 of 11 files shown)


tests/Types.hs
```diff
@@ -125,11 +125,11 @@ failure func msg v = counterexample
 newtype BCEDay = BCEDay Day
   deriving (Eq, Show)
 
-zeroDay :: Day
-zeroDay = fromGregorian 0 0 0
-
 instance Arbitrary BCEDay where
     arbitrary = fmap (BCEDay . ModifiedJulianDay . (+ toModifiedJulianDay zeroDay)) arbitrary
+        where
+            zeroDay :: Day
```
phadej: weird indentation, is it intentional?

tests/DataFamilies/Types.hs
```diff
@@ -18,7 +18,7 @@ import Types (ApproxEq(..))
 
 data family Nullary a
 data instance Nullary Int  = C1 | C2 | C3 deriving (Eq, Show)
-data instance Nullary Char = C4           deriving (Eq, Show)
+-- data instance Nullary Char = C4           deriving (Eq, Show)
```
phadej: I wonder if this was used to trigger / test some bug/regression. Hopefully not.

tests/Types.hs
bergmark: Intentional :-) how would you do it?

phadej:
```haskell
instance
    arbitrary
      where
        zeroDay
```

I.e. use half-indentation for `where`

bergmark:
Probably better to start from scratch if we want to include this.

PR #550, opened 2017-06-06
Add parserThrowError and parserCatchError

phadej:
I ran into need for `parserCatchError` and cannot workaround it anymore :/ (`<|>` doesn't give me an error message, but is virtually the same)


bergmark:
I pushed a hlint fix

phadej:
The travis breakage seems transient.

bergmark:
Thanks!

Issue #529, opened 2017-03-21
Refactor to localize/minimize CPP


bergmark:
Ideally I'd have one module defining all the CPP, it makes it easier to re-use and removes a lot of noise. This can be done incrementally by moving any of it to a new `Data.Aeson.Compat` module. 

We probably can't extract everything, but here's the current list of potential suspects:

```
Data/Aeson/Encoding/Builder.hs:#if MIN_VERSION_bytestring(0,10,4)
Data/Aeson/Encoding/Builder.hs:#if !(MIN_VERSION_bytestring(0,10,4))
Data/Aeson/Internal/Time.hs:#if MIN_VERSION_base(4,7,0)
Data/Aeson/Parser/Internal.hs:#if MIN_VERSION_ghc_prim(0,3,1)
Data/Aeson/Parser/Internal.hs:#if MIN_VERSION_ghc_prim(0,3,1)
Data/Aeson/Parser/Internal.hs:#if MIN_VERSION_ghc_prim(0,3,1)
Data/Aeson/Parser/Unescape.hs:#ifdef CFFI
Data/Aeson/TH.hs:#if __GLASGOW_HASKELL__ >= 800
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,8,0) && !MIN_VERSION_template_haskell(2,10,0)
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,8,0)
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,7,0) && !(MIN_VERSION_template_haskell(2,8,0))
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,8,0) && !(MIN_VERSION_template_haskell(2,10,0))
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,11,0)
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,11,0)
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,11,0)
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,11,0)
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,11,0)
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,11,0)
Data/Aeson/TH.hs:#if MIN_VERSION_template_haskell(2,11,0)
[...]

PR #553, opened 2017-06-11
Use the th-abstraction library to factor out some TH boilerplate

RyanGlScott:
A lot of the messy TH internals have been factored out into the `th-abstraction` library. I advocate for using it here in `aeson`, since it saves almost 300 lines of code, most of which is ugly CPP.


bergmark:
Good stuff, thank you! This should sort out most of what I had in mind with #529. It makes sense to have this in an external package.

PR #555, opened 2017-06-16
Allow QuickCheck 2.10.*

bergmark:


phadej:
In the longer run, it would be great to bump lower bounds and remove conditional/CPP'd instances

bergmark:
Depending on what versions of aeson and QuickCheck end up in LTS 9 I may want to keep QC 2.9 support.

Issue #558, opened 2017-06-29
Uninformative invalid JSON error


ghost:
> Error in $: Failed reading: satisfy

I had an issue today where I saw this error. Having never seen it before it took me some time to track down what my problem was.

It was a simple case of my JSON not being valid after me making an innocent minor change after already checking for validity. Is it possible for Aeson to at least state that the attempted parse failed because the supplied JSON is invalid.

If someone could point to the place where this is handled I could have a go at doing it myself?

Orignally discussed here: https://github.com/bos/aeson/issues/113#issuecomment-311925548

phadej:
@JoeShanahanT https://github.com/bos/aeson/blob/2bc6f22f164d172d0d77d878422b67433a57a9d8/Data/Aeson/Parser/Internal.hs#L260

The `satisfy` is in the `msg` which we get from `attoparsec`. Maybe adding `<?>` in `json` and `value` parsers above in the file could improve some of the errors too (i.e what is not satisfied).

bergmark:
If you don't end up sending a PR then please provide a reproduction.

phadej:
```haskell
Prelude Data.Aeson> eitherDecode "{ 1 " :: Either String ()
Left "Error in $: Failed reading: satisfy"
```

JoeShanahanT:
I won't be able to submit something until Monday at the earliest, if that's fine then leave it to me otherwise please feel free to fix this.

phadej:
@JoeShanahanT as I said previously, I think it shouldn't be hard to fix this. Which makes this a perfect *first-time contributor* issue. Please hack on this when you have time. No need to hurry.

JoeMShanahan:
Moving on from my work centric account, I'm here for now! I'll take a look at this over the weekend as I didn't get around to do it previously!

JoeMShanahan:
@phadej 

I've littered `<?>` in many places, leading to no change in the error message.

Looking at the following, I don't see how this would come into play anyway?

https://github.com/bos/attoparsec/blob/36b2adc4ca869550c4c9feb7427e267126eb78c4/Data/Attoparsec/Text/Internal.hs#L109
https://github.com/bos/attoparsec/blob/36b2adc4ca869550c4c9feb7427e267126eb78c4/Data/Attoparsec/Internal/Types.hs#L154

bergmark:
Can we modify the message using something like `modifyFailure`?

JoeMShanahan:
[...]

PR #559, opened 2017-07-09
Better invalid json message (issue #558)

JoeMShanahan:
See #558.

First pass. Since these shows _all_ errors for parses named with `<?>` it leads to some other jankiness:

    > eitherDecode "{ 3" :: Either String Value
    Left "Error in $: Failed reading: satisfy, valid json, 34"

The "34" comes from https://hackage.haskell.org/package/attoparsec-0.13.1.0/docs/src/Data-Attoparsec-ByteString-Internal.html#word8

Suggestions welcome. I feel like this is _slightly_ better this could potentially be even more confusing.

### DIFF
--- a/Data/Aeson/Parser/Internal.hs
+++ b/Data/Aeson/Parser/Internal.hs
@@ -39,7 +39,8 @@ import Prelude.Compat
 import Control.Applicative ((<|>))
 import Control.Monad (void, when)
 import Data.Aeson.Types.Internal (IResult(..), JSONPath, Result(..), Value(..))
-import Data.Attoparsec.ByteString.Char8 (Parser, char, decimal, endOfInput, isDigit_w8, signed, string)
+import Data.Attoparsec.ByteString.Char8 (Parser, (<?>), char, decimal, endOfInput, isDigit_w8, signed, string)
+import Data.List (intercalate)
 import Data.Scientific (Scientific)
 import Data.Text (Text)
 import Data.Vector as Vector (Vector, empty, fromListN, reverse)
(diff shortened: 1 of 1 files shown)


Lysxia:
I think it should read in this order `"valid json, 34, satisfy"`, or the other way around, the reason being that the error comes from `value'` calling `word8` calling `satisfy`.

The `34` and `satisfy` parts of the message suggest some improvements to be made in attoparsec rather than in aeson: `word8` could be more descriptive, saying instead something like `"expected character 34 (")"`, and `satisfy` (or a new version of it) could require an error string (which `word8` would provide, thus removing the `(<?>)` wrapper) instead of providing an uninformative `"satisfy"` error.

Data/Aeson/Parser/Internal.hs
```diff
@@ -165,43 +166,47 @@ arrayValues val = do
...
-    DOUBLE_QUOTE  -> A.anyWord8 *> (String <$> jstring_)
-    OPEN_CURLY    -> A.anyWord8 *> object_
-    OPEN_SQUARE   -> A.anyWord8 *> array_
-    C_f           -> string "false" *> pure (Bool False)
-    C_t           -> string "true" *> pure (Bool True)
-    C_n           -> string "null" *> pure Null
-    _              | w >= 48 && w <= 57 || w == 45
-                  -> Number <$> scientific
-      | otherwise -> fail "not a valid json value"
+value = parser <?> "valid json"
```
Lysxia: `"valid json"` is indeed misleading, since the error is that the JSON is in fact invalid, but that's easy to change of course. Another issue is that annotating this recursive parser is going to result in large stacks of duplicated `"valid json"` strings when parsing nested structures like `"[[[[[[[[[[[[[[[[[[[[[[["`. Instead, you can insert this error string just in `eitherDecodeWith`, or perhaps add more descriptive information here, such as the current position, that may become quite heavy. It can be helpful as well, for instance by indicating where the potential mismatched brackets are, but there may also be a more lightweight solution.

Data/Aeson/Parser/Internal.hs
```diff
@@ -257,7 +262,8 @@ eitherDecodeWith p to s =
       L.Done _ v     -> case to v of
                           ISuccess a      -> Right a
                           IError path msg -> Left (path, msg)
-      L.Fail _ _ msg -> Left ([], msg)
+      L.Fail _ namedFailures msg -> Left ([], failMsg)
```
Lysxia: Maybe also add the first component of `Fail` (position?) in the error message.

JoeMShanahan: 👍

bergmark:
Don't mind the benchmark failure, it's unrelated. I tried to fix it yesterday but travis was confusing me.

It would be nice to have some tests that show what the error messages will look like now for simple cases and some more complex ones. Like @Lysxia mentioned it might have some unwanted consequences.

PR #560, opened 2017-07-11
RFC: Token stream

phadej:
While working on the sax-like stream based JSON parser, I noticed that parsing `Value` with it is faster than with `attoparsec`. It started as a separate library, but as it makes isolated (and quite internal) part of `aeson` a bit faster, I made this PR. If you think this doesn't below into `aeson`, I'm ok with making it into own library (exposing few internals `jsstring_` and `scientific` would be useful though).

Notes:
- The token stream doesn't contain commas `,` or colons `:` as they are redundant, at makes things a bit more straightforward.
- Variant with `[Token]` seems to be faster then fusing the structures rogether. The decoder variant as in https://github.com/well-typed/binary-serialise-cbor/blob/0236307ba953d8d2a8351455587e4c12b813397d/cborg/src/Codec/CBOR/Decoding.hs could be an improvement too (we could have different decoder actions to read number as a bytestring/text for example. I should try and benchmark that variant too.

Differences: 
- slightly better error reporting: #558
```haskell
> decodeValue "{ 1 "
Left "Expecting record key or '}', got token TkError \"Expecting record key or '}', got 1 \""
```
- slightly better performance
- cannot separate prime-less and primeless versions (`decode` & `decode'`)

Further work
- Write version consuming strict `ByteString`.
- Should this be default or not?
- It's possible to make a variant of `TokenStream` which rules out invalid streams. I think it's possible to have an encoding such for `TokenStream` without `TkError` (which can be enforced with a GADT parameter, but I don't see a need to do so), one could write total ` TokenStream -> Value` function.
This would require `GADTs,`, `DataKinds` and `TypeFamilies`. This encoding might be useful for hand-written "fancy" `TokenStream` consumers.
- Build a parser consuming `TokenStream` (the list version). First iteration could be done using `parsec`. Probably for little more performance, one would need specialized version at the end.
   - parsing records a little problem:
         - tag-encoding of sum (one would need to retain the pointer to the point where object tokens started). Similarly monadic parsers are problematic
         - record's unordered nature is a problem too, but not impossible
    - as a con, we can avoid building hashmaps and vectors when we don't need to (e.g. parsing `Map`!)
    - makes possible to write parsers taking advantage of the order of keys in JSON!
    - makes it simpler to write record parser disallowing keys (will fail earlier)
    - skipping parser is trivial (`decode` and `decode'` separating might not be needed!)
  

## Benchmarks

my machine is surprisingly noisy right now:

```
benchmarking decode/en/aeson/lazy
[...]

PR #563, opened 2017-07-12
Benchmarks fix #562

### LINKED ISSUE #562: aeson-benchmark-typed broken
Caused by renaming of the field from `id` to ` id_`.

What's the better way to fix it:
- rename back to `id`
- make `Twitter.Options` with `Options` dealing with that special case? (I prefer this)

Ping @bergmark

phadej:
Includes (hopefully ok) changes from #560.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ dist-newstyle
 .cabal-sandbox/
 cabal.sandbox.config
 .stack-work/
+.stack-work-bench/
 
 *.o
 *.hi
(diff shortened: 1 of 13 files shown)


phadej:
Also added typed decode benchmarks, as seems there weren't any:

screenshot from 2017-07-13 16-15-46

last two TH baseline benchmarks are severely inflated.

Otherwise that looks good. Generic parsing is only slightly slower than manual or TH for twitter data.

bergmark:
Cool, thanks a lot!

PR #567, opened 2017-07-18
Export jstring_ and scientific parsers

phadej:
This will let me play with `TokenStream` stuff independently of `aeson`.

### DIFF
--- a/Data/Aeson/Parser.hs
+++ b/Data/Aeson/Parser.hs
@@ -33,6 +33,7 @@ module Data.Aeson.Parser
       json
     , value
     , jstring
+    , scientific
     -- * Strict parsers
     -- $strict
     , json'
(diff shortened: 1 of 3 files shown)


bergmark:
Thanks!

If we consider these "internal", should we really export them from `Data.Aeson.Parser`?

phadej:
@bergmark `Parser.Internal` is not exported module.

bergmark:
Yeah, I meant to export Parser.Internal instead, sorry for not being explicit.

Since we already have the `Internal` convention of things we don't version I'd vote to keep that even if the name doesn't really match, to avoid confusion. `jsstring_` would be weird to have in the public API both based on the naming and implementation O_o. What do you think?

phadej:
I think
- `scientific` would be great to have in a public (versioned) API
- `jstring` (only one `s`!) can stay only in `Parser.Internal`, which is exposed.

bergmark:
Sounds like a good plan!

phadej:
amended the commit

Data/Aeson/Parser.hs
```diff
@@ -46,7 +47,7 @@ module Data.Aeson.Parser
 
 import Prelude ()
 
-import Data.Aeson.Parser.Internal (decodeStrictWith, decodeWith, eitherDecodeStrictWith, eitherDecodeWith, json, json', jstring, value, value')
+import Data.Aeson.Parser.Internal (decodeStrictWith, decodeWith, eitherDecodeStrictWith, eitherDecodeWith, json, json', jstring, jstring_, scientific, value, value')
```
bergmark: This probably gives a warning now that jstring_ is unused?

bergmark:
Cheers!

PR #573, opened 2017-08-02
Add INCOHERENT pragma to RecordToPair

### LINKED ISSUE #572: Overlapping instance of `RecordToPairs` should be incoherent
There is a discrepancy between `FromJSON` and `ToJSON` in the special casing of `Maybe`:

- [`FromJSON` marks the relevant instance `INCOHERENT`](https://github.com/bos/aeson/blob/08fcbcd6ed65d9b3d80300e2bd363bfcd76e38bd/Data/Aeson/Types/FromJSON.hs#L1029) (thanks to #470)

- [`ToJSON` marks the instance `OVERLAPPING`](https://github.com/bos/aeson/blob/08fcbcd6ed65d9b3d80300e2bd363bfcd76e38bd/Data/Aeson/Types/ToJSON.hs#L1012)

This seems to be a bug which has not affected users only thanks to another old bug in GHC, making the instance marked `OVERLAPPING` behave incoherently anyway ([ticket 7296](https://ghc.haskell.org/trac/ghc/ticket/7296), also [ticket 9820](https://ghc.haskell.org/trac/ghc/ticket/9820) for a simpler example).

This bug, if it is indeed one, can be fixed before it gets unmasked: make that `OVERLAPPING` instance `INCOHERENT` instead.

Lysxia:
Fixes #572

### DIFF
--- a/Data/Aeson/Types/ToJSON.hs
+++ b/Data/Aeson/Types/ToJSON.hs
@@ -20,6 +20,7 @@
 #endif
 
 #include "overlapping-compat.h"
+#include "incoherent-compat.h"
 
 -- TODO: Drop this when we remove support for Data.Attoparsec.Number
 {-# OPTIONS_GHC -fno-warn-deprecations #-}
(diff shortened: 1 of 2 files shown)


phadej:
@Lysxia could you add  a test case which this PR makes pass?

Lysxia:
I added a test.

bergmark:
Thanks!

PR #582, opened 2017-08-31
Make Options fields appear in documentation

### LINKED ISSUE #576: Documentation of Option fields disappeared
The fields of `Option`, and thus their individual descriptions, are no longer visible in the docs:

https://hackage.haskell.org/package/aeson-1.2.1.0/docs/Data-Aeson-Types.html#t:Options

It would also be nice to export `Option` and its fields in `Data.Aeson`.

Lysxia:
Fixes #576


bergmark:
Thanks!

PR #584, opened 2017-09-01
FromJSON and ToJSON instances for CTime

Daniel-Diaz:
There are other types in `Foreign.C.Types` that could benefit from having these instances, but at the moment I have only added instances for `CTime`. Mainly because I needed it in a project. The instances are based on the `Int64` instances.


bergmark:
Thank you!

PR #586, opened 2017-09-08
Fix parentheses around example in documentation

pcrama:
Hi,

I noticed a minor error in the documentation of Aeson and fixed it.

Best regards and thanks for this library,

Philippe Crama


bergmark:
Thank you!

PR #587, opened 2017-09-12
Update package versions in stack configurations

Lysxia:
Some metadata fixing on Hackage broke stack builds.


bergmark:
Thanks! This conflicted with my quickcheck PR so I upadted accordingly in 3a77854

PR #583, opened 2017-08-31
Extend omitNothingFields for Option newtype

Lysxia:
Resolves #498

PR #588, opened 2017-09-14
Refactor TH for toJSON and toEncoding

Lysxia:
HLint was complaining about code duplication for #583. This fixes it.


bergmark:
Thanks!

I tried to naively resolve the conflicts against master but that failed, would you mind rebasing this? You should be able to remove the stack commit as well.

It might also be interesting to run the benchmarks to make sure there isn't a performance regression.

Lysxia:
Okay! I'll take a look at benchmarks.

Lysxia:
[Benchmarks](https://gist.github.com/Lysxia/e388c13db32c92866918a0e55b46df19), again with `benchmarks/AesonCompareAutoInstances.hs`: side by side, current `master` on the left, this PR on the right. GHC 8.0.2, lts-9.3

Apart from `encode/th`, the optimized core of everything else is actually identical, yet there are noticeable (10%) speedups (and a few slow downs) in various entries that I can reproduce, so it doesn't seem like it's just noise.
I'm not sure what I'm missing. This phenomenon was also observed before in the generics refactoring.

bergmark:
Thanks for checking the benchmarks! And sorry for the delayed response.

PR #589, opened 2017-09-17
Add a helper for decoding embedded json strings

jessekempf:


bergmark:
Thanks for the contribution! I'm very sorry that I missed looking at this earlier.

I'll add a test case for this now, then it should be good to merge!

Data/Aeson.hs
```diff
@@ -220,6 +224,15 @@ eitherDecodeStrict' =
   eitherFormatError . eitherDecodeStrictWith jsonEOF' ifromJSON
 {-# INLINE eitherDecodeStrict' #-}
 
+-- | Decode a nested JSON-encoded string.
+embeddedJSON :: (FromJSON a) => String -> (Value -> Parser a)
+embeddedJSON str = withText str (either fail return . eitherDecode . fromStrict . T.encodeUtf8)
+  where
+#if MIN_VERSION_bytestring(0, 9, 2)
```
bergmark: `L.fromChunks . (:[])` (or `L.fromChunks . pure`) works for all versions of bytestring so there is no need for CPP.

Data/Aeson.hs
```diff
@@ -220,6 +224,15 @@ eitherDecodeStrict' =
   eitherFormatError . eitherDecodeStrictWith jsonEOF' ifromJSON
 {-# INLINE eitherDecodeStrict' #-}
 
+-- | Decode a nested JSON-encoded string.
+embeddedJSON :: (FromJSON a) => String -> (Value -> Parser a)
```
bergmark: Accepting a `Value` here is a bit strange since it will only suceed for strings, right? Is there a reason for it? A usage example would be helpful.

Data/Aeson.hs
```diff
@@ -120,6 +122,7 @@ module Data.Aeson
     , (.:!)
     , (.!=)
     , object
+    , embeddedJSON
```
bergmark: To match the other functions I think this function should be placed in `Data.Aeson.Types.FromJSON` with the name `withEmbeddedJSON`.

Data/Aeson.hs
```diff
@@ -220,6 +224,15 @@ eitherDecodeStrict' =
...
+-- | Decode a nested JSON-encoded string.
+embeddedJSON :: (FromJSON a) => String -> (Value -> Parser a)
+embeddedJSON str = withText str (either fail return . eitherDecode . fromStrict . T.encodeUtf8)
+  where
+#if MIN_VERSION_bytestring(0, 9, 2)
+  fromStrict = L.fromChunks . (:[])
+#else
+  fromStrict = L.fromStrict
+#endif
 -- $use
```
bergmark: I guess there should also be an INLINE pragma like the rest of the functions have, since the implementation is just composition it should be fine to do.

bergmark:
I noticed a few things while I was starting to write the test, do you have time to take a look?

Data/Aeson.hs
jessekempf: It should take an aeson `String` instead?

Data/Aeson.hs
jessekempf: But `L.fromStrict` doesn't and it's more complex than just `L.fromChunks . (:[])`. So the code should use it except when it's unavailable.

Data/Aeson.hs
jessekempf: Will do.

Data/Aeson.hs
jessekempf: Will do.

Data/Aeson.hs
bergmark: There is no `String` type in aeson, so it would take a `Text` instead.

jessekempf: Right, that's what I meant. Sorry, I'm running a bit sleep-deprived at the moment.

Data/Aeson.hs
bergmark: Do you mean it's more performant? They should both be O(1) since there is only one chunk, right? Is there a measured performance improvement for it (doesn't have to be within aeson)?

I'd like to avoid CPP, but if we keep it please put it in an unexposed `Data.Aeson.Compat` module,  I've wanted to move most our CPP in there.

jessekempf: https://hackage.haskell.org/package/bytestring-0.10.0.2/docs/src/Data-ByteString-Lazy.html#toStrict

The version of `fromStrict` that ByteString provides is performance-optimized, yes. It uses no extra space beyond the output bytestring and copies the data over with `memcpy()`.

`Data.Aeson.Compat` sounds like a decent compromise. Literally the only reason we're having this discussion is because of GHC 7.4.2.

bergmark:
Thanks! I'll take another look tonight.

Data/Aeson/Types/FromJSON.hs
```diff
@@ -665,6 +668,16 @@ withBool _        f (Bool arr) = f arr
...
+-- | Decode a nested JSON-encoded string.
+withEmbeddedJSON :: (FromJSON a) => String -> (Value -> Parser a) -> Value -> Parser a
+withEmbeddedJSON _ innerParser (String txt) =
+    either fail innerParser $ eitherDecode (Compat.fromStrict $ T.encodeUtf8 txt)
+    where
+        eitherDecode = eitherFormatError . eitherDecodeWith jsonEOF ifromJSON
+        eitherFormatError = either (Left . uncurry formatError) Right
+withEmbeddedJSON name _ v = typeMismatch name v
+{-# INLINE withEmbeddedJSON #-}
+
```
bergmark: I suppose the inlining is less of a given now. As a rule of thumb I only inline one-clause functions without benchmarking. But since this is the pattern of all these combinators let's keep the inline!

Data/Aeson/Compat.hs
```diff
@@ -0,0 +1,16 @@
...
+
+module Data.Aeson.Compat
+  (
+    fromStrict
+  ) where
+
+import qualified Data.ByteString      as S
+import qualified Data.ByteString.Lazy as L
+
+fromStrict :: S.ByteString -> L.ByteString
```
bergmark: This can definitely be inlined!

bergmark:
Thanks for the update!
I'm not sure if you missed it or didn't finish yet but my remaining commend was "Accepting a Value here is a bit strange since it will only suceed for strings, right? Is there a reason for it? A usage example would be helpful."

Data/Aeson/Compat.hs
jessekempf: Won't the compiler decide to inline the function automatically?

bergmark: I hope so! But I'm not sure...

jessekempf: If we're talking about GHC, the documentation says this:

>GHC (with -O, as always) tries to inline (or “unfold”) functions/values that are “small enough,” thus avoiding the call overhead and possibly exposing other more-wonderful optimisations. GHC has a set of heuristics, tuned over a long period of time using many benchmarks, that decide when it is beneficial to inline a function at its call site. The heuristics are designed to inline functions when it appears to be beneficial to do so, but without incurring excessive code bloat. If a function looks too big, it won’t be inlined, and functions larger than a certain size will not even have their definition exported in the interface file. Some of the thresholds that govern these heuristic decisions can be changed using flags, see -f*: platform-independent flags.

>Normally GHC will do a reasonable job of deciding by itself when it is a good idea to inline a function. However, sometimes you might want to override the default behaviour. For example, if you have a key function that is important to inline because it leads to further optimisations, but GHC judges it to be too big to inline.

https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html?inline-and-noinline-pragmas#inline-and-noinline-pragmas.

jessekempf:
@bergmark: Here's a usage example:

```
data S3Event =
  S3TestEvent {
    s3eTimestamp :: UTCTime,
    s3eBucketId  :: Text,
    s3eRequestId :: Text,
    s3eHostId    :: Text
  } | S3PutObject {
    s3ePut :: S3PutObjectRecord
  } deriving Show

data S3PutObjectRecord = S3PutObjectRecord {
  s3porTimestamp   :: UTCTime,
  s3porPrincipalId :: Text,
  s3porBucket      :: Text,
  s3porObjectKey   :: Text,
  s3porObjectSize  :: Int64
} deriving Show

instance FromJSON S3Event where
  parseJSON v =
    withObject "S3 Test Event" (\o ->
      S3TestEvent
        <$> (iso8601parse =<< o .: "Time")
        <*> o .: "Bucket"
        <*> o .: "RequestId"
        <*> o .: "HostId"
    ) v <|> withObject "S3 Events" (\o ->
      o .: "Records" >>= withArray "Records" (\records ->
        case toList records of
          [record] -> S3PutObject <$> parseJSON record
          _        -> fail "S3 Events must have only a single record"
      )
    ) v <|> withObject "SNS S3 Events" (\o ->
      o .: "Message" >>= withEmbeddedJSON "Message" (parseJSON :: Value -> Parser S3Event)
    ) v

instance FromJSON S3PutObjectRecord where
  parseJSON = withObject "S3 Event Record" $ \o ->
    S3PutObjectRecord
      <$> (iso8601parse =<< o .: "eventTime")
      <*> (o .: "userIdentity" >>= (.: "principalId"))
      <*> (o .: "s3" >>= (.: "bucket") >>= (.: "name"))
      <*> (o .: "s3" >>= (.: "object") >>= (.: "key"))
      <*> (o .: "s3" >>= (.: "object") >>= (.: "size"))
```

Our code's purpose is to decode SQS messages, specifically ones having to do with S3 object arrivals. It's possible that the SQS messages we get came straight from S3, in case we're all set. It's also possible that the SQS message came through SNS, and in that case the original message will be carried in the SNS message as a string. That string contains the original S3 message serialized to JSON.

We could be getting both kinds of messages on the same SQS queue, so we have to autodetect the message format. If we detect we're dealing with an SNS message we recurse one level by applying ourselves to the embedded JSON string.

bergmark:
Thanks, I think the current type signature makes sense even though it could be more specific. I added a test case so I'll merge if travis is happy!

PR #591, opened 2017-09-24
Doc bug: Missing lambda in haddock comment

jpaugh:
Changing `\` to `\\` fixes this; verified with Haddock version 2.16.1

Occurrences within "bird track" code blocks do NOT have this issue; but occurrences between @'s do.


bergmark:
Thanks!

Issue #593, opened 2017-10-03
UnescapePure is slow (was: and memory hungry)


ndmitchell:
In recent aeson releases my program has started using more memory and taking more time. It seems to coincide with UnescapePure vs UnescapeFFI switch, so I wrote a benchmark. First I modified `aeson` to expose both modules, and then wrote a comparison test harness:

```haskell
{-# OPTIONS_GHC -O2 #-}

import Data.Aeson.Parser.UnescapeFFI as FFI
import Data.Aeson.Parser.UnescapePure as Pure
import qualified Data.Text as T
import qualified Data.ByteString.Char8 as BS
import System.Environment

main = do
    [mode,n] <- getArgs
    let f = case mode of
            "ffi" -> FFI.unescapeText
            "pure" -> Pure.unescapeText
            _ -> error "Unknown mode"
    putStrLn $ either show (show . T.length) $ f $ BS.concat $ replicate (read n) $ BS.pack "\\\""
```

Running `test pure 10000000 +RTS -s` gave:

```
   2,318,823,632 bytes allocated in the heap
   3,241,094,200 bytes copied during GC
     679,727,872 bytes maximum residency (15 sample(s))
     213,781,760 bytes maximum slop
            1713 MB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0      2141 colls,     0 par    5.641s   6.121s     0.0029s    1.7980s
  Gen  1        15 colls,     0 par    0.031s   0.017s     0.0011s    0.0051s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time    0.734s  (  0.770s elapsed)
  GC      time    5.672s  (  6.138s elapsed)
  EXIT    time    0.016s  (  0.180s elapsed)
  Total   time    6.422s  (  7.088s elapsed)

  %GC     time      88.3%  (86.6% elapsed)

  Alloc rate    3,157,547,073 bytes per MUT second

  Productivity  11.7% of total user, 13.4% of total elapsed
```

Running `test ffi 10000000 +RTS -s` gave:

```
     540,061,448 bytes allocated in the heap
     495,318,408 bytes copied during GC
     130,575,192 bytes maximum residency (8 sample(s))
       2,094,248 bytes maximum slop
             297 MB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
[...]

PR #595, opened 2017-10-06
Memory usage improvements in unescapeText. #593

jprider63:
@eskimor

### DIFF
--- a/pure/Data/Aeson/Parser/UnescapePure.hs
+++ b/pure/Data/Aeson/Parser/UnescapePure.hs
@@ -170,6 +170,7 @@ unescapeText' bs = runText $ \done -> do
         (st', p) ->
             return (pos, StateUtf st' p)
 
+      loop :: A.MArray s -> (Int, State) -> Int -> ST s (Int, State)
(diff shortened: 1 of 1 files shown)


pure/Data/Aeson/Parser/UnescapePure.hs
```diff
@@ -172,11 +170,11 @@ unescapeText' bs = runText $ \done -> do
...
             return (pos, StateUtf st' p)
 
-      {-# INLINE runUtf #-}
-
-      f' dest m c = m >>= \s -> f dest s c
-
-      {-# INLINE f' #-}
+      loop _ ps i | i >= len = return ps
+      loop dest ps i = do
+        let c = B.index bs i -- JP: We can use unsafe index once we prove bounds with Liquid Haskell.
```
eskimor: Access fine because of line 173 wit definition of len on line 160.
i is checked to smaller than the len in 173, so B.index should not throw.

Also bs is not an argument to loop (I just confused it with ps) - so it never changes.

bergmark:
Great, thank you both!

PR #596, opened 2017-10-08
Optimize TH toEncoding

Lysxia:
[Benchmarks](https://gist.github.com/Lysxia/7f4bab4856a42bf74a175f601c9a24a4)

`AesonCompareAutoInstances` is the one that is in `benchmarks/`, just to check that I didn't introduce any performance regression. The `SmallRecord` file compares results with a record having 5 fields.

The speedup is in the `BigRecord/encode/th` case. 

The reason for the slowdown was that we were using a recursive function to convert lists of pairs (whose size is known statically) to a comma-separated encoding.

### DIFF
--- a/Data/Aeson/Types/ToJSON.hs
+++ b/Data/Aeson/Types/ToJSON.hs
@@ -50,6 +50,8 @@ module Data.Aeson.Types.ToJSON
     , contramapToJSONKeyFunction
     -- * Object key-value pairs
     , KeyValue(..)
+    , KeyValuePair(..)
+    , FromPairs(..)
     -- * Functions needed for documentation
     -- * Encoding functions
     , listEncoding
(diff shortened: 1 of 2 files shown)


Data/Aeson/Types/ToJSON.hs
```diff
@@ -2716,20 +2718,24 @@ packChunks lbs =
...
-    gPair :: String -> v -> kv
+-- | Like 'KeyValue' but the value is already converted to JSON
+-- ('Value' or 'Encoding'), and the result actually represents lists of pairs
+-- so it can be readily concatenated.
+class Monoid kv => KeyValuePair v kv where
+    pair :: String -> v -> kv
 
-instance ToJSON v => GKeyValue v (DList Pair) where
-    gPair k v = DList.singleton (pack k .= v)
+instance (v ~ Value) => KeyValuePair v (DList Pair) where
```
bergmark: Is there a reason this isn't `KeyValuePair Value (DList Pair)` ?

Lysxia: It improves type inference. Otherwise the TH deriving results in ambiguous type variables and we would need more type annotations.

PR #600, opened 2017-10-29
[acc-errors] Accumulate errors for Seq and Vector parsers

bergmark:
Naming has room for improvements, maybe we should even have a module for accumulation related functions, or should we do this in some other way like using a newtype so we can have proper traversable instances (not sure if that'll work out)?

Ping @Lysxia

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -25,8 +25,12 @@
 
 module Data.Aeson.Types.Internal
     (
+      accumulateSequenceList
+    , accumulateTraverseList
+    , accumulateTraverseVector
+
     -- * Core JSON types
-      Value(..)
+    , Value(..)
     , Array
     , emptyArray, isEmptyArray
     , Pair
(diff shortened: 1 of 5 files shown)


Data/Aeson/Types/Internal.hs
```diff
@@ -349,6 +353,20 @@ liftP2 f pa pb = Parser $ \path kf ks ->
...
+accumulateSequenceList = accumulateTraverseList id
+
+accumulateTraverseList :: (a -> Parser b) -> [a] -> Parser [b]
+accumulateTraverseList f s = case s of
+  []    -> pure mempty
+  h : t -> (:) <$> (f h) <*>+ (accumulateTraverseList f t)
+
+accumulateTraverseVector :: (a -> Parser b) -> Vector a -> Parser (Vector b)
+accumulateTraverseVector f v =
+    if V.null v
```
phadej: I wonder, if it makes sense to define

```haskell
newtype AccParser a = AccParser { runAccParser :: Parser a }

instance Applicative AccParser where
    AccParser f <*> AccParser x = AccParser (f <*>+ x) -- !!!
```

and use it, and not re-implement `traverse` for everything.

Lysxia: I agree, it's a good idea to use `Traversable` here. Reexporting `Parser` versions that use `AccParser` under the hood is nice too.

bergmark:
I'm not sure when I'll get around to playing more with this, so feel free to steal the commit if anyone wants to keep going before I get the chance.

bergmark:
I changed this to use an `AccParser` newtype, much nicer!

PR #602, opened 2017-10-31
Use nix for hlint, change travis language to nix

### LINKED ISSUE #598: Run hlint on travis using precompiled binary
This should speed up that build step quite a bit, instructions for using the binary are available here https://github.com/ndmitchell/hlint#running-with-continuous-integration but note that i'd still like it to be run with `make lint` (or in some other way that doesn't add duplication) since we need to pass along a bunch of options.

troglotit:
I hacked a little bit, tried to use https://github.com/ndmitchell/neil/blob/master/misc/travis.sh - Neil's script, but it had problems with BSD `sed`; thought about doing something with `perl`, but then used `nixpkgs`. I don't think that bringing new dependency is the right way, but on the other hand I'd say it is production ready.

But maybe I'm just overthinking 😅 
Closes #598

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 # This file has been modified from generated version
 # see https://github.com/hvr/multi-ghc-travis
-language: c
+language: nix
 sudo: false
 
 # Bryan threatens to turn off Travis-CI if he gets any email :-)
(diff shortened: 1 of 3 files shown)


bergmark:
Thanks for the contribution! And sorry for the late reply.

I don't mind using nix on travis, it should keep hlint fairly recent. But is there a way to keep the makefile nix-free? Otherwise everyone who wants to run this locally also needs nix-shell.

PR #604, opened 2017-11-17
Add benchmarks for derived instances

Lysxia:
For some reason AesonCompareAutoInstances was not in the aeson-benchmarks package. I also split the file so that each type and derivation strategy (generic/TH) can be more easily tested in isolation.


bergmark:
Thanks!

PR #561, opened 2017-07-11
don't unescape text in clear cases

phadej:
After:

```
benchmarking decode/en/aeson/lazy
time                 1.630 ms   (1.606 ms .. 1.655 ms)
                     0.998 R²   (0.996 R² .. 0.999 R²)
mean                 1.638 ms   (1.626 ms .. 1.664 ms)
std dev              59.77 μs   (35.45 μs .. 104.0 μs)
variance introduced by outliers: 23% (moderately inflated)

benchmarking decode/jp/aeson
time                 2.382 ms   (2.359 ms .. 2.406 ms)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 2.404 ms   (2.392 ms .. 2.431 ms)
std dev              60.54 μs   (36.76 μs .. 112.9 μs)
variance introduced by outliers: 12% (moderately inflated)
```

before

```
benchmarking decode/en/aeson/lazy
time                 2.053 ms   (2.027 ms .. 2.077 ms)
                     0.998 R²   (0.996 R² .. 0.999 R²)
mean                 2.069 ms   (2.056 ms .. 2.098 ms)
std dev              62.97 μs   (34.19 μs .. 105.8 μs)
variance introduced by outliers: 17% (moderately inflated)

benchmarking decode/jp/aeson
time                 2.666 ms   (2.636 ms .. 2.699 ms)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 2.695 ms   (2.672 ms .. 2.735 ms)
std dev              103.1 μs   (64.72 μs .. 159.6 μs)
variance introduced by outliers: 22% (moderately inflated)
```

Even for japanese there seems to be small speedup, because record keys are ASCII

PR #605, opened 2017-11-18
Add flag for benchmarks to select local or external aeson dependency

Lysxia:
Avoids the duplication due to executables integrating two versions of aeson at the same time.

### DIFF
--- a/benchmarks/AesonFoldable.hs
+++ b/benchmarks/AesonFoldable.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE PackageImports #-}
 
 module Main (main) where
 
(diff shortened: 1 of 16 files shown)


bergmark:
I pushed a hlint fix

bergmark:
Thanks!

phadej:
I really dislike this. Now it's way harder to see regressions against released `aeson`. The setup was complicated, but there was a reason.

phadej:
Note: I know that someone (read: me) should have documented how to benchmark aeson, as the setup _is_ complicated, and not just `stack bench`. I would have done that, as I saw the activity around that code, haven't time to go thru notifications earilier this week.

Lysxia:
Is it really harder now? We can build benchmarks with development or released versions of aeson by switching a flag.

phadej:
See e.g. https://github.com/bos/aeson/pull/561#issuecomment-315108405

phadej:
These graphs (and other stats criterion renders html) are very useful, at least for me. Text output isn't as good.

Lysxia:
Okay. I think it would be nicer to make criterion render data from multiple sources than maintaining two copies of the same code.

PR #607, opened 2017-11-18
Remove option stopping simplifier in Dates benchmark

### LINKED ISSUE #606: Compiling aeson-benchmark-dates with GHC 7.8 and 7.10 exhausts simplifier ticks
Like #502, but different GHC versions. Perhaps there is another `INLINE` to kill somewhere.

Lysxia:
Fixes #606


bergmark:
I asked in #606 about why it was there to begin with!

PR #611, opened 2017-12-05
Remove unnecessary constraint in withEmbeddedJSON

### LINKED ISSUE #610: withEmbeddedJSON has an unnecessary FromJSON constraint
This constraint is unused (we take a parser explicitly rather than relying on `parseJSON`).

mithrandi:
Closes #610. I don't think any other changes are needed here besides deleting the constraint.


bergmark:
Thanks! This will be merged into master which tracks the next major release. I would be hesitant about including this in a minor release as we are changing the type of the function, but I can't think of a way this can break existing code.

PR #613, opened 2017-12-13
Fix doc mentioning encodeToBuilder (fixes #612)

### LINKED ISSUE #612: docs of Data.Aeson.Text.encodeToTextBuilder mention "encodeToBuilder"
but that does not seem to exist/be exposed (since 1.*)

i suppose that function would be `encodingToLazyByteString . toEncoding` ?

lspitzner:
is it too late to [ci skip] ? lets see.

### DIFF
--- a/Data/Aeson/Text.hs
+++ b/Data/Aeson/Text.hs
@@ -46,8 +46,8 @@ encodeToLazyText = LT.decodeUtf8 . encodingToLazyByteString . toEncoding
 -- embedded efficiently in a text-based protocol.
 --
 -- If you are going to immediately encode straight to a
--- 'L.ByteString', it is more efficient to use 'encodeToBuilder'
--- instead.
+-- 'L.ByteString', it is more efficient to use 'encode' (lazy ByteString)
+-- or @'fromEncoding' . 'toEncoding'@ (ByteString.Builder) instead.
 --
 -- /Note:/ Uses 'toJSON'
 encodeToTextBuilder :: ToJSON a => a -> Builder


lspitzner:
dang it :p i never think of that before i press "commit".

Data/Aeson/Text.hs
```diff
@@ -46,8 +46,9 @@ encodeToLazyText = LT.decodeUtf8 . encodingToLazyByteString . toEncoding
 -- embedded efficiently in a text-based protocol.
 --
 -- If you are going to immediately encode straight to a
--- 'L.ByteString', it is more efficient to use 'encodeToBuilder'
--- instead.
+-- 'L.ByteString', it is more efficient to use
+-- @'encodingToLazyByteString' . 'toEncoding'@ (lazy ByteString) or
```
phadej: that's is just `encode` from `Data.Aeson`.

lspitzner: eh.. right

lspitzner:
better?

phadej:
that's up to @bergmark, I'm just an observer :)

bergmark:
I would add you as a maintainer if i could @phadej :) Bryan suggested we move the repo to the haskell org a while back, hopefully that will happen at some point...

Thanks @lspitzner!

hvr:
@phadej @bergmark we'd need @bos to initiate the transfer, and I gave him all necessary github-privs for that when `text` was migrated to the Haskell org already

PR #617, opened 2017-12-27
Remove some CPP by using recent versions of text

RyanGlScott:
Recent versions of `text` expose the `encodeUtf8BuilderEscaped` function on all versions of `bytestring` (see [this pull request](https://github.com/haskell/text/pull/155)). This means we can remove the hack put in place in https://github.com/bos/aeson/pull/426#issuecomment-225358139 wherein we redefine `encodeUtf8BuilderEscaped` locally in `aeson` if old `bytestring`s are used, allowing us to axe some CPP.

### DIFF
--- a/Data/Aeson/Encoding/Builder.hs
+++ b/Data/Aeson/Encoding/Builder.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE CPP #-}
 
 -- |
 -- Module:      Data.Aeson.Encoding.Builder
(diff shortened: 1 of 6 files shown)


hvr:
wow... you're picking up a lower bound on `text >= 1.2.3` quite fast... which is  good thing I guess, given the rather nasty bug hidden in `text < 1.2.3` :-)

RyanGlScott:
Actually, now that I look closer, the corresponding fix in `text` has been in place [since `1.2.2.0`](https://github.com/haskell/text/commit/6c4cc7f6ba77a72c7ce9d07fce67b0e67f1623ee), so we _could_ go lower than `1.2.3.0` if desired.

hvr:
@RyanGlScott well, do you really want to use 1.2.2 given https://github.com/haskell/text/issues/197 ;-)

RyanGlScott:
Well, we've been using it for a while in `aeson` without observing that bug, right? But if it's really that severe, I could limit the lower bound to `1.2.3`.

PR #618, opened 2017-12-28
Improve error message on day parse fail

gbaz:
Before:

    *> parse utcTime "asd"
    Fail "asd" [] "Failed reading: takeWhile1"

After:

    *> parse utcTime "asd"
    Fail "asd" [] "Failed reading: date must be of form [+,-]YYYY-MM-DD"

There's probably a better don't-repeat-yourself refactor here of the error message, but my parsec-fu is not strong.

This is important in particular because this parser's error messages often get sent directly to users when serving json APIs via aeson.


bergmark:
Thanks! I agree that error messages are really important so this is great stuff.

This will be merged into master which is currently tracking the `1.3.0.0` release, it may take a while before that gets shipped, If someone would like this released sooner please file a PR against the `1.2.x.x` branch and I can make a separate release.

Issue #623, opened 2018-02-22
Export something similar to Yaml's encodeFile/decodeFile


Pastafarianist:
`yaml` has two handy helpers called [`encodeFile`](https://hackage.haskell.org/package/yaml-0.8.2/docs/Data-Yaml.html#v:encodeFile) and [`decodeFile`](https://hackage.haskell.org/package/yaml-0.8.2/docs/Data-Yaml.html#v:decodeFile). It would be convenient if `aeson` exported something similar as well.

phadej:
If someone makes a PR, please use `strict` variants of `encode` / `decode` :)

bergmark:
What is the preferred way to do this nowadays, is using base good? or good enough?

bergmark:
#625

PR #625, opened 2018-03-01
Add encodeFile as well as decodeFile* variants

markus1189:
Saw #623 in the HaskellWeekly newsletter and gave it a shot.

Please review ;)

### DIFF
--- a/Data/Aeson.hs
+++ b/Data/Aeson.hs
@@ -38,11 +38,16 @@ module Data.Aeson
     , eitherDecode
     , eitherDecode'
     , encode
+    , encodeFile
     -- ** Variants for strict bytestrings
     , decodeStrict
+    , decodeFileStrict
     , decodeStrict'
+    , decodeFileStrict'
     , eitherDecodeStrict
+    , eitherDecodeFileStrict
     , eitherDecodeStrict'
+    , eitherDecodeFileStrict'
     -- * Core JSON types
     , Value(..)
     , Encoding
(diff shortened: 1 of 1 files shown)


bergmark:
Thanks! Looks good.

@phadej mentioned using the strict encoder as well, but do we even have such a thing??

Data/Aeson.hs
```diff
@@ -169,6 +173,18 @@ decodeStrict :: (FromJSON a) => B.ByteString -> Maybe a
...
+-- If this fails due to incomplete or invalid input, 'Nothing' is
+-- returned.
+--
+-- The input file's content must consist solely of a JSON document,
+-- with no trailing data except for whitespace.
+--
+-- This function parses immediately, but defers conversion.  See
+-- 'json' for details.
+decodeFileStrict :: (FromJSON a) => FilePath -> IO (Maybe a)
+decodeFileStrict = fmap decodeStrict . B.readFile
```
phadej: I think `either` variants will be useful as well.

markus1189:
I added the `either` versions. :+1:

bergmark:
Great!

I think I'll cut the 1.3.0.0 release soon, I've been holding it off too long.

bergmark:
I just noticed, these functions need to be exported as well. Maybe travis should build with -Werror for the latest GHC to spot these things.

markus1189:
yeah that was dumb, ;)

phadej:
Writing tests also helps!

markus1189:
@phadej True :smile: I can add some tests, no problem, although I was too dumb to find tests for things like `decodeStrict'` and therefore didn't spend too much effort on it.

bergmark:
👍 
I'm fine merging this without tests, but adding them in another PR would be very welcome.

PR #629, opened 2018-03-27
Fix generic `FromJSON` for non-record constructors with one field

### LINKED ISSUE #627: Genericly derived FromJSON instance incorrectly reports structural error
The following code *should* in my opinion work.

However the output is not `success` but this:

```
"0"
Error in $: expected Int, encountered Object
```

### Code for the example:

```hs
import Data.Aeson
import GHC.Generics

newtype MyType = MyType Int deriving (Eq, Show, Generic)


baseOptions :: Options
baseOptions =
    defaultOptions
        {unwrapUnaryRecords = True, fieldLabelModifier = camelTo2 '_'}


instance ToJSON MyType where
    toJSON = genericToJSON baseOptions
    toEncoding = genericToEncoding baseOptions

instance FromJSON MyType where
    parseJSON = genericParseJSON baseOptions


main =
    let example = MyType 0
        asJSON = encode example
     in do print asJSON
           case eitherDecode asJSON of
               Left err -> putStrLn err
               Right v
                   | v == example -> putStrLn "success"
                   | otherwise -> print v
```

### A little more explanation

I started to encounter this, if not erroneous, unintuitive behaviour after I removed the fields from some newtypes. Meaning before I had structures like these:

```hs
newtype MyType = MyType { unwrapMyType :: Int } deriving (Eq, Show, Generic)
```

They had the same instances for `ToJSON` and `FromJSON` with the same options, the only things I changed was removing the accessor for the field. Furthermore they were compiled with the same `aeson` version. 

Also adding an accessor back into the newtype declaration makes it run as expected again!

Observed with `aeson` version `1.2.3.0`

Lysxia:
Fixes #627

Also some refactoring to get rid of `isUnary`.

### DIFF
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -1004,7 +1004,7 @@ instance ( IsRecord            f isRecord
 
 instance OVERLAPPING_
          ( GFromJSON arity a, FromRecord arity (S1 s a)
-         ) => ConsFromJSON' arity (S1 s a) 'True where
+         ) => ConsFromJSON' arity (S1 s a) True where
     consParseJSON' opts fargs
       | unwrapUnaryRecords opts = Tagged . gParseJSON opts fargs
       | otherwise = Tagged . withObject "unary record" (parseRecord opts fargs)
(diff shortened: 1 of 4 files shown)


bergmark:
Thanks!

PR #630, opened 2018-04-05
Allow base-compat-0.10.0

phadej:


phadej:
we could use `base-compat-batteries` too, I can make a PR. That could cleanup `.cabal` a little.

RyanGlScott:
I'm waiting for `aeson` to support `base-compat-0.10` before I can upgrade some of my packages, so this would definitely be nice to land.

PR #634, opened 2018-04-22
Switch to tasty

Lysxia:
It turns out it was just a matter of renaming.


bergmark:
Nice :)

bergmark:
@Lysxia i think test failures got obfuscated with this change, can that be improved?

```
    ErrorMessages
      Int:                            FAIL
        Exception: HUnitFailure (Just (SrcLoc {srcLocPackage = "main", srcLocModule = "ErrorMessages", srcLocFile = "/Users/adam/repos/hs/aeson/tests/ErrorMessages.hs", srcLocStartLine = 70, srcLocStartCol = 15, srcLocEndLine = 70, srcLocEndCol = 56})) (ExpectedButGot (Just "Invalid error message") "\"Error in $: expected Int1, encountered String\"" "\"Error in $: expected Int, encountered String\"")
```

Lysxia:
Surprisingly, [*tasty-hunit*](https://hackage.haskell.org/package/tasty-hunit) provides its own repackaging of *HUnit* (note that it does not depend on *HUnit*). In particular, the *tasty-hunit* provider only catches its own flavor of `HUnitException`. I'll try to replace the *HUnit* imports with their *tasty-hunit* equivalent.

PR #635, opened 2018-04-28
Haddock on Parser

tomjaguarpaw:
To explain that it might not be what users typically think of as a "parser"


bergmark:
Thanks! Let's merge this, if anyone else has suggestions we can always iterate.

PR #639, opened 2018-05-10
Replace HUnit tests with tasty-hunit

Lysxia:
Following up on #634.

With `tasty-hunit-0.10.0.1`, HUnit errors are now pretty-printed like this:

```
tests
  unit
    ErrorMessages
      Int:                                                  FAIL
        tests/ErrorMessages.hs:33:
        [error message here]
        expected: Just ()
         but got: Nothing
```


bergmark:
Thanks! It seems like i broke the benchmarks so don't worry if they fail, i have a fix for that in the queue.

PR #647, opened 2018-06-13
Export `withBoundedScientific`

qrilka:
http://hackage.haskell.org/package/aeson-1.4.0.0/docs/Data-Aeson.html#v:withScientific advises to use `withBoundedScientific` for unbounded types but the function wasn't exported

### DIFF
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -52,6 +52,7 @@ module Data.Aeson.Types.FromJSON
     , withObject
     , withText
     , withArray
+    , withBoundedScientific
     , withScientific
     , withBool
     , withEmbeddedJSON


bergmark:
Thanks,

What is your use case for `withBoundedScientific`? I'm interested in knowing if exporting it is the best way forward or if there are other solutions we should encourage.

There are some things do do/discuss:
* Adding bounds on a type during deserialization is not ideal if you control the underlying type. I think it's preferable to use a smart constructor in combination with `withScientific` instead (implementing the bounds check yourself is easy). Would that work for your use case?
* Do people want a different bound? It's easy to parameterize it, so it's probably best to do that before adding the export.
* Currently we only have an upper bound on the exponent. @phadej suggested we add a lower bound for converting to e.g. `Decimal`. A lower bound isn't currently needed inside aeson but adding it wouldn't affect the current instances.
* If we export it, I think it should be exported from all modules that export `withScientific`.

PR #648, opened 2018-06-22
IsRecord has no methods

ggreif:
MORE TO COME

fixes warning:
```
Data/Aeson/Types/Generic.hs:28:7: warning: [-Wdodgy-exports]
    The export item `IsRecord(..)' suggests that
    `IsRecord' has (in-scope) constructors or class methods,
    but it has none
```

### DIFF
--- a/Data/Aeson/Types/Generic.hs
+++ b/Data/Aeson/Types/Generic.hs
@@ -25,7 +25,7 @@
 
 module Data.Aeson.Types.Generic
     (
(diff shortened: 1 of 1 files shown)


bergmark:
iirc this is an issue only for older ghcs, but if CI passes then i’m wrong :)

ggreif:
Oh, I was seeing those warnings when compiling with GHC HEAD (v8.7)

PR #649, opened 2018-07-02
fix build

vmchale:
This fixes `aeson` for the latest version of `base-compat`.


phadej:
Could you not reformat the imports, hard to see what's changed

vmchale:
I've made it minimally invasive with the latest commit.

phadej:
I think this is #650 , but that was fixed in `base-compat-0.10.3` (and `base-compat-0.10.2` marked uninstallable). Could you `cabal update` and see if the fix is actually needed at all anymore?

Sorry for not figuring that immediately

vmchale:
No worries. This works with the new `base-compat` so I'll close this PR :)

PR #668, opened 2018-09-11
Handle leap seconds corretly (Fixes #557)

### LINKED ISSUE #557: decode . encode not always id for UTCTime
It is possible to get valid (as in: accepted by the `time` package) `UTCTime` values that can be encoded but not decoded.

Here's the expected behavior:

``` haskell
λ: t :: UTCTime <- parseTimeM True defaultTimeLocale  "%Y-%m-%d %H:%M:%S" "2000-01-01 23:59:59"
λ: encode t
"\"2000-01-01T23:59:59Z\""
λ: decode (encode t) :: Maybe UTCTime 
Just 2000-01-01 23:59:59 UTC
λ: decode "\"2000-01-01T23:59:59Z\"" :: Maybe UTCTime 
Just 2000-01-01 23:59:59 UTC
```

The encoded `UTCTime` retains the same structure (time of day), and it's always possible to decode it. However here's an `UTCTime` with a leap second:

``` haskell
λ: t :: UTCTime <- parseTimeM True defaultTimeLocale  "%Y-%m-%d %H:%M:%S" "2000-01-01 23:59:60"
λ: encode t
"\"2000-01-01T24:00:00Z\""
λ: decode (encode t) :: Maybe UTCTime 
Nothing
λ: decode "\"2000-01-01T23:59:60Z\"" :: Maybe UTCTime 
Just 2000-01-01 23:59:60 UTC
```

The encoding changes the time of day (`23:59:60` -> `24:00:00`) and it's not possible to decode the resulting `ByteString`. Note that it's still possible to decode a hand crafted `ByteString` where the time of day is kept (`23:59:60`).

Taking this further (`23:59:61` or `24:00:01`) it is not possible to decode the `UTCTime` at all:

``` haskell
λ: t :: UTCTime <- parseTimeM True defaultTimeLocale  "%Y-%m-%d %H:%M:%S" "2000-01-01 23:59:61"
λ: encode t
"\"2000-01-01T24:00:01Z\""
λ: decode (encode t) :: Maybe UTCTime 
Nothing
λ: decode "\"2000-01-01T23:59:61Z\"" :: Maybe UTCTime 
Nothing
```

I'm not sure what a solution to this would be. But since `time` allows these `UTCTime`s I think `aeson` should be able to encode and decode them while retaining the original structure.

adamschoenemann:
I've basically just lifted the guard from `Data.Time.LocalTime.timeToTimeOfDay`. This problem has bitten me but has an easy fix :)

### DIFF
--- a/tests/UnitTests.hs
+++ b/tests/UnitTests.hs
@@ -198,7 +198,7 @@ utcTimeGood = do
   assertEqual "utctime" (parseWithRead "%FT%H:%MZ" ts10) t10
   assertEqual "utctime" (parseWithRead "%F %H:%MZ" ts11) t11
   assertEqual "utctime" (parseWithRead "%FT%T%QZ" "2015-01-01T14:30:00Z") t12
-  
+
   -- leap seconds are included correctly
   let ts13 = "2015-08-23T23:59:60.128+00" :: LT.Text
   let (Just (t13 ::  UTCTime)) = parseWithAeson ts13
@@ -226,6 +226,7 @@ utcTimeBad = do
   verifyFailParse "2015-01-01T12:30:00.00+00:00Z" -- no Zulu if offset given
   verifyFailParse "2015-01-03 12:13:00.Z" -- decimal at the end but no digits
   verifyFailParse "2015-01-03 12:13.000Z" -- decimal at the end, but no seconds
+  verifyFailParse "2015-01-03 23:59:61Z"  -- exceeds allowed seconds per day
   where
     verifyFailParse (s :: LT.Text) =
       let (dec :: Maybe UTCTime) = decode . LT.encodeUtf8 $ LT.concat ["\"", s, "\""] in
--- a/stack-lts6.yaml
+++ b/stack-lts6.yaml
@@ -3,7 +3,6 @@ packages:
 - '.'
 - attoparsec-iso8601
 extra-deps:
-- integer-logarithms-1
 - semigroups-0.18.2
 - tagged-0.8.5
 - transformers-compat-0.5.1.4
@@ -13,6 +12,7 @@ extra-deps:
 - integer-logarithms-1
 - quickcheck-instances-0.3.16
 - th-abstraction-0.2.2.0
+- text-1.2.3.0
 flags:
 flags:
   aeson:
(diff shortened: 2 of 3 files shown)


attoparsec-iso8601/Data/Attoparsec/Time/Internal.hs
```diff
@@ -50,11 +50,16 @@ data TimeOfDay64 = TOD {-# UNPACK #-} !Int
...
+posixDayLength :: DiffTime
+posixDayLength = fromInteger 86400
+
 diffTimeOfDay64 :: DiffTime -> TimeOfDay64
-diffTimeOfDay64 t = TOD (fromIntegral h) (fromIntegral m) s
-  where (h,mp) = fromIntegral pico `quotRem` 3600000000000000
-        (m,s)  = mp `quotRem` 60000000000000
-        pico   = unsafeCoerce t :: Integer
+diffTimeOfDay64 t
+  | t >= posixDayLength = TOD 23 59 (60000000000000 + (unsafeCoerce (t - posixDayLength)))
```
phadej: every unsafeCoerce have to be justified.

phadej:
I'd prefer for old `time`:

```
Prelude Data.Time> truncate (realToFrac (secondsToDiffTime 123) * 1000000000000 :: Rational) :: Integer
123000000000000
```

phadej:
Ah, never mind. if we use `unsafeCoerce` for `Pico`, I guess we can for `DiffTime` too...

tests/UnitTests.hs
```diff
@@ -198,6 +198,14 @@ utcTimeGood = do
   assertEqual "utctime" (parseWithRead "%FT%H:%MZ" ts10) t10
   assertEqual "utctime" (parseWithRead "%F %H:%MZ" ts11) t11
   assertEqual "utctime" (parseWithRead "%FT%T%QZ" "2015-01-01T14:30:00Z") t12
+  
+  -- leap seconds are included correctly
+  let ts13 = "2015-08-23T23:59:60.128+00" :: LT.Text
```
phadej: is `23:59:61` however rejected?

adamschoenemann: Should be -- I added a negative test case.

adamschoenemann:
Awesome!

PR #672, opened 2018-09-26
Add instance KeyValue Object

roberth:
Although Object (a.k.a. HashMap Text Value) is typically not exposed
for constructing objects in the aeson interface, this instance does
help when third-party packages demand Object values.


bergmark:
Thanks!

Issue #673, opened 2018-10-09
Possible DoS with very large numbers


lehins:
It looks like parsing doesn't short circuit if input is too large. From what it looks like it is because `Scientific` is arbitrary precision and conversion to basic `Int` happens afterwards. Here is a sample code which simply supplies a long string of `11111..` to the parser:
```haskell
import System.Environment
import Data.Aeson

main :: IO ()
main = do
  [fname] <- getArgs
  eNum <- decodeFileStrict fname
  print (eNum :: Maybe Int)
```
running this demonstrates the problem:
```shell
$ printf '1%.0s' {1..524288} > bignum0.json
$ time ./foo bignum0.json
Nothing

real	0m9.801s
user	0m9.576s
sys	0m0.196s
$ printf '1%.0s' {1..1048576} > bignum1.json
$ time ./foo bignum1.json +RTS -s
Nothing

real	0m40.162s
user	0m39.396s
sys	0m0.720s
$ printf '1%.0s' {1..2097152} > bignum2.json
$ time ./foo bignum2.json
Nothing

real	2m49.058as
user	2m47.100s
sys	0m1.956s
```
It's easy to see with a naked eye that time is quadratic in number of digits being parsed. This could be easily exploited to take down a server that accepts JSON HTTP requests. All an attacker needs to do is submit a dozen bogus transactions with really large amounts.

That was tested on latest versions (with lts-12.4):
* aeson 1.3.1.1
* scientific 0.3.6.2

Only thing I discovered that's related  was this ticket: https://github.com/bos/aeson/issues/198 but that doesn't talk about number of digits, just about huge numbers in scientific notation.

phadej:
yes, scientific notation vs. number digits is different issue.

With current `aeson` design we cannot short-circuit even for obviously large numerals, as we always parse to `Scientific`.

But we shouldn't be quadratic!

---

This looks more like an issue:

> Handling of strings that contain backslash escape sequences is greatly improved. For a pathological string containing almost a megabyte of consecutive backslashes, the new implementation is 27x faster and uses 42x less memory.

fixed in 0.8.1.0

PR #674, opened 2018-10-12
Add issue 673 benchmark

phadej:
screenshot from 2018-10-12 11-59-13

#673 is easily to reproduce, now we need to fix it...

### DIFF
--- a/benchmarks/Issue673.hs
+++ b/benchmarks/Issue673.hs
@@ -1,0 +1,70 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Main (
+    main,
+    input17,
+    input32,
+    input64,
+    input128,
+    input256,
+    input2048,
+    input4096,
+    input8192,
+    input16384,
+  ) where
+
+import Criterion.Main
+import Prelude.Compat
+import Data.Int (Int64)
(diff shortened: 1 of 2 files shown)


phadej:
screenshot from 2018-10-12 12-07-05

The problem is in a `Value` parser, not `Scientific -> Int` conversion.

phadej:
the `scientific` attoparsec parser seems to exercise the quadratic behavior.

phadej:
screenshot from 2018-10-13 10-47-48

The slow part is `BS.foldl' (\a w -> a *  10 +  fromIntegral (w - zero)) 0 digits` where accumulator is an `Integer`. If we accumulate as `Int`, benchmarks will be as fast as `String` - but the results will be wrong

@hvr any idea how to convert `ByteString` of digits into `Integer` in linear time?

phadej:
`base` had a solution for quadratic `Digits -> Integer`

screenshot from 2018-10-13 21-45-17

bergmark:
Nice, thanks! I guess it's useful to ship this even before attoparsec has been updated?

phadej:
@bergmark we don't experience `attoparsec` "bug" anywhere. We do use `decimal` from attoparsec, but only to parse `Int` - which behaves nicely (multiplications are constant operation).

PR #676, opened 2018-10-18
Bumped version of QuickCheck

coot:
Allow QuickCheck-2.12

PR #684, opened 2019-01-02
Bump QuickCheck + base-orphan; faster test compilation

kquick:
The upper bounds for QuickCheck and base-orphan are bumped (similar to https://github.com/bos/aeson/pull/676) but additionally the properties tests are broken into multiple files to speed up compilation and hopefully avoid Travis CI failures.

The discussion of https://github.com/bos/aeson/pull/676 indicated that QuickCheck 2.12 was causing significantly slower compilation times, but local testing for me showed little difference between 2.12.6.1 and 2.11.3 (see the data in the commit comment for commit 3f553c4), but breaking the tests into multiple files improved compilation times for both QuickCheck versions.  All tests pass for all configurations, so there is no indication that there is any validity regression involved.

This commit does not change the API nor does it add any new features, so it should only require a minor version bump (C of A.B.C) per the PVP.


bergmark:
Thanks!

kquick:
Thanks for merging this.  It would be great to get a Hackage release of this.  It looks like one of the trustees bumped the base-orphans version allowing GHC 8.6, but the QuickCheck < 2.12 restriction is creating a lot of issues with other packages and transitive dependencies.

PR #694, opened 2019-04-15
Export isEmptyArray from Data.Aeson.Types

callummance:
Export `isEmptyArray` found in Data.Aeson.Types.Internal from Data.Aeson.Types.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 For the latest version of this document, please see [https://github.com/haskell/aeson/blob/master/changelog.md](https://github.com/haskell/aeson/blob/master/changelog.md).
 
+### Upcoming
+
+* Expose `isEmptyArray` from `Data.Aeson.Types` module.
+
 ### 2.3.0.0 - 2026-05-21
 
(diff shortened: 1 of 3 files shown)


bergmark:
Thanks for the contribution!

My initial thought was that it's a bit weird that we have a function `Value -> Bool`. Just looking at the type signature it could have a lot of implementations. And if `isEmptyArray x == False` you might think you have a non-empty array, but in this case you could also have something else. Seems we only use it in one place inside aeson btw!

Instead, maybe we could have a function `withArrayRenameMe :: a -> (Array-> a) -> Value -> a` (`withArray` is already taken) and then you could write `withArray False V.null` instead to be explicit about what happens if you don't have an Array. I think the type also gives more indication of what the implementation is. Maybe stick several of these functions inside a new module like `Data.Aeson.ValueCombinators`. Maybe `withArray` could be implemented in terms of this new function.

Thoughts?

Lysxia:
Looks reasonable to me.

- It's like `(==) emptyArray` but inlines better
- The concern of biasing people to only think about arrays also makes sense, and I've chosen to mitigate it with a comment about it in the doc.

PR #695, opened 2019-04-17
Add labels to json object and list parsing

### LINKED ISSUE #558: Uninformative invalid JSON error

FintanH:
Fixes #558 

This PR adds better error messages to the parsing of objects and lists.

I'm aware of #559 but this change seems more to the point of giving a better message :)

### DIFF
--- a/Data/Aeson/Parser/Internal.hs
+++ b/Data/Aeson/Parser/Internal.hs
@@ -123,9 +123,9 @@ objectValues str val = do
   -- Why use acc pattern here, you may ask? because 'H.fromList' use 'unsafeInsert'
   -- and it's much faster because it's doing in place update to the 'HashMap'!
   loop acc = do
-    k <- str <* skipSpace <* char ':'
-    v <- val <* skipSpace
-    ch <- A.satisfy $ \w -> w == COMMA || w == CLOSE_CURLY
+    k <- (str A.<?> "expected object key") <* skipSpace <* (char ':' A.<?> "expecting ':'")
+    v <- (val A.<?> "expected object value") <* skipSpace
+    ch <- (A.satisfy $ \w -> w == COMMA || w == CLOSE_CURLY) A.<?> "expecting ',' or '}'"
     let acc' = (k, v) : acc
     if ch == COMMA
       then skipSpace >> loop acc'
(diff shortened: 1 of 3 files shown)


bergmark:
Thanks a lot! 

It would be really nice to have some tests for this, do you think you have time to add them? Probably in `tests/ErrorMessages.hs`. I'm sure it works but I can't really tell from the code :-) I was thinking about doing a release later this week so I can add them then otherwise.

Also hlint is nitpicking a bit
```haskell
Data/Aeson/Parser/Internal.hs:128:11: Suggestion: Move brackets to avoid $
Found:
  (A.satisfy $ \ w -> w == 44 || w == 125) A.<?>
    "expecting ',' or '}'"
Perhaps:
  A.satisfy (\ w -> w == 44 || w == 125) A.<?> "expecting ',' or '}'"
Data/Aeson/Parser/Internal.hs:153:13: Suggestion: Move brackets to avoid $
Found:
  (A.satisfy $ \ w -> w == 44 || w == 93) A.<?>
    "expecting ',' or ']'"
Perhaps:
  A.satisfy (\ w -> w == 44 || w == 93) A.<?> "expecting ',' or ']'"
```

FintanH:
Sure I'll try get it done this week :ok_hand:

FintanH:
Should be good now :smile: :crossed_fingers:

bergmark:
Nice!

The old ghc builds started failing now... looks like we should bump the base compat lower bound to `0.10.1` to get `<>` in scope.

FintanH:
> The old ghc builds started failing now... looks like we should bump the base compat lower bound to 0.10.1 to get <> in scope.

Since these values are definitely `String`s I'll switch it to `++` and it can be backward compat :)

phadej:
@bergmark bumping `base-compat` won't get you `<>`.
See http://hackage.haskell.org/package/base-compat-batteries

I'm neutral on depending on `base-compat-batteries`, I guess `aeson` already depends on all of those dependencies.

bergmark:
Ah, thanks, I misread the base-compat changelog.

I pushed a hopefully final fix, weird that it only affected GHC 7.10.

bergmark:
re -batteries, i prefer it and would be happy to switch to it.

FintanH:
@bergmark: This is weird, if you click "Details" to look at the progress on Travis it all looks good but GitHub doesn't seem to be picking this up :thinking:

bergmark:
I'm guessing there was a hiccup in the travis-github integration. No worries.

PR #696, opened 2019-04-26
Adapt to breaking changes in th-abstraction-0.3.0.0

RyanGlScott:
This adapts to a breaking change in the type of `datatypeVars` that was introduced in `th-abstraction-0.3.0.0`.


bergmark:
Thanks!!

PR #697, opened 2019-05-01
Support GHC 8.8

fumieval:
Pardon an automatically-posted pull request; this is a part of [Operation Vanguard](https://github.com/haskell-vanguard/haskell-vanguard). Note that this does not guarantee that tests and benchmarks are buildable.

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -158,8 +158,10 @@ instance Monad IResult where
     IError path err >>= _ = IError path err
     {-# INLINE (>>=) #-}
 
+#if !MIN_VERSION_base(4,13,0)
     fail = Fail.fail
     {-# INLINE fail #-}
+#endif
 
 instance Fail.MonadFail IResult where
     fail err = IError [] err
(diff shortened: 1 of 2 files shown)


bergmark:
Thanks! Cool project, so vanguard verifies that the library builds? Is there a reason you don't build tests and benchmarks? (Running them could be more complicated so that makes sense).

We should add ghc 8.8 to the travis CI job before merging this, looks like it's available there now. I'll hopefully get around to it soon if no one beats me to it.

bergmark:
There are some dependencies needed before we can get (mainly the test-suite) building on 8.8: dlist, integer-logarithms, attoparsec, optparse-applicative (and maybe more that I missed).

asr:
Blocking https://github.com/agda/agda/issues/3725.

phadej:
I think the code is adjusted, only bounds need to be relaxed. So `--allow-newer=aeson` should "just work".

phadej:
The `aeson` library dependencies are updated, except `dlist`. https://github.com/spl/dlist/issues/34

tests/benchmarks need updates (at least `microstache`, which I'll update once `dlist` is updated).

AndreasPK:
dlist has been updated.

RyanGlScott:
It appears that `aeson-1.4.4.0` was [revised](http://hackage.haskell.org/package/aeson-1.4.4.0/revisions/) to permit `template-haskell-2.15.0.0` and `time-1.9`. In light of this, is this PR still necessary, or can it be closed?

phadej:
Ryan: you are right. There are some “feature full” commits after 1.4.4.0 tag, but aeson compiles with GHC-8.8 without them as well.

Adam: thanks for the revision.

Let me see what I can release/review myself now...

bergmark:
Thanks.

PR #698, opened 2019-05-12
Use time-compat and fix MonadFail breakages

phadej:
This is demo of how new `time-compat` (@bergmark check your mail!) could easy the `time` story.

I also added MonadFail fixes; that will be required for GHC-8.8; I however didn't relax any bounds yet.

This PR depends on unreleased versions of `time-compat` and `quickcheck-instances`, so CI will fail. (obviously, also didn't update `stack.yaml`s)


bergmark:
LGTM!

phadej:
HLint job fails, but #699 should fix it.

Otherwise this is ready.

PR #701, opened 2019-05-15
Require recent bytestring with ghc-8.0 or later

phadej:
This is somewhat band-aid solution for:

Data/Aeson/Encoding/Builder.hs:87:21: error:
    • No instance for (Semigroup Builder) arising from a use of ‘<>’

But there aren't better solution: we don't include anything
from semigroups on ghc >= 8.0; so semigroups couldn't even
in theory provide that instance.

Also a lot of older aeson releases are "broken" because of this,
if one downgrades bytestring. Usually no-one does that, except weird
automatic QA tools.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,2 +1,3 @@
 with-compiler: ghc
-packages: ., attoparsec-iso8601, benchmarks
+packages: .
+packages: attoparsec-iso8601, benchmarks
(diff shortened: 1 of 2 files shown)


cabal.project
```diff
@@ -1,2 +1,3 @@
 with-compiler: ghc
-packages: ., attoparsec-iso8601, benchmarks
+packages: .
+packages: attoparsec-iso8601, benchmarks
```
phadej: This is to make it easier to comment out locally

bergmark:
Thanks!

PR #703, opened 2019-05-18
Support primitive 0.7.* (fixes #702)

### LINKED ISSUE #702: support primitive >= 0.7
the unlifted array instances no longer exist :) 

i think older releases still build fine since they dont depend on it, but the most recent two do

bergmark:


andrewthad:
This looks good to me.

PR #705, opened 2019-05-30
Allow semigroups-0.19 and hashable-1.3

phadej:
Finally there is an install plan with these:

```
% cabal new-test all --constraint='semigroups>=0.19' --constraint='hashable>=1.3' --enable-tests
```

builds and tests pass.

I also removed bounds from `aeson-benchmarks` as that package is not intended to be released.


bergmark:
thanks!

Issue #690, opened 2019-03-25
omitNothingFields does not work with Maybe from type parameter


htmue:
Maybe-fields are always required if the field's type is introduced by type parameter. Example:
```haskell
> data G a = G { a :: a, b :: Maybe Int } deriving (Show, Generic)
> instance FromJSON a => FromJSON (G a) where parseJSON = genericParseJSON defaultOptions {omitNothingFields=True}
> eitherDecode "{}" :: Either String (G (Maybe Int))
Left "Error in $: key \"a\" not present"
```
When the key is missing and `omitNothingFields = True`, parsing should succeed, as it does for `b`:
```haskell
> eitherDecode "{\"a\": null}" :: Either String (G (Maybe Int))
Right (G {a = Nothing, b = Nothing})
```

Lysxia:
One way is to specialize the variable `a` in the instance declaration:

    instance FromJSON b => FromJSON (G (Maybe b)) where
      parseJSON = genericParseJSON defaultOptions {omitNothingFields=True}

You'll have to deal with overlapping instances if you also want other instantiations of `G`.

> When the key is missing and `omitNothingFields = True`, parsing should succeed

There is a fundamental limitation in the way type classes work that makes this infeasible. The implementation of the `omitNothingFields` option is quite fragile (and my above suggestion abuses that). Properly supporting this use case would require a significant redesign of aeson's generic machinery.

htmue:
That's a clear answer, thanks a lot!

Your suggestion works well for my toy example, but in the complex context where I actually need it, I get link errors (symbol missing). I guess that's due to the fragility you mention.

I found a simple solution though - inserting `"a": null` if the optional key is missing.
```haskell
instance FromJSON a => FromJSON (G a) where
  parseJSON =
    genericParseJSON defaultOptions {omitNothingFields = True} . fixOptionalKey "a"

fixOptionalKey :: Text -> Value -> Value
fixOptionalKey key (Object o) = Object $ H.insertWith (flip const) key Null o
fixOptionalKey _ v = v
```

Lysxia:
Link errors sound like an unrelated problem. But I'm glad you found a solution!

htmue:
Me too.

However, I think it would be good to document this behaviour, it's unexpected and needs a workaround.

[...]

PR #708, opened 2019-06-13
Document surprising behavior for omitNothingFields

### LINKED ISSUE #707: Record field must be present when `Maybe a` introduced via type parameter
Introducing a `Maybe` via a type parameter allows the value to be `Null` but the field is required to be present in order for parsing to succeed.

```
data Example a = Example { field :: a } deriving (Show, Generic)
instance FromJSON a => FromJSON (Example a)

λ> eitherDecode "{}" :: Either String (Example (Maybe ()))
Left "Error in $: key \"field\" not present"
```

This behavior is contrary to how `Maybe` works when declared inline in the type declaration.

```
data Example2 a = Example2 { field :: Maybe a } deriving (Show, Generic)
instance FromJSON a => FromJSON (Example2 a)

λ> eitherDecode "{}" :: Either String (Example2 ())
Right (Example2 {field = Nothing})
```

This distinction is subtle and unintuitive and I couldn't find it documented anywhere.  The obvious workaround is to avoid using `Maybe` via type parameter, but hopefully this is an accidental omission rather than a limitation in the generic deriving machinery :)

Lysxia:
Closes #707, #690.

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -568,6 +568,39 @@ data Options = Options
       -- ^ If 'True' record fields with a 'Nothing' value will be
       -- omitted from the resulting object. If 'False' the resulting
       -- object will include those fields mapping to @null@.
+      --
+      -- === Note
+      --
+      -- Setting 'omitNothingFields' to 'True' only affects fields which are of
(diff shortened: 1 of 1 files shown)


bergmark:
Oh i missed this one, LG thanks!

Issue #463, opened 2016-08-27
Mismatch between derived To/FromJSONKey and To/FromJSON instances for "strings"


bergmark:
I think this was an oversight and we need generics for To/FromJSONKey to make those instances match To/FromJSON.

The reason this happens is that deriving ToJSON for some types such as "enumerables" produces a string, but the ToJSONKey instance's default is always to produce an Array of arrays instead of an object, even when the key serializes to a string.

```haskell
#!/usr/bin/env stack
-- stack --resolver nightly-2016-08-27 --install-ghc runghc --package aeson-1.0.0.0 --package semigroups-0.18.2 --package unordered-containers --package hashable
{-# LANGUAGE DeriveGeneric #-}

import Data.Aeson
import Data.Hashable
import GHC.Generics
import qualified Data.ByteString.Lazy.Char8 as L
import qualified Data.HashMap.Strict as H

data T = A | B deriving (Eq, Generic)
instance Hashable T
instance ToJSON T
instance ToJSONKey T

main :: IO ()
main = do
  L.putStrLn $ encode A                       -- => "A"
  L.putStrLn $ encode (H.fromList [(A, 'a')]) -- => [["A","a"]]
```

phadej:
We could add
`gToTextJsonKey` and friends, so the change can be done in backwards compatible way, i.e. you'll write
```hs
instance Hashable T
instance ToJSON T
instance ToJSONKey T
  toJSONKey = gToTextJsonKey
```

Also, my gut feeling it would be much simpler to do this way, than try to conditionally do either one encoding, based on types shape...

bergmark:
That sounds like a good first step!

PR #711, opened 2019-06-15
Generic implementations for FromJSONKey and ToJSONKey

### LINKED ISSUE #710: Generic ToJSONKey and FromJSONKey
There seems to be no easy way to write common instances of `ToJSONKey`  and `FromJSONKey`. In the following answer I use `Show` and `Read` but that still takes some typing and knowledge about the library.

https://stackoverflow.com/questions/56609201/how-to-use-a-sum-type-as-a-key-in-map-when-using-aeson/56610450#56610450

We should be able to derive this with `Generic`, defining a `gToJSONKeyText` and `gFromJSONKeyTextParser`. Related to issue #463 which suggests this solution as an intermediate step for a harder problem.

Lysxia:
Fixes #710

### DIFF
--- a/Data/Aeson/Types/Class.hs
+++ b/Data/Aeson/Types/Class.hs
@@ -58,6 +58,11 @@ module Data.Aeson.Types.Class
     , fromJSONKeyCoerce
     , coerceFromJSONKeyFunction
     , mapFromJSONKeyFunction
+    -- ** Generic keys
+    , GToJSONKey()
+    , genericToJSONKey
+    , GFromJSONKey()
+    , genericFromJSONKey
     -- * Object key-value pairs
     , KeyValue(..)
 
--- a/Data/Aeson/Types/ToJSON.hs
+++ b/Data/Aeson/Types/ToJSON.hs
@@ -47,6 +47,10 @@ module Data.Aeson.Types.ToJSON
     , ToJSONKeyFunction(..)
     , toJSONKeyText
     , contramapToJSONKeyFunction
+
+    , GToJSONKey()
+    , genericToJSONKey
+
     -- * Object key-value pairs
     , KeyValue(..)
     , KeyValuePair(..)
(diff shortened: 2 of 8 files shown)


Lysxia:
I don't understand the CI failure. Seems unrelated.

bergmark:
That's just the hlint step, so don't worry about it.

bergmark:
Thanks! Looks like this also resolves the first step of #463?

Might be nice to have some full fledged example in the test suite as well, but no blocker for me to merge this.

I'll look into the hlint issue before merging this, to see if it's just this PR causing it or not.

Lysxia:
> Thanks! Looks like this also resolves the first step of #463?

Yes it does!

Data/Aeson/Types/ToJSON.hs
```diff
@@ -515,6 +529,28 @@ contramapToJSONKeyFunction h x = case x of
     ToJSONKeyText  f g -> ToJSONKeyText (f . h) (g . h)
     ToJSONKeyValue f g -> ToJSONKeyValue (f . h) (g . h)
 
+-- 'toJSONKey' for 'Generic' types.
+-- These types must be sums of nullary constructors, whose names will be used
```
phadej: Deriving is supported for enumeration types, i.e. the sums of nullary constructors. The names of constructors will be used as keys for JSON objects.

Data/Aeson/Types/Class.hs
```diff
@@ -58,6 +58,11 @@ module Data.Aeson.Types.Class
     , fromJSONKeyCoerce
     , coerceFromJSONKeyFunction
     , mapFromJSONKeyFunction
+    -- ** Generic keys
+    , GToJSONKey()
+    , gToJSONKey
```
phadej: `genericToJSONKey` to to be consistent with `genericToJSON` etc.

Lysxia:
thanks @phadej I incorporated your suggestions.

Lysxia:
Good idea, I'll do that in the next few days.

Lysxia:
For `Nullary` which has only three values, well, we can easily enumerate them, so a simple unit test gives full input coverage (as in this last updated commit).

phadej:
@Lysxia except when you add another case, etc. E.g. when Generics mechanism is expanded, there would be simple property to copy...

Lysxia:
We'll see if and when someone adds such functionality. Until then (which might well be forever), a roundtrip test is redundant with the existing unit tests.

bergmark:
Would probably be nice to have that test as well but not a blocker, I think. I didn't get time to fix travis+hlint before going away over the weekend so I'll merge this now.

PR #714, opened 2019-06-22
Add handling of duplicate keys

### LINKED ISSUE #531: Duplicate keys
The behavior of parsing duplicate keys has changed from last key wins to first key wins. I do not see any mention of this change in the changelog -- was it intentional? Is there a way to force the previous behavior?

E.g. With aeson-0.11.2.1:
```shell
$ stack --resolver lts-7.7 ghci --package aeson
:set -XOverloadedStrings
import Data.Aeson
decode "{\"a\": false, \"a\": true}" :: Maybe Object
# Just (fromList [("a",Bool True)])
```

Version 1.0.2.1 results in `Just (fromList [("a",Bool False)])`.

Lysxia:
Fixes #531.

This adds a parameterized parser `jsonWith` to let users decide what to do with objects with duplicate keys. A few common variants are also included.

Benchmarks to guard against regression (using the aeson-parse benchmark) (no difference on GHC 8.6, 8.4, there seems to be a 2% slowdown on 8.2.2) https://gist.github.com/Lysxia/beb1aa201c79f576b94c509af7f0abfa

### DIFF
--- a/Data/Aeson/Parser/Internal.hs
+++ b/Data/Aeson/Parser/Internal.hs
@@ -5,6 +5,10 @@
 #if MIN_VERSION_ghc_prim(0,3,1)
 {-# LANGUAGE MagicHash #-}
 #endif
+#if __GLASGOW_HASKELL__ <= 710 && __GLASGOW_HASKELL__ >= 706
+-- Work around a compiler bug
+{-# OPTIONS_GHC -fsimpl-tick-factor=200 #-}
+#endif
 -- |
 -- Module:      Data.Aeson.Parser.Internal
 -- Copyright:   (c) 2011-2016 Bryan O'Sullivan
--- a/Data/Aeson/Parser/Internal.hs
+++ b/Data/Aeson/Parser/Internal.hs
@@ -21,30 +21,43 @@ module Data.Aeson.Parser.Internal
     (
     -- * Lazy parsers
       json, jsonEOF
+    , jsonWith
+    , jsonLast
+    , jsonAccum
+    , jsonNoDup
     , value
     , jstring
     , jstring_
     , scientific
     -- * Strict parsers
     , json', jsonEOF'
+    , jsonWith'
+    , jsonLast'
+    , jsonAccum'
+    , jsonNoDup'
     , value'
     -- * Helpers
     , decodeWith
     , decodeStrictWith
     , eitherDecodeWith
     , eitherDecodeStrictWith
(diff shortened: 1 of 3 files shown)


Lysxia:
Woohoo, a compiler bug on 7.6.3 - 7.10.3 https://travis-ci.org/bos/aeson/jobs/549001593

bergmark:
Cool, thanks!

Data/Aeson/Parser/Internal.hs
```diff
@@ -98,23 +120,38 @@ json = value
...
   jstring' = do
     !s <- jstring
     return s
+{-# INLINE object_' #-}
 
-objectValues :: Parser Text -> Parser Value -> Parser (H.HashMap Text Value)
-objectValues str val = do
+-- The object parser 'objectValues' is parameterized by the object constructor.
+-- See also 'jsonWith'.
+type HFromList = [(Text, Value)] -> Parser Object
```
phadej: I'd have `Either String Object` as a return type. `Parser` is "too big".

And actually wouldn't use a type synonym, `hFromList` is confusing naming, `mkObject` is what it is

Data/Aeson/Parser/Internal.hs
```diff
@@ -157,42 +196,104 @@ arrayValues val = do
...
+-- associated values from the original list @kvs@.
+--
+-- >>> fromListAccum [("apple", Bool True), ("apple", Bool False), ("orange", Bool False)]
+-- fromList [("apple", [Bool False, Bool True]), ("orange", [Bool False])]
+fromListAccum :: [(Text, Value)] -> Object
+fromListAccum =
+  fmap (Array . Vector.fromList . ($ [])) . H.fromListWith (.) . (fmap . fmap) (:)
+
+-- | @'fromListNoDup' kvs@ fails if @kvs@ contains duplicate keys.
+parseListNoDup :: [(Text, Value)] -> Parser Object
```
phadej: point in case: here the `HFromList` is not used, so connecting the dots is difficult.

I'd remove the type alias.

Data/Aeson/Parser/Internal.hs
Lysxia: I see. I forgot we're in attoparsec's `Parser`, not aeson's `Parser` (which is closer to `Either`). That's a fair point about the synonym, I'll remove it. Thanks for the review!

PR #715, opened 2019-06-22
Fix dependencies of the benchmarks package for old GHCs

Lysxia:
Apparently Travis doesn't build these benchmarks (in spite of a comment indicating the contrary in [`travis/script.sh`](https://github.com/bos/aeson/blob/a09a0bcb6e04ed18ca6207ea855349a9ade2a1af/travis/script.sh#L12)).


bergmark:
Thanks for the fix! IIRC not building the benchmark for every GHC is on purpose to speed up build times.

PR #717, opened 2019-06-23
Change the type of jsonWith so its first argument is in Either instead of Parser

Lysxia:
Follow up from @phadej's review in #714 

Updated benchmarks (pasted next to each other, no difference with pre-#714 aeson, for 8.2-8.6) https://gist.github.com/Lysxia/a8c1825f6a946f2196677082f02582e1


phadej:
Looks great!

bergmark:
LG, thanks @Lysxia & @phadej!

PR #727, opened 2019-09-06
Allow building with base-compat-0.11.0

RyanGlScott:
`base-compat-0.11.0` makes `Prelude.fail` refer to `Control.Monad.Fail.fail`, not `Control.Monad.fail`, to mirror the API that GHC 8.8's version of `base` offers. One consequence of this change is that `aeson-1.4.4.0` does not compile out of the box with `base-compat-0.11.0`. Luckily, this can be rectified in a way that is backwards-compatible with earlier versions of `base-compat` by making more use of qualified imports, which this patch accomplishes.


bergmark:
Thanks!

PR #729, opened 2019-09-13
Add formatPath function

robbiemcmichael:
Exposes a function to format a `JSONPath` as a `String` so that it can be used for more than formatting error messages.

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -49,6 +49,8 @@ module Data.Aeson.Types.Internal
     , parserThrowError
     , parserCatchError
     , formatError
+    , formatPath
+    , formatRelativePath
     , (<?>)
     -- * Constructors and accessors
     , object
@@ -458,7 +460,17 @@ parseEither m v = runParser (m v) [] onError Right
 -- | Annotate an error message with a
 -- <http://goessner.net/articles/JsonPath/ JSONPath> error location.
 formatError :: JSONPath -> String -> String
-formatError path msg = "Error in " ++ format "$" path ++ ": " ++ msg
+formatError path msg = "Error in " ++ formatPath path ++ ": " ++ msg
+
+-- | Format a <http://goessner.net/articles/JsonPath/ JSONPath> as a 'String',
+-- representing the root object as @$@.
+formatPath :: JSONPath -> String
+formatPath path = "$" ++ formatRelativePath path
+
+-- | Format a <http://goessner.net/articles/JsonPath/ JSONPath> as a 'String'
+-- which represents the path relative to some root object.
+formatRelativePath :: JSONPath -> String
+formatRelativePath path = format "" path
   where
     format :: String -> JSONPath -> String
     format pfx []                = pfx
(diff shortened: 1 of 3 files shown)


Data/Aeson/Types/Internal.hs
```diff
@@ -458,7 +459,12 @@ parseEither m v = runParser (m v) [] onError Right
 -- | Annotate an error message with a
 -- <http://goessner.net/articles/JsonPath/ JSONPath> error location.
 formatError :: JSONPath -> String -> String
-formatError path msg = "Error in " ++ format "$" path ++ ": " ++ msg
+formatError path msg = "Error in " ++ formatPath "$" path ++ ": " ++ msg
+
+-- | Format a <http://goessner.net/articles/JsonPath/ JSONPath> as a 'String'.
+-- The @root@ parameter is typically set to @$@ when there is no root object.
+formatPath :: String -> JSONPath -> String
```
bergmark: What about dropping `root` and always setting it to `$`?

robbiemcmichael: I thought it would be nice to also give users a way to construct:
- A path referencing a named root object (since `$` is just for an anonymous root object)
- A relative path which could be appended to another path (achieved by setting the root to an empty string)

Some other implementation options:
- Two separate functions, `formatPath` for an absolute path (i.e. include `$` prefix) and `formatPathRelative` for a relative path (i.e. empty string for root, users free to prefix a root in their own code)
- Just make users drop the `$` prefix themselves if they want a relative path

That's the context behind my original choice, but I'm happy to go with whichever approach you prefer.

bergmark: Thanks! Supporting all of this sounds like a good idea.

I think I'd prefer having `formatPath` and `formatPathRelative` (or perhaps `formatRelativePath`?). At least from my use cases I've always wanted the `$` so it would be a bit more cumbersome to have to pass `"$"` as the first argument

bergmark: Hmm or is there a better name than`Relative`? I guess it's not accurate if you want to use it for a named root object?

robbiemcmichael: I think `Relative` is the right term as the JSON path is relative to some object, which may be the root object or may be an object referenced by another JSON path. I'll go with your suggestion of `formatRelativePath`.

robbiemcmichael:
@bergmark I've made the changes. The build seems to be failing with some versions of GHC but I don't think it's related to my changes.

bergmark:
Thanks! I'll try to look into the build failures soon.

bergmark:
🎉

robbiemcmichael:
@bergmark Any chance of a Hackage release in the near future? I was hoping to make use `formatPath` to improve some of the error messages in the `json-to-dhall` and `yaml-to-dhall` programs from [dhall-haskell](https://github.com/robbiemcmichael/dhall-haskell).

bergmark:
I'll do it right now, thanks for the reminder!

bergmark:
Done: http://hackage.haskell.org/package/aeson-1.4.6.0

robbiemcmichael:
That was quick! Thanks 😄

Issue #719, opened 2019-07-20
"key tag not present" error message is somewhat confusing


nineonine:
I saw this error message and it confused me - I started looking for an actual "tag" key in my JSON... 

After some research I realized tag is something related to Aeson internal machinery (?)

Perhaps we could improve this message?

I am willing to submit a patch, but we need a better message first ...

I wonder whether that happened to anyone else ... Maybe it is ok after all.

Lysxia:
This error message could definitely be improved.

> After some research I realized tag is something related to Aeson internal machinery (?)

I'm a bit puzzled by this. The string `" not present"` only occurs in `explicitParseField` so, for that error to occur, a user of that library, maybe another dependency you're using, must have used that function with a `Text` argument `"key"` which is then expected to be in a JSON object. That's not an internal detail.

Is there more context about how you came across this error?

nineonine:
That happened when I tried to write a FromJSON instance for one of my types. The JSON for this type is special, it has 1 field which is another stringified JSON. I think I used `withEmbeddedJSON` (although I am not sure the problem was here)
I guess by internal' I meant "something related to Aeson parsing routine".
Unfortunately, I still don't really understand what this error message is trying to tell me so I am unable to suggest something better or more meaningful.

Lysxia:
Ah I have an idea where that might come from. `"tag"` is the default key used for the generic implementation of `FromJSON`. So given an ADT with multiple constructors with a derived `FromJSON`,

```
data T = Foo Int | Bar String deriving (Generic, FromJSON)
```

the corresponding JSON must look like:

```
{ "tag": "Foo", "contents": 3 }  // Foo 3
{ "tag": "Bar", "contents": "baz" }  // Bar "baz"
```

If your JSON input doesn't look like that, and in particular, if it is an object without such a `tag` field, you get the error you mentioned. Now I can imagine how this can occur accidentally, so that this error feels like it comes out of nowhere.

nineonine:
Aha, that makes sense now!
I think we should either attempt to come up with better error or explain it somewhere in documentation.

bergmark:
> I think we should either attempt to come up with better error message or explain it somewhere in documentation.

Either or both of those sound good to me, but I'm not sure how the error message can be improved, any ideas? Would `missing required field "tag"` instead of `key "tag" not present` help?

nineonine:
[...]

PR #730, opened 2019-09-18
Provide a clearer message when the required tagKey is missing, #719

gdevanla:
Please let me know if this is the direction we want to proceed in. I can add tests for this before the merge. Also, this is my first PR to this repo, so any tips on the process is greatly appreciated.


Lysxia:
I think the problem is to explain *why* the key `"tag"` is needed. One way to do so may be to explain more clearly the expected format:

```
Error in $: parsing T failed, expected an object with keys "tag" and "contents", where "tag" is associated to one of ["Foo", "Bar"], could not find key "tag"
```

(I think "could not find key X" seems a tiny bit clearer than "key X is not present")

Lysxia:
Also you're doing just fine with the process!

gdevanla:
Thanks for the comment. I believe you cannot get the 'cnames' you need if you cannot get to the 'tag'. Is there another way to get the cnames which I am not seeing?

gdevanla:
@Lysxia I see that the message you suggested would be useful. But, I am not able to see how we could get the `cname` at that point. Without, calling into a function like `parseFromTaggedObject` seems like I cannot force the type of `f a`.  Can you provide some direction on this.  I am still trying to wrap my head around the dispatching going on around the typeclasses.

Lysxia:
You can define `cnames` as `cnames = unTagged2 (constructorTags opts :: Tagged2 f [String])`, provided there is a  `ConstructorNames f` instance (which there is here).

That constraint is already implicitly used through `failWith_` (which is defined using `failWithCTags`, which defines `cnames` as above), but `failWith_` is only meant to be called as the last action of a `do`-block (in order to make use of type information and avoid an explicit type annotation).

Here you will need the constructor names to provide context to `obj .: tagKey`, where there is no such information available, so you will need a type annotated definition of `cnames`.

gdevanla:
@Lysxia Thanks for the comment. That was my first attempt. But, I think the (ConstructorNames f) is ambiguous between a :+: b and (C1 c a) instances of ConstructorNames.

Do you think might need another type-class to handle both scenarios?

Lysxia:
I guess the problem you ran into is that `f` is not scoped in the body of `parseNonAllNullarySum`. You can add `forall f c arity.` to its type signature.

```
parseNonAllNullarySum :: forall f c arity. ...
parseNonAllNullarySum ... =
   ...
   where
     cnames = unTagged2 (constructorTags opts :: Tagged2 f [String])
```

gdevanla:
Great. Thank you for that tip. I totally overlooked that `f` was not visible in the body of the function with scoping the it :(

gdevanla:
@Lysxia That was helpful. Until now I always thought `forall` was only for quantitying Existentials and did not know also affected scoping!

gdevanla:
@Lysxia Please could you review my changes and let me know if this PR is good to go.

Data/Aeson/Types/FromJSON.hs
```diff
@@ -951,6 +952,19 @@ type ConName = String
 contextType :: TypeName -> Parser a -> Parser a
 contextType = prependContext
 
+
+
```
Lysxia: Can you trim these extra two lines (and the other addition above)?

tests/golden/generic.expected
```diff
@@ -10,7 +10,7 @@ Error in $: parsing Types.SomeType(Unary) failed, key "contents" not present
 Error in $: parsing Types.SomeType(Record) failed, key "testone" not present
 Error in $.testone: parsing Double failed, unexpected Boolean
 Error in $.tag: parsing Types.SomeType failed, expected tag field to be one of ["nullary","unary","product","record","list"], but found tag "X"
-Error in $: parsing Types.SomeType failed, key "tag" not present
+Error in $: parsing Types.SomeType failed, expected an object with key "tag" where "tag" is associated to one of ["nullary","unary","product","record","list"]. The parser returned error was: key "tag" not present
```
Lysxia: - "expected Object" instead of "expected an object" would make the error look more uniform with the rest.
- "with key "tag" associated to" instead of "with key "tag" where "tag" is associated to"
- you can also remove "the parser returned error was"

Data/Aeson/Types/FromJSON.hs
```diff
@@ -1140,14 +1154,17 @@ parseNonAllNullarySum p@(tname :* opts :* _) =
...
-                  parseFromTaggedObject (tag :* contentsFieldName :* p) obj
+                parseFromTaggedObject (tag :* contentsFieldName :* p) obj
         where
           tagKey = pack tagFieldName
           badTag tag = failWith_ $ \cnames ->
               "expected tag field to be one of " ++ show cnames ++
               ", but found tag " ++ show tag
+          cnames_ = unTagged2 (constructorTags (constructorTagModifier opts) :: Tagged2 f [String])
+          --cname_ = conName (undefined :: M1 _i c _f _p)
+          -- forall f a. ConstructorNames f
```
Lysxia: You can remove these comments and the extra indentation change line 1159.

tests/golden/generic.expected
Lysxia: and I think "not found" sounds better than "not present"

Data/Aeson/Types/FromJSON.hs
```diff
@@ -952,18 +952,15 @@ type ConName = String
...
 -- | "contents", where "tag" i-- |s associated to one of ["Foo", "Bar"],
 -- | The parser returned error was: could not find key "tag"
 contextTag :: Text -> [String] -> Parser a -> Parser a
-contextTag tagKey cnames = prependFailure
-  ("expected an object with key \"" ++ unpack tagKey ++
-    "\" where \"tag\" is associated to one of " ++ show cnames ++
-    ". The parser returned error was: ")
+contextTag tagKey cnames = modifyFailure (\_ ->
+  ("expected Object with key \"" ++ unpack tagKey ++ "\"" ++
+    " associated to one of " ++ show cnames ++ "."))
```
Lysxia: `prependFailure` would still be good here, only with a single comma after this new prefix.

gdevanla: `prependFailure` would add the `key 'tag' not present' phrase here, which may be redundant, I think. That is the reason I removed it.

Lysxia: I think error messages should be explicit about both what was expected and what was not expected, rather than let users guess the source of the mismatch, unless it's really obvious, which doesn't seem to be the case here. In particular, another potential source of error at this location would be if the `"key"` tag exists, but is associated to some non-string value.

What I was complaining about earlier was specifically the extra string `"the parser returned error was"`, because it does not seem like a user can easily understand what "the parser" refers to; it could refer to the "parser" `obj .: tagkey`, which reveals too much internal details that users may not be aware of, or it could refer to the toplevel function they are calling (`decode`, `parseJSON`...), which does make that part of the message redundant.

gdevanla: Sure. I understand the we had to provide the original information to the user. And that is the reason I had added used 'prependFailure' earlier.  I can remove the 'the parser error was' as you suggested. But, in this context I found the last part of message 'key 'tag' not present' to be repetitive. If you think otherwise,  I can leave it there.

Data/Aeson/Types/FromJSON.hs
```diff
@@ -1140,14 +1151,15 @@ parseNonAllNullarySum p@(tname :* opts :* _) =
     case sumEncoding opts of
       TaggedObject{..} ->
           withObject tname $ \obj -> do
-              tag <- contextType tname $ obj .: tagKey
+              tag <- contextType tname . contextTag tagKey cnames_ $ obj .: tagKey
               fromMaybe (badTag tag <?> Key tagKey) $
-                  parseFromTaggedObject (tag :* contentsFieldName :* p) obj
+                parseFromTaggedObject (tag :* contentsFieldName :* p) obj
```
Lysxia: This edit can be avoided.

Data/Aeson/Types/FromJSON.hs
Lysxia: `obj .: tagKey` does two things

- it looks up `tagKey` in `obj`, which fails if the key is not there, that's the first possible source of error, and the cause of the issue this PR is meant to fix.

- and it parses the associated value according to the `FromJSON` instance of the expected result's type, which in this case is `Text` (or `String` I'm not sure). This parsing can also fail, that's the second source of error that can happen at this point of the program.

gdevanla: @Lysxia Thanks for the info. Here is now the message might look with `prependFailure` included.

```
>data T = Foo Int | Bar String deriving (Generic, FromJSON, ToJSON, Show)
>eitherDecode "{\"tag1\": \"Foo1\", \"contents\": 10}" ::  Either String T
Left "Error in $: parsing Ghci7.T failed, expected Object with key \"tag\" associated to one of [\"Foo\",\"Bar\"].key \"tag\" not present"
>eitherDecode "{\"tag\": 10, \"contents\": 10}" ::  Either String T
Left "Error in $.tag: parsing Ghci7.T failed, expected Object with key \"tag\" associated to one of [\"Foo\",\"Bar\"].parsing Text failed, expected String, but encountered Number"
```

You see that the second part of message in both scenarios are not Capitalized. Should we do that? Or is is common across all error message in aeson? Also, to my point the second part of the first error message 'key \"tag\" not present" is kind of redundant, and hence my previous change.

I think the current solution is to leave 'prependFailure' intact and let the redundant message be as shown above.

Data/Aeson/Types/FromJSON.hs
gdevanla: Curious, what are the indenting rules around this. I was using the default Haskell mode indentation in Emacs.

bergmark: We use 4 space indentation (See CONTRIBUTING.md), but we are definitely not consistent about it.

Data/Aeson/Types/FromJSON.hs
gdevanla: Also, we have other places where we use 'not present'. Changing 'not present' to 'not found' might preclude changing those places as well.

Data/Aeson/Types/FromJSON.hs
gdevanla: Thanks @bergmark. I just noticed that in the CONTRIBUTING.md.  I will rollback my edit.

Data/Aeson/Types/FromJSON.hs
Lysxia: The "convention" is to prepend using a comma as a separator. It's just what I came up with when I refactored the error handling in this module, so don't hesitate to discuss it if you have better ideas!

I don't think the second part of the error message is redundant. If you say `expected Object with key \"tag\" associated to one of [\"Foo\",\"Bar\"].`, then there are two ways to break that expectation:

- the key `"tag"` doesn't exist
- or the key tag is not associated to one of these strings: it could be a different string (that is handled later on in the `do` block), or not a string at all.

So that's an extra bit of information you can give directly, rather than expect users from deducing it themselves in an error prone way.

---

I think "not found" is better than "not present" in all the other places too.

Lysxia:
Looks good to me. :)

bergmark:
Thanks to both of you!

Looks really good. I was actually stumbling on explaining the old message to some new haskellers the other day.

I have one comment about `expected Object with key "tag" associated to one of ["nullary","unary","product","record","list"],`

What does `associated` mean here? That the `tag` key should contain one of the specified values?

If so, what do you think about saying `expected Object with key "tag" containing one of the values ["nullary","unary","product","record","list"]` instead? I think that would be clearer to me, but what do you think?

Lysxia:
> If so, what do you think about saying expected Object with key "tag" containing one of the values ["nullary","unary","product","record","list"] instead? I think that would be clearer to me, but what do you think?

That also sounds better to me too.

bergmark:
🎉

PR #743, opened 2019-11-30
Drop pre-GHC-7.8 support

phadej:
This commit makes only non-breaking cleanups. Stricter lower-bounds for

- `__GLASGOW_HASKELL__`
- base
- bytestring
- template-haskell

allow to remove some CPP


phadej:
Green

bergmark:
Thanks!

PR #745, opened 2020-01-03
Allow FromJSON instance for Ratio a parse numbers

Shimuuar:
This PR modify `FromJSON` instance of `Ratio a` to accept JSON numbers. `ToJSON` is unchanged so this PR only increases number of valid inputs for `Ratio` data type.

I think that for any numeric type it should be possible to parse it from JSON numbers. I have been bitten by fact that `say `1.4` wouldn't parse as `Rational` for two times already.

### DIFF
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -1541,10 +1541,15 @@ instance FromJSONKey Float where
         _           -> Scientific.toRealFloat <$> parseScientificText t
 
 instance (FromJSON a, Integral a) => FromJSON (Ratio a) where
-    parseJSON (Number x) = prependContext "Ratio"
-                         $ withBoundedScientific' (pure . realToFrac)
-                         $ Number x
-    parseJSON o          = objParser o
+    parseJSON (Number x)
+      | exp10 <= 1024
+      , exp10 >= -1024 = return $! realToFrac x
+      | otherwise      = prependContext "Ratio" $ fail msg
+      where
+        exp10 = base10Exponent x
+        msg = "found a number with exponent " ++ show exp10
+           ++ ", but it must not be greater than 1024 or less than -1024"
+    parseJSON o = objParser o
       where
         objParser = withObject "Rational" $ \obj -> do
             numerator <- obj .: "numerator"
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -1541,12 +1541,15 @@ instance FromJSONKey Float where
         _           -> Scientific.toRealFloat <$> parseScientificText t
 
 instance (FromJSON a, Integral a) => FromJSON (Ratio a) where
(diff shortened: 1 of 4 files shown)


Data/Aeson/Types/FromJSON.hs
```diff
@@ -1541,12 +1541,15 @@ instance FromJSONKey Float where
         _           -> Scientific.toRealFloat <$> parseScientificText t
 
 instance (FromJSON a, Integral a) => FromJSON (Ratio a) where
-    parseJSON = withObject "Rational" $ \obj -> do
-        numerator <- obj .: "numerator"
-        denominator <- obj .: "denominator"
-        if denominator == 0
-        then fail "Ratio denominator was 0"
-        else pure $ numerator % denominator
+    parseJSON (Number x) = pure (realToFrac x)
```
phadej: This is security risk, think `1e9999999`, see how `Integer` is parsed, e.g.

Shimuuar: Sorry for the delay. I sort of forgot about this PR. I updated it with check analogous to one for Integer 

I also noticed that there's unused import

Data/Aeson/Types/FromJSON.hs
```diff
@@ -1541,7 +1541,9 @@ instance FromJSONKey Float where
         _           -> Scientific.toRealFloat <$> parseScientificText t
 
 instance (FromJSON a, Integral a) => FromJSON (Ratio a) where
-    parseJSON (Number x) = pure (realToFrac x)
+    parseJSON (Number x) = prependContext "Ratio"
+                         $ withBoundedScientific' (pure . realToFrac)
```
phadej:
```
withBoundedScientific_ :: (Parser a -> Parser a) -> (Scientific -> Parser a) -> Value -> Parser a
withBoundedScientific_ whenFail f (Number scientific) =
    if exp10 > 1024
    then whenFail (fail msg)
    else f scientific
```

tests only for overflows, not undeerflows. I.e. `1e-2000` won't fail (but it should).

Shimuuar: You're right. Fixed this as well. Thank for pointing that stuff out

bergmark:
Sorry for the delay! probably good to go, but i’m not at a computer right
now. And big thanks!

Shimuuar:
No problem! Not that it's urgent. I only encounter problem with this instance once a year or so. But it always result in debugging and WAT?!

changelog.md
```diff
@@ -1,5 +1,10 @@
 For the latest version of this document, please see [https://github.com/bos/aeson/blob/master/changelog.md](https://github.com/bos/aeson/blob/master/changelog.md).
 
+### upcoming
+
+* `FromJSON` instance of `Ratio a` now parses numbers in addtion to
```
phadej: addition

bergmark: I'll fix this later, hopefully :-P

PR #749, opened 2020-01-31
Add option to reject unknown fields (#207)

### LINKED ISSUE #207: Add generic encoding option to reject unknown field-names
I've started converting my previously hand-written instances in terms of `genericParseJSON` and `genericToJSON`, but one thing I'm missing though is to
have `genericParseJSON` (optionally) result in a parser error, if an unexpected/unknown field occurs in the JSON object. 

For me, this is quite useful in (some) cases (especially when `Maybe`-fields are involved) when implementing JSON services to notice clients which have sent data with mistyped field-names, as well as on the client-side, to notice new/unhandled field-labels returned by a service (or when I mistyped the field-name in the client-code).

/cc @basvandijk

rmanne:
This implements #207, and supports both TH and Generics. The Generics implementation is not particularly efficient, but as the cost of traversing the datatype twice is strictly compile-time, this should be acceptable.

### DIFF
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -11,6 +11,7 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE ViewPatterns #-}
 
@@ -1200,9 +1201,9 @@ class FromTaggedObject' arity f isRecord where
         :: String :* ConName :* TypeName :* Options :* FromArgs arity a
         -> Object -> Tagged isRecord (Parser (f a))
 
-instance (RecordFromJSON arity f) => FromTaggedObject' arity f True where
+instance (RecordFromJSON arity f, FieldNames f) => FromTaggedObject' arity f True where
     -- Records are unpacked in the tagged object
-    parseFromTaggedObject' (_ :* p) = Tagged . recordParseJSON p
+    parseFromTaggedObject' (_ :* p) = Tagged . recordParseJSON (True :* p)
 
 instance (ConsFromJSON arity f) => FromTaggedObject' arity f False where
     -- Nonnullary nonrecords are encoded in the contents field
@@ -1243,11 +1244,11 @@ instance OVERLAPPING_
          ) => ConsFromJSON' arity (S1 s a) True where
     consParseJSON' p@(cname :* tname :* opts :* fargs)
         | unwrapUnaryRecords opts = Tagged . fmap M1 . gParseJSON opts fargs
-        | otherwise = Tagged . withObject (showCons cname tname) (recordParseJSON p)
+        | otherwise = Tagged . withObject (showCons cname tname) (recordParseJSON (False :* p))
 
 instance RecordFromJSON arity f => ConsFromJSON' arity f True where
     consParseJSON' p@(cname :* tname :* _) =
-        Tagged . withObject (showCons cname tname) (recordParseJSON p)
+        Tagged . withObject (showCons cname tname) (recordParseJSON (False :* p))
 
 instance OVERLAPPING_
          ConsFromJSON' arity U1 False where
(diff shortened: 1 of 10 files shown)


bergmark:
Cool, thanks!

Would you mind reviewing this @Lysxia? My generics-fu needs some work...

Data/Aeson/Types/FromJSON.hs
```diff
@@ -1273,38 +1274,71 @@ instance (ProductFromJSON arity f, ProductSize f
...
+            knownFields = (if fromTaggedSum
+                           then (pack (tagFieldName (sumEncoding opts)) :)
+                           else id) $
+                fieldNames (undefined :: f a)
+            checkUnknown =
+                if not (rejectUnknownFields opts)
+                then return ()
+                else case H.keys obj \\ knownFields of
+                    [] -> return ()
+                    unknownFields -> fail ("Unknown fields: " ++ show unknownFields)
```
Lysxia: For a more uniform-looking error message, make this start with lower case and (generics only) append some context with `contextCons`.

```suggestion
                    unknownFields -> contextCons cname tname (fail ("unknown fields: " ++ show unknownFields))
```

`cname` and `tname` would be the two arguments before `opts` inside `p`.

It would be good to add a test case for this to `test/ErrorMessages.hs` (and then run the test with `--accept` to refresh the reference messages).

Data/Aeson/Types/FromJSON.hs
```diff
@@ -1273,38 +1274,71 @@ instance (ProductFromJSON arity f, ProductSize f
 
 --------------------------------------------------------------------------------
 
+class FieldNames f where
+    fieldNames :: f a -> [Text]
+
+instance (FieldNames a, FieldNames b) => FieldNames (a :*: b) where
+    fieldNames _ =
+      fieldNames (undefined :: a x) ++
```
Lysxia: Use a difference list here (if we are to use set difference instead of list difference `\\` below).

```haskell
    fieldNames :: f a -> [Text] -> [Text]
```

```haskell
    fieldNames _ =
      fieldNames (undefined :: a x) .
      fieldNames (undefined :: b x)
```

```haskell
    fieldNames _ = (pack (selName (undefined M1 _i s _f _p)) :)
```

rmanne:
@Lysxia thanks for the feedback, I pushed another commit addressing your previous feedback. Let me know if it looks good now.

A more general question, I assume in the end, you'll want to squash the commits and push as a single commit. How is this done? Does GitHub offer this functionality by default? Or am I supposed to squash locally and `git push -f` to my fork? I haven't attempted this yet in case it'll cause problems with the PR, but do let me know.

Data/Aeson/Types/FromJSON.hs
```diff
@@ -1273,38 +1274,72 @@ instance (ProductFromJSON arity f, ProductSize f
...
 class RecordFromJSON arity f where
     recordParseJSON
+        :: Bool :* ConName :* TypeName :* Options :* FromArgs arity a
+        -> Object -> Parser (f a)
+
+instance ( FieldNames f
+         , RecordFromJSON' arity f
+         ) => RecordFromJSON arity f where
+    recordParseJSON (fromTaggedSum :* p@(cname :* tname :* opts :* _)) =
+        \obj -> checkUnknown obj >> recordParseJSON' p obj
```
Lysxia: Is there a reason for these lambdas? Binding these variables on the left of `=` seems more natural.

Lysxia:
This is in great shape!

Indeed, if you want to squash commits, you must do it locally and force push to Github. Github's web interface is quite limited in that respect (PRs can be squash-merged, but it's not quite the same as "squash-then-merge"). I squash my PRs all the time out of personal preference, but actually I don't know whether there is a policy about it here. @bergmark do you have a preference?

Data/Aeson/Types/FromJSON.hs
rmanne: I'm not really sure how Haskell optimizes. My idea is that I don't want knownFields or if statement in checkUnknown to be computed only when given obj, but rather precomputed, before the function is aware of obj. I know in some functional languages, this matters (eg: sml), which is why I put it this way. Constant propagation would probably make this a non-concern but it doesn't significantly hurt readability anyway.

Lysxia: Fair enough.

rmanne:
@bergmark @Lysxia let me know if there's anything else needed for this PR. I also prefer to squash, so I went ahead and did that.

bergmark:
Agree, looks great! And thanks for reviewing @Lysxia!

I can choose to squash commits when merging on github, but no harm done here.

Tried to figure out the versioning level of this, am I correct that this is just non-breaking additions? It's fine if it's breaking but then I'll rearrange the PR merges to get a minor release out first.

rmanne:
Correct it is non-breaking. The only thing it might break is if someone is using the constructor Options instead of defaultOptions (I don't remember if the constructor is exported so this might not even be possible).

bergmark:
Nope! We un-exposed the constructor exactly for this reason.

PR #750, opened 2020-01-31
Add more informative parsing error message

v0d1ch:
IMO it helps a bit to pinpoint the json part that failed to parse. I just used the information we already have to add slightly more informative error message when using `eitherDecode`.
So for this piece of json where there is a missing `,` before `postal_code` :
```
  {
    "data": {
      "city": "Malakoff",
      "country": null
      "postal_code": "14236",
      "region": "NY",
      "street": "2992 Cameron Road"
    },
    "primary": true
  }

```
We would get:

 `Left "Error in $: Failed reading: satisfy. Expecting object value"`

And with this small improvement, we get:

`Error in $: Failed reading: satisfy. Expecting object value before the line 'postal_code:'`

### DIFF
--- a/Data/Aeson/Parser/Internal.hs
+++ b/Data/Aeson/Parser/Internal.hs
@@ -59,6 +59,7 @@ import Data.Attoparsec.ByteString.Char8 (Parser, char, decimal, endOfInput, isDi
 import Data.Function (fix)
 import Data.Functor.Compat (($>))
 import Data.Scientific (Scientific)
+import Data.Semigroup ((<>))
 import Data.Text (Text)
 import Data.Vector (Vector)
 import qualified Data.Vector as Vector (empty, fromList, fromListN, reverse)
@@ -67,6 +68,9 @@ import qualified Data.Attoparsec.Lazy as L
 import qualified Data.ByteString as B
 import qualified Data.ByteString.Unsafe as B
 import qualified Data.ByteString.Lazy as L
+import qualified Data.ByteString.Lazy as BSL
+import qualified Data.ByteString.Lazy.Char8 as C
+import qualified Data.ByteString.Builder as B
 import qualified Data.HashMap.Strict as H
 import qualified Data.Scientific as Sci
 import Data.Aeson.Parser.Unescape (unescapeText)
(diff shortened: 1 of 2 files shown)


phadej:
What if I have invalid oneline (packed) 10MB json file? There should be some reasonable limit.

Data/Aeson/Parser/Internal.hs
```diff
@@ -390,12 +391,18 @@ eitherDecodeWith p to s =
...
-    buildMsg [] msg = msg
-    buildMsg (expectation:_) msg =
-      msg ++ ". Expecting " ++ expectation
+    buildMsg :: L.ByteString -> [String] -> String -> String
+    buildMsg notParsed [] msg = msg ++ grabErrorLine notParsed
+    buildMsg notparsed (expectation:_) msg =
+      msg ++ ". Expecting " ++ expectation ++ grabErrorLine notparsed
+    grabErrorLine :: L.ByteString -> String
+    grabErrorLine bs =
+        let notParsedStr = C.unpack bs
```
phadej: unpacking 10MB input is also not a good idea.

phadej:
I'm very strict with performance in this parts, the services should not be DDoSable by giving them invalid inputs.

Data/Aeson/Parser/Internal.hs
v0d1ch: Would you be happy with more performant solution? or you want to drop the idea completely?

phadej: Something which
1. [`take`](https://hackage.haskell.org/package/bytestring-0.10.10.0/docs/Data-ByteString-Lazy.html#v:take)s say ~100 characters
2. then after [`takeWhile`](https://hackage.haskell.org/package/bytestring-0.10.10.0/docs/Data-ByteString-Lazy.html#v:takeWhile) `/= 10` (for `'\n'`) and maybe `dropWhile (\w8 -> w8 == 32)` (space, you may want to add tab `9` and CR `13`), i.e. taking single line and dropping whitespace
3. and finally converts that small lazy `ByteString` to `String` for display should be fine

Note: I'm not a maintainer though, it's up to @bergmark 

I also wouldn't try to reformat the input any further, the literal part works quite well in my experience (I use something similar in other parsers). You'll get something like

```
Error in $: Failed reading: satisfy. Expecting object value at '"postal_code": "14236"'
```

v0d1ch: Nice, let's see what @bergmark has to say about this and if the light is green we continue.

Data/Aeson/Parser/Internal.hs
```diff
@@ -390,12 +392,25 @@ eitherDecodeWith p to s =
...
+    buildMsg notparsed (expectation:_) msg =
+      msg ++ ". Expecting " ++ expectation ++ grabErrorLine notparsed
+    grabErrorLine :: L.ByteString -> String
+    grabErrorLine bs =
+        let notParsedStr =
+                C.unpack $
+                 -- if newline is present cut at that position
+                BSL.takeWhile (10 /=)
+                 -- take 100 bytes
+                 (BSL.take 100 $
```
phadej: `BSL.take 100` should be the first thing done. I'm *not sure*, but if one can make parser fail before "very large" whitespace block, the `dropWhile` can do a lot of work.

OTOH, if you have 100 of space in your JSON document, that's pretty easy to spot.

Data/Aeson/Parser/Internal.hs
bergmark: @phadej is as much a maintainer, just permission issues on github T_T. (@phadej if you want to be added to maintainers in .cabal and hackage LMK! I thought I already added you on hackage...)

What will you get from this if your input is only one line? Will it start after the error occurs or show the beginning of the line? I think I usually use compacted JSON, even for testing, but I might be an outlier here!

Sounds good overall

v0d1ch: @phadej I just pushed some changes related to performance. 

> What will you get from this if your input is only one line?  Will it start after the error occurs or show the beginning of the line?

@bergmark It will show the unparsed portion of the bytestring so you should have easier time spotting where exactly the error occured. The newline is not so relevant anymore since we are taking the first 100 characters to show that ` at blah blah`. If error message turns out to be empty this will default to the message we previously had in here.

v0d1ch: I still want to test this a bit after the latest changes.

v0d1ch: Ok guys, I tested this a bit and it seems like it is working fine. If you are happy with the changes we can move on. I think this is just a step forward, perhaps we could utilize a proper diffing algorithm using the input bytestring and the parsed portion so we get even nicer output.

bergmark:
LGTM, thanks!

Do you have anything else to add @phadej?

One concern I always have with changing error messages is that it may change the output users expect (be it test cases or things returned from actual programs), but so far my view has been that we want to encourage improvements like this, so that we can include this in minor releases. Perhaps we should stick this in the README later.

PR #751, opened 2020-02-12
Dry eitherDecodeFileStrict impementation

turboMaCk:
Looking into source of Aeson I noticed these two two minor simplification. `decodeFileStrict` are already implemented in terms of non file specific functions so this makes implementation more consistent.


turboMaCk:
I think this is quite minor and can be merged.

turboMaCk:
@bergmark don't worry about that. I just know how easy it is to forget about PRs in between them being opened and tests passing. Thanks for the merge! :tada:

PR #755, opened 2020-03-17
Clear Warnings

fosskers:
This PR massages some imports to reduce the number of warnings that fly by everyone's screen whenever they build `aeson` as a dependency.


Data/Aeson/Text.hs
```diff
@@ -26,7 +27,9 @@ import Prelude.Compat
 
 import Data.Aeson.Types (Value(..), ToJSON(..))
 import Data.Aeson.Encoding (encodingToLazyByteString)
+#if !MIN_VERSION_base(4,11,0)
```
phadej: I don't like this. We have `NoImplicitPrelude` and `base-compat` / `Prelude.Compat` exactly because we don't want (enable and) copy CPP to every file.

I'd say we should then rather add `other-modules: Data.Aeson.Prelude` and use it, to have CPP in one place only.

attoparsec-iso8601/Data/Attoparsec/Time/Internal.hs
```diff
@@ -19,23 +19,15 @@ module Data.Attoparsec.Time.Internal
 
 import Prelude.Compat
 
+import Data.Fixed (Fixed(MkFixed), Pico)
```
phadej: This is ok because lower bound for `base >=4.7` good!

attoparsec-iso8601/Data/Attoparsec/Time/Internal.hs
```diff
@@ -19,23 +19,15 @@ module Data.Attoparsec.Time.Internal
...
-import Data.Time.Clock (diffTimeToPicoseconds)
-
-#endif
-
-import Data.Fixed (Pico, Fixed(MkFixed))
 
 #if !MIN_VERSION_time(1,6,0)
+import Unsafe.Coerce (unsafeCoerce)
 
 diffTimeToPicoseconds :: DiffTime -> Integer
```
phadej: This always exist through `time-compat` https://hackage.haskell.org/package/time-compat-1.9.3/docs/Data-Time-Clock-Compat.html#v:diffTimeToPicoseconds, i.e. import  `Data.Time.Clock.Compat` and we won't need CPP here.

Good catch.

Data/Aeson/Text.hs
fosskers: Wonderful, let's do that instead.

attoparsec-iso8601/Data/Attoparsec/Time/Internal.hs
fosskers: Great, I'll fix this too.

Data/Aeson/Text.hs
bergmark: we can import Data.Semigroup.Compat instead

fosskers: The already imported `Prelude.Compat` seemed to do the job.

phadej:
Sorry. Seems that I was unclear. `Prelude.Compat` in fact does re-export `Semigroup` when it's in `base`, but not when it isn't. So we need to make own `Data.Aeson.Prelude` to be a superset (or replacement of `base-compat`).

Another alternative is using `base-compat-batteries`. That would add a dependency on `bifunctors` and `type-equality` for older GHCs. That would be simple solution. Opinions @bergmark ?

fosskers:
Currently hitting this issue with the 7.10 and 7.8 builds in CI. We could otherwise leave the CPP in for the two modules where the `Semigroup` symbol is actually required.

bergmark:
> Sorry. Seems that I was unclear. Prelude.Compat in fact does re-export Semigroup when it's in base, but not when it isn't.

Ah of course. I'd vote for base-compat-batteries.

phadej:
I made a PR with `base-compat-batteries` based on this one: https://github.com/bos/aeson/pull/759

bergmark:
Thanks! This was merged as part of #759.

fosskers:
Thanks guys!

PR #756, opened 2020-03-17
MonadFail-less failing

fosskers:
This PR introduces the combinator `parseFail` which allows downstream users to signal a failure without interacting with `MonadFail`. Yes, it is implemented in terms of `fail`, but that's not something that the user needs to know, and it's free to be changed in future.

This brings the `Parser` type from `aeson` in line with similar `Parser` types from other libraries.

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -346,7 +346,6 @@ instance Monoid (Parser a) where
 -- | Raise a parsing failure with some custom message.
 parseFail :: String -> Parser a
 parseFail = fail
-{-# INLINE parseFail #-}
 
 apP :: Parser (a -> b) -> Parser a -> Parser b
 apP d e = do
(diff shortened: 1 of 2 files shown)


Data/Aeson/Types/Internal.hs
```diff
@@ -342,6 +343,11 @@ instance Monoid (Parser a) where
     mappend = (<>)
     {-# INLINE mappend #-}
 
+-- | Raise a parsing failure with some custom message.
+parseFail :: String -> Parser a
+parseFail = fail
+{-# INLINE parseFail #-}
```
phadej: `{-# INLINE #-}` is IMO not necessary. I'm quite sure GHC does right thing here. (`aeson` have too many `INLINE`s; we should really remove probably most of them).

fosskers: Haha I added the INLINE precisely because I thought I'd be yelled at for not! ;)

Fixed now.

bergmark:
Thanks!

PR #758, opened 2020-03-18
Fix #757 by making type family detection smarter

### LINKED ISSUE #757: Deriving ToJSON1 et al. chokes on type families with an appropriate return kind
Consider the following setup:

```hs
{-# LANGUAGE DeriveAnyClass #-}
{-# LANGUAGE DerivingStrategies #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE StandaloneDeriving #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE UndecidableInstances #-}
{-# OPTIONS_GHC -ddump-splices #-}
module Bug where

import Data.Aeson
import Data.Aeson.TH
import Data.Kind
import GHC.Generics

type family F :: Type -> Type
type instance F = Maybe
data T a = MkT (F a)
```

One can derive a `ToJSON1` instance for `T` using `GHC.Generics`:

```hs
deriving stock instance Generic1 T
deriving anyclass instance ToJSON1 T
```

Trying to use Template Haskell to accomplish the same thing, however, fails:

```hs
$(deriveToJSON1 defaultOptions ''T)
```
```
Bug.hs:1:1: error:
    Exception when trying to run compile-time code:
      Constructor ‘MkT‘ must only use its last 1 type variable(s) within the last 1 argument(s) of a data type
CallStack (from HasCallStack):
  error, called at ./Data/Aeson/TH.hs:1880:35 in aeson-1.4.6.0-25a1b5ff7920dcfd75dcc5d329c556719dbf4e7f39048818f67588ff29ec9daf:Data.Aeson.TH
    Code: deriveToJSON1 defaultOptions ''T
  |
1 | {-# LANGUAGE DeriveAnyClass #-}
  | ^
```

The issue is that `deriveToJSON1` is too conservative: it bails out if the last type variable (`a`) occurs as the argument to any type family whatsoever. Although `F` is a type family, its use in `MkT` is benign, since the type `F a` oversaturates `F` (it has zero type variable binders). `deriveToJSON1` should take the number of type variable binders a type family has into account so as to permit the program above.

Patch incoming.

RyanGlScott:
This fixes #757 by replacing `isTyFamily` with `isInTypeFamilyApp`, which considers the number of type variable binders a type family has when determining if a type variable appears as an argument to the type family or not.

### DIFF
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -1304,9 +1304,8 @@ dispatchFunByType jc jf conName tvMap list ty = do
         tyVarNames :: [Name]
         tyVarNames = M.keys tvMap
 
-    itf <- isTyFamily tyCon
-    if any (`mentionsName` tyVarNames) lhsArgs
-          || itf && any (`mentionsName` tyVarNames) tyArgs
+    itf <- isInTypeFamilyApp tyVarNames tyCon tyArgs
+    if any (`mentionsName` tyVarNames) lhsArgs || itf
        then outOfPlaceTyVarError jc conName
        else if any (`mentionsName` tyVarNames) rhsArgs
             then appsE $ varE (jsonFunValOrListName list jf $ toEnum numLastArgs)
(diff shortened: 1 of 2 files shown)


RyanGlScott:
I'm not sure where a good place to add a regression test would be, so let me know what your preference is.

bergmark:
Thanks! I'm not sure where to place the test either...

RyanGlScott:
It appears that `tests/UnitTests.hs` is the only place where `deriveToJSON1` is currently tested, so I added a regression test there. I've put the changes in a separate commit just in case.

bergmark:
Great!

PR #760, opened 2020-03-26
Implement liftTyped in the Lift Value instance

RyanGlScott:
When building `aeson` with GHC 8.10.1, I get the following warning:

```
[ 5 of 24] Compiling Data.Aeson.Types.Internal ( Data/Aeson/Types/Internal.hs, /home/rgscott/Documents/Hacking/Haskell/aeson/dist-newstyle/build/x86_64-linux/ghc-8.10.1/aeson-1.4.7.0/build/Data/Aeson/Types/Internal.o, /home/rgscott/Documents/Hacking/Haskell/aeson/dist-newstyle/build/x86_64-linux/ghc-8.10.1/aeson-1.4.7.0/build/Data/Aeson/Types/Internal.dyn_o )

Data/Aeson/Types/Internal.hs:415:10: warning: [-Wmissing-methods]
    • No explicit implementation for
        ‘TH.liftTyped’
    • In the instance declaration for ‘TH.Lift Value’
    |
415 | instance TH.Lift Value where
    |          ^^^^^^^^^^^^^
```

This is because `template-haskell-2.16.0.0` introduced the `liftTyped` method to the `Lift`. Normally, this method would be filled in by `DeriveLift`, but the `Lift Value` instance is hand-written. In such a scenario, the simplest, most backwards-compatible solution is to implement `liftTyped` as `unsafeTExpCoerce . lift`. The use of `unsafeTExpCoerce` here is safe as long as `lift` is implemented correctly. (Which it should be!)

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -425,6 +425,9 @@ instance TH.Lift Value where
       where a' = V.toList a
     lift (Object o) = [| Object (H.fromList . map (first pack) $ o') |]
       where o' = map (first unpack) . H.toList $ o
+#if MIN_VERSION_template_haskell(2,16,0)
+    liftTyped = TH.unsafeTExpCoerce . TH.lift
+#endif
 
 -- | The empty array.
 emptyArray :: Value
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -106,8 +106,8 @@ library
     bytestring       >= 0.10.4.0 && < 0.11,
     containers       >= 0.5.5.1 && < 0.7,
     deepseq          >= 1.3.0.0 && < 1.5,
-    ghc-prim         >= 0.2     && < 0.6,
-    template-haskell >= 2.9.0.0 && < 2.16,
+    ghc-prim         >= 0.2     && < 0.7,
+    template-haskell >= 2.9.0.0 && < 2.17,
     text             >= 1.2.3.0 && < 1.3,
     time             >= 1.4     && < 1.10
 


phadej:
Oh crap. So my bound relaxation is actually invalid.

@bergmark I'll prepare a tarball based on `1.4.7.0`.

phadej:
https://github.com/bos/aeson/compare/master...phadej:v1.4.7._ the patch. I cannot make an upload.

RyanGlScott:
> Oh crap. So my bound relaxation is actually invalid.

It depends on what you mean by "invalid". `aeson-1.4.7.0` still compiles with GHC 8.10.1, and pre-existing code paths involving `aeson` should still work under 8.10.1. The only thing that wouldn't work is any new code that uses `liftTyped`, but given that that wasn't introduced until 8.10.1 either, I don't think it's worth sweating over too much.

> [master...phadej:v1.4.7](https://github.com/bos/aeson/compare/master...phadej:v1.4.7)._ the patch. Yet I cannot upload it.
> 
> EDIT: @RyanGlScott does it look right?

I can't tell if it looks right or not. The link above simply takes me to a page that says "There isn’t anything to compare."

RyanGlScott:
On the topic of this PR itself, it looks like there are two spurious Travis failures. I am unable to restart them myself, however.

phadej:
> I can't tell if it looks right or not. The link above simply takes me to a page that says "There isn’t anything to compare."

weird, there are changes:

```
% git log master.. --oneline
3577981 (HEAD -> v1.4.7._, phadej/v1.4.7._) Bump version to 1.4.7.1
9caeea8 Implement liftTyped in the Lift Value instance
fbd5c57 Bump ghc-prim, template-haskell version bounds to build with GHC 8.10.1
```

phadej:
> I don't think it's worth sweating over too much.

No. I'd still like to have a release, and undo the revision I made for `1.4.7.0`. Having it as such for too long would cause Matthew Pickering to discover this bug. I'm like 101% sure.

phadej:
EDIT: In particular `DeriveLift` probably picks `liftTyped`, so we can transitively and implicitly depend on broken instance.

RyanGlScott:
> weird, there are changes:

Ah, I see what happened. GitHub's auto-hyperlinker accidentally left out the `._` part of the URL! If I click the [full URL](https://github.com/bos/aeson/compare/master...phadej:v1.4.7._), _then_ I can see the changes you were referring to. And yes, those look fine to me, although there should probably be a changelog entry as well. Let me know if I should write one, or if the maintainers would prefer writing it themselves.

> I'd still like to have a release, and undo the revision I made for `1.4.7.0`. Having it as such for too long would cause Matthew Pickering to discover this bug. I'm like 101% sure.

Fair enough—I certainly won't complain about more frequent releases. And I definitely wouldn't want to invite the wrath of Matthew Pickering :)

bergmark:
Thanks! I'll make a release.

bergmark:
Pushed through @phadej's branch.

phadej:
@bergmark thanks for prompt release! I undid the revision of 1.4.7.0 https://hackage.haskell.org/package/aeson-1.4.7.0/revisions/

PR #765, opened 2020-04-02
Use haskell-ci for travis

phadej:


bergmark:
Thanks!

Looks like the custom build steps are not included here? Would be nice to keep them.
* Build benchmarks
* Run HLint
* Build & test with the `cffi` flag

phadej:
I can make those as follow-ups. Last two (HLint and cffi) are easy. Benchmarks needs a bit of stuff, (and I might be motivated by #766 )

bergmark:
Sounds good!

Issue #659, opened 2018-08-06
FromJSONKeyCoerce warning


MaxGabriel:
Hi, I just installed aeson-1.4.0.0 and got this warning:

```
Data/Aeson/Types/FromJSON.hs:468:3: warning: [-Winline-rule-shadowing]
    Rule "FromJSONKeyCoerce: fmap id" may never fire
      because ‘id’ might inline first
    Probable fix: add an INLINE[n] or NOINLINE[n] pragma for ‘id’
    |
468 |   "FromJSONKeyCoerce: fmap id"     forall (x :: FromJSONKeyFunction a).
    |
```

It's not causing me any problems, just wanted to let you know.

Looks like the warning may have been around awhile since it appears here as well: https://github.com/bos/aeson/issues/543

bergmark:
Thanks,

I'm not sure if id will always inline first or not? If it does we might as well remove the rule, but otoh it shouldn't hurt to keep it. We would want to keep the rule if it triggers sometimes, if so we could just disable the warning. I'm sure  there is a reason `id` isn't marked with (NO)INLINE.

PR #770, opened 2020-04-04
Fix #741: Remove CoerceText, GHC >=7.8 has Coercible

### LINKED ISSUE #741: Export CoerceText for usage.
Around https://github.com/bos/aeson/pull/399/commits/5ef5530700f1f44a72cbac959070d5a0ffe1ada4 a new type was added, CoerceText. This made coercing keys safer which is great. Sadly, this is not exported and it'd be useful if it was.

A motivating story below.

We have a [maplike class](https://github.com/takano-akio/maplike) that we use for working with strict/lazy [EnumMap](http://hackage.haskell.org/package/enummapset-th)s. I wanted to add an (orphan) instance for these EnumMap's. I started with a simple `EnumMap.fromList` but that obviously gave us the ugly list representation after serialisation. I looked in `aeson` source and followed the footsteps of `Data.Map` implementation. I got something like this

```hs
maplikeParseJSON
    :: forall k v m. (Maplike.Maplike k m, Typeable (m v), A.FromJSONKey k, A.FromJSON v)
    => A.Value -> A.Parser (m v)
maplikeParseJSON = case A.fromJSONKey of
    -- Neither coerceFromJSONKeyFunction nor CoerceText that we
    -- need to match on to get a Coercible witness are exposed
    -- from a public aeson module so we have to do an unsafeCoerce
    -- just like older aeson on older GHC version does. The
    -- constructor here is our witness.
    A.FromJSONKeyCoerce _ -> A.withObject typeName
        $ fmap (HM.foldrWithKey (Maplike.insert . unsafeCoerce) Maplike.empty)
        . HM.traverseWithKey (\k v -> A.parseJSON v A.<?> A.Key k)
    A.FromJSONKeyText f -> A.withObject typeName
        $ fmap (HM.foldrWithKey (Maplike.insert . f) Maplike.empty)
        . HM.traverseWithKey (\k v -> A.parseJSON v A.<?> A.Key k)
    A.FromJSONKeyTextParser f -> A.withObject typeName
        $ HM.foldrWithKey (\k v m -> Maplike.insert <$> f k A.<?> A.Key k <*> A.parseJSON v A.<?> A.Key k <*> m) (pure Maplike.empty)
    A.FromJSONKeyValue f -> A.withArray typeName
        $ fmap Maplike.fromList . sequence
        . zipWith (parseIndexedJSONPair f) [0 ..] . V.toList
    where
        typeName = show $ typeRep (Proxy :: Proxy (m v))
{-# INLINE maplikeParseJSON #-}
```

Notably, the problem is on FromJSONKeyCoerce case. We're forced to use `unsafeCoerce` even though we should be able to get `Coercible` constraint from its argument. If `CoerceText` was exported, I could simply use `coerce`.

phadej:
This is obviously a breaking change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -160,7 +160,7 @@ script:
   - if [ $HCNUMVER -ge 80800 ] && [ $HCNUMVER -lt 81000 ] ; then (cd ${PKGDIR_aeson} && hlint -h ${TOP}/.hlint.yaml . attoparsec-iso8601/ pure ffi) ; fi
   - if [ $HCNUMVER -ge 80800 ] && [ $HCNUMVER -lt 81000 ] ; then (cd ${PKGDIR_attoparsec_iso8601} && hlint -h ${TOP}/.hlint.yaml .) ; fi
   # haddock...
-  - ${CABAL} v2-haddock $WITHCOMPILER --with-haddock $HADDOCK ${TEST} ${BENCH} all
+  - if [ $HCNUMVER -ge 80600 ] ; then ${CABAL} v2-haddock $WITHCOMPILER --with-haddock $HADDOCK ${TEST} ${BENCH} all ; fi
   # Building without installed constraints for packages in global-db...
   - rm -f cabal.project.local
   - ${CABAL} v2-build $WITHCOMPILER --disable-tests --disable-benchmarks all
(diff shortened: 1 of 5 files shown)


phadej:
Also fixes https://github.com/bos/aeson/issues/659

phadej:
This fails due haddocks for GADTs. There are two options:
- Either remove comments (me dislikes)
- disable haddock buildings for GHC <8.6 on CI. It might break some users, but Hackage and Stackage will be able to build docs.

I pushed a commit for the second option. This will conflict with #767 though.

phadej:
Resolved conflict.

PR #772, opened 2020-04-04
Add cabal.bench.project for running benchmarks

phadej:


CONTRIBUTING.md
```diff
@@ -96,3 +96,23 @@ Of course before submitting a PR, the following steps are recommended:
...
+
+### Running benchmarks
+
+You need to install `cabal-plan` and `criterion-compare` with
+
+```
+cabal install cabal-plan criterion-compare
+```
+
+Then to build benchmarks we use different project, which builds
```
bergmark, line 108:
```suggestion
Then to build benchmarks we use a different project, which builds
```

CONTRIBUTING.md
```diff
@@ -96,3 +96,23 @@ Of course before submitting a PR, the following steps are recommended:
...
+`aeson` as a package of different name to avoid rebuilding `criterion`
```
bergmark, line 109:
```suggestion
`aeson` as a package with a different name to avoid rebuilding `criterion`
```

phadej:
The CI is green now. So next I'll refactor the benchmarks a bit.

aeson.cabal
```diff
@@ -35,9 +35,6 @@ extra-source-files:
     benchmarks/json-data/*.json
     cbits/*.c
     changelog.md
-    examples/*.cabal
```
bergmark, line 38: Why remove these?

bergmark:
Looks like we need base-compat-batteries for the benchmarks as well

aeson.cabal
phadej: There are two points:
1. I don't think anyone is actually looking for the examples in `.tar.gz`. The tarball is quite well abstracted and hidden by tools like `cabal` and `stack`. I think it's more likely people will find examples through GitHub, than by manually unpacking the tarball and looking there.
2. It may make easier to add examples which are not under BSD-3-Clause license. I thought about e.g. `cabal-plan` implementation of `plan.json`, which is GPLd. One could find something under Apache-2.0 too. Anyway, not worrying about licenses is simpler, and one doesn't need to worry if one doesn't distribute :)

phadej:
CI green (finally). Please give it through review. I'd like to merge what one has here before adding more decode (and encode) cases and refactoring the existing benchmarks to be part of the suite.

aeson.cabal
bergmark: Sounds good!

bergmark:
LG! Tested it out, easy to use by just reading CONTRIBUTING. Very nice to avoid the recompilations.

Issue #700, opened 2019-05-14
Expose GToJSON


LukaHorvat:
Currently, only `GFromJSON` is exposed. `GToJSON` is only exposed via the two type synonyms which means that I can't write custom instances for that class. My use case is that I have a type (`data Unset = Unset`) that I want to use for some fields in a record and I want those fields to be dropped from the JSON representation of that record. If `GToJSON` was exposed, I could just duplicate the trick used for `Maybe`. 

Since `GFromJSON` is already exposed, this already has a precedent.

phadej:
Writing custom instances of `GToJSON` seems somewhat phishy.

LukaHorvat:
Sure, and I would discourage that in the docs, but it probably shouldn't be forbidden.

bergmark:
Can we export it without making a breaking change (due to the type synonym)? Sounds reasonable to export it since GFromJSON is exported.

@LukaHorvat how are you using Unset, with a parameterized data type? Maybe we could add that to aeson?

LukaHorvat:
@bergmark Yep, a parametrized data type. Maybe this could be added but it seems very specific to be honest.

Currently I'm working around this issue by defining a newtype which imitates the generic rep of the type it's wrapping but drops the Unset fields. Then I use XDerivingVia to get the JSON instances for my original types. I'm kind of happy with the solution since it's very general and I might need to reuse it for other classes, but I'd be even happier if I could drop these 50 lines of type class/family magic and just define a custom GToJSON instance.

blinky3713:
I'm trying to do something similar, see [this comment](https://github.com/f-o-a-m/hs-abci-server/issues/19#issuecomment-515793384). If this is considered a bad idea, I would also take your advice and avoid doing it.

Lysxia:
@LukaHorvat's trick sounds like what [generic-data-surgery](https://hackage.haskell.org/package/generic-data-surgery) does, and @blinky3713's code looks like what [higgledy](https://hackage.haskell.org/package/higgledy) does. There's certainly a cost in complexity for these abstractions in exchange for the added modularity.

phadej:
@bergmark there is

```haskell
Data/Aeson/Types/Class.hs:type GToJSON = ToJSON.GToJSON Value
Data/Aeson/Types/ToJSON.hs:class GToJSON enc arity f where
```

IMO we should rename either, and have both exported.

PR #774, opened 2020-05-03
Fix reject unknown fields with field label modifier

### LINKED ISSUE #773: `rejectUnknownFields` does not respect `fieldLabelModifier`
Assuming a `FromJSON` instance like this:

```haskell
-- have not run this version, just for illustration
data Foo = Foo
  { a :: Int
  }

instance FromJSON Foo where
  parseJSON = genericParseJSON defaultOptions { rejectUnknownFields = True, fieldLabelModifier = rename }
    where
      rename   "a" = "b"
      rename other = other
```

Parsing of will fail with a valid input like: `{"b":1}` will fail with an error like:

```
parsing Foo(Foo) failed, unknown fields ["b"]
```

The `fieldLabelModifier` in the code path to `knownFields` does never apply the `fieldLabelModifier`. 

I'm trying to come up with a PR to fix this.

mbj:
My attempt to fix #773 I reported earlier.

This PR comes in 2 commits. The first adds error message gets for the TH parsing. The 2nd commit fixes the problem I reported with changing the (IMO wrong) error expectations to demonstrate the bugfix.

There are some minor implementation trade-offs. I'll comment in code below to highlight my thoughts on them.

### DIFF
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -1260,7 +1260,7 @@ instance (ProductFromJSON arity f, ProductSize f
 --------------------------------------------------------------------------------
 
 class FieldNames f where
-    fieldNames :: f a -> [Text] -> [Text]
+    fieldNames :: f a -> [String] -> [String]
 
 instance (FieldNames a, FieldNames b) => FieldNames (a :*: b) where
     fieldNames _ =
@@ -1268,7 +1268,7 @@ instance (FieldNames a, FieldNames b) => FieldNames (a :*: b) where
       fieldNames (undefined :: b y)
 
 instance (Selector s) => FieldNames (S1 s f) where
-    fieldNames _ = (pack (selName (undefined :: M1 _i s _f _p)) :)
+    fieldNames _ = (selName (undefined :: M1 _i s _f _p) :)
 
 class RecordFromJSON arity f where
     recordParseJSON
@@ -1282,9 +1282,10 @@ instance ( FieldNames f
         \obj -> checkUnknown obj >> recordParseJSON' p obj
         where
             knownFields :: H.HashMap Text ()
-            knownFields = H.fromList $ map (,()) $
-                fieldNames (undefined :: f a)
-                [pack (tagFieldName (sumEncoding opts)) | fromTaggedSum]
+            knownFields = H.fromList $ map ((,()) . pack) $
+                [tagFieldName (sumEncoding opts) | fromTaggedSum] <>
+                (fieldLabelModifier opts <$> fieldNames (undefined :: f a) [])
+
             checkUnknown =
                 if not (rejectUnknownFields opts)
                 then \_ -> return ()
--- a/changelog.md
+++ b/changelog.md
@@ -1,9 +1,11 @@
 For the latest version of this document, please see [https://github.com/bos/aeson/blob/master/changelog.md](https://github.com/bos/aeson/blob/master/changelog.md).
 
+## 1.5.0.0
+* Fix bug in `rejectUnknownFields` not respecting `fieldLabelModifier`, thanks to Markus Schirp.
+
 #### 1.4.7.1
 
 * GHC 8.10 compatibility, thanks to Ryan Scott.
-* Fix bug in `rejectUnknownFields` not respecting `fieldLabelModifier`.
 
 ### 1.4.7.0
 
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@ For the latest version of this document, please see [https://github.com/bos/aeso
 #### 1.4.7.1
 
 * GHC 8.10 compatibility, thanks to Ryan Scott.
+* Fix bug in `rejectUnknownFields` not respecting `fieldLabelModifier`.
 
 ### 1.4.7.0
 
--- a/tests/ErrorMessages.hs
+++ b/tests/ErrorMessages.hs
@@ -31,6 +31,7 @@ tests :: [TestTree]
 tests =
   [ aesonGoldenTest "simple" "tests/golden/simple.expected" output
   , aesonGoldenTest "generic" "tests/golden/generic.expected" (outputGeneric G)
+  , aesonGoldenTest "generic" "tests/golden/th.expected" (outputGeneric TH)
   ]
 
 output :: Output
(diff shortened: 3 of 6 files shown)


Data/Aeson/Types/FromJSON.hs
```diff
@@ -1260,15 +1260,15 @@ instance (ProductFromJSON arity f, ProductSize f
 --------------------------------------------------------------------------------
 
 class FieldNames f where
-    fieldNames :: f a -> [Text] -> [Text]
+    fieldNames :: f a -> [String] -> [String]
```
mbj, line 1263: I changed the type signature from `[Text]` to `[String]` deliberately to be able to call the `fieldLabelModifier` without a `pack` / `unpack` cycle.

Data/Aeson/Types/FromJSON.hs
```diff
@@ -1282,9 +1282,10 @@ instance ( FieldNames f
         \obj -> checkUnknown obj >> recordParseJSON' p obj
         where
             knownFields :: H.HashMap Text ()
-            knownFields = H.fromList $ map (,()) $
-                fieldNames (undefined :: f a)
-                [pack (tagFieldName (sumEncoding opts)) | fromTaggedSum]
+            knownFields = H.fromList $ map ((,()) . pack) $
+                [(tagFieldName (sumEncoding opts)) | fromTaggedSum] <>
+                (fieldLabelModifier opts <$> fieldNames (undefined :: f a) [])
```
mbj, line 1287: The `fieldLabelModifier` should not be applied to the "tag" fields. Hence I could not anymore pass the tag fields into the field names reflection.

I had to choose between: A list concatentation (and these can be inefficient) and a union of 2 hashes, one build from the tag fields and one from the reflected field names.

I went with the list concatenation, as I assume that the number of tag fields will usually be lower and appending to a "small" list is quite fast. Hence also putting the "smaller" list on the LHs of `<>`.

changelog.md
```diff
@@ -3,6 +3,7 @@ For the latest version of this document, please see [https://github.com/bos/aeso
 #### 1.4.7.1
 
 * GHC 8.10 compatibility, thanks to Ryan Scott.
+* Fix bug in `rejectUnknownFields` not respecting `fieldLabelModifier`.
```
mbj, line 6: I've noted that the changelog entries typically say `thanks to $Author` but I think I should not thank myself, especially while my code is under review. I hope the reviewer edits it in, if appropriate ;)

tests/ErrorMessages.hs
```diff
@@ -141,7 +141,7 @@ outputGeneric choice = concat
       (select
         thSomeTypeParseJSONRejectUnknownFields
         gSomeTypeParseJSONRejectUnknownFields)
-      [ "{\"tag\": \"record\", \"testOne\": 1.0, \"testZero\": 1}"
+      [ "{\"tag\": \"record\", \"testone\": 1.0, \"testZero\": 1}"
```
mbj, line 144: `optsRejectUnknownFields` actually had the `fieldLabelModifier = map toLower` set all the time. Hence this test input should have rejected the `testOne` originally and prior to my bugfix did not. 

I chose to not change the expectations in the golden test but instead made the test input comply with the expectation.

tests/golden/generic.expected
```diff
@@ -34,7 +34,7 @@ Error in $: not enough input. Expecting json list value
 SomeType (reject unknown fields)
 Error in $: parsing Types.SomeType(Record) failed, unknown fields: ["testZero"]
 Error in $: parsing Types.SomeType failed, expected Object with key "tag" containing one of ["nullary","unary","product","record","list"], key "tag" not found
-Error in $: parsing Types.SomeType(Record) failed, unknown fields: ["testtwo","testone","testthree"]
+Error in $.testone: parsing Double failed, unexpected Boolean
```
mbj, line 37: The "real" error message was uncovered by my change. The previously reported fields where not unknown as the test case has `fieldLabelModifier = map toLower` applied.

tests/golden/th.expected
```diff
@@ -34,7 +34,7 @@ Error in $: not enough input. Expecting json list value
 SomeType (reject unknown fields)
 Error in $: Unknown fields: ["testZero"]
 Error in $: key "tag" not found
-Error in $: Unknown fields: ["testtwo","testone","testthree"]
+Error in $.testone: parsing Double failed, unexpected Boolean
```
mbj, line 37: Same as https://github.com/bos/aeson/pull/774/commits/70d0927aa32481d5f4d312c8c550763fe7d41e27#r419038724 but for TH.

I noticed there where no tests to update to validate the TH bugfix so I created the first commit and extracted it so the reviewer can also see the TH behavior change being reflected in tests.

mbj: For this comment to make sense, look at the 2nd commit in isolation.

Data/Aeson/Types/FromJSON.hs
bergmark: Sounds good!

changelog.md
bergmark: Yeah i always go through this before release, but 1.4.7.1 is already out. We usually put new stuff under an Upcoming header, but I can do that when merging.

mbj:
@bergmark Thanks for the quick review. What is the process after your approval?

bergmark:
Nothing for this PR!

@phadej Did you have any other breaking changes you wanted to include in the next release or should we ship 1.5.0.0?

mbj:
> Nothing for this PR!
> 
> @phadej Did you have any other breaking changes you wanted to include in the next release or should we ship 1.5.0.0?

Any chance we can get this released? Can I help somewhere to make it happen?

phadej:
@bergmark the scanner rework is something which will force breaking change. Yet, I don't think that will be ready soon.

I'd merge
- #653
and fix #700

and then I think 1.5 would be solid.

mbj:
> and then I think 1.5 would be solid.

Chances we can get 1.4.7.2 meanwhile, which includes this PR? The [changelog](https://github.com/bos/aeson/blob/master/changelog.md) so far does not list a breaking change, so cutting another 1.4 may be eligible. This way we do not have to wait to get the fix out.

bergmark:
Sorry for the delay @mbj, 1.5.0.0 was released now. We already had some breaking changes that weren't listed in the changelog. Best place to check for things like that is versioning labels on the current milestone, e.g. https://github.com/bos/aeson/milestone/21?closed=1

mbj:
@bergmark Thanks for the release and pinging me, much appreciated. Cheers.

PR #775, opened 2020-05-16
Resolve #700. Expose GToJSON.

### LINKED ISSUE #700: Expose GToJSON

phadej:
Rename the type-class to be `GToJSON'` (with prime),
to avoid name clash with type-aliases GToJSON and GToEncoding

---

This is the second of remaining 1.5 tasks.

### DIFF
--- a/Data/Aeson.hs
+++ b/Data/Aeson.hs
@@ -91,6 +91,7 @@ module Data.Aeson
     , FromArgs
     , GToJSON
     , GToEncoding
+    , GToJSON'
     , ToArgs
     , Zero
     , One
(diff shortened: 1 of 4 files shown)


phadej:
EDIT: In fact, I'll be happy with the state for 1.5 to be released. #653 needs a bit of thinking, I won't have time for that soon. (And we have another major release on the horizon with scanner stuff). So let's not wait any longer. ping @bergmark

bergmark:
Thanks!

PR #789, opened 2020-06-18
Make unit parsing more flexible

### LINKED ISSUE #788: Instance for `()` is unnecessarily strict
```
Prelude Data.Aeson> encode ()
"[]"
Prelude Data.Aeson> decode "[]"
Just ()
Prelude Data.Aeson> decode "{}"
Nothing
```

Could we make `decode "{}" == Just ()`?  Are there strong objections to such a change?  Happy to make a PR.

TomMD:
Fix #788 

Current behavior:

```
Prelude Data.Aeson> encode ()
"[]"
Prelude Data.Aeson> decode "[]"
Just ()
Prelude Data.Aeson> decode "{}"
Nothing
```

New behavior:

```
*Data.Aeson> decode "[]" :: Maybe ()
Just ()
*Data.Aeson> decode "{}" :: Maybe ()
Just ()
```

But still:

```
*Data.Aeson> decode "{ \"foo\" : \"bar\" }" :: Maybe ()
Nothing
```

### DIFF
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -1526,7 +1526,7 @@ instance FromJSON () where
                     then pure ()
                     else prependContext "()" $ fail "expected an empty array or object") j <|>
                 (withObject "()" $ \o ->
-                       if H.empty 0
+                       if H.empty o
                          then pure ()
                          else prependContext "()" $ fail "expected an empty array or object") j
     {-# INLINE parseJSON #-}
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -1526,7 +1526,7 @@ instance FromJSON () where
                     then pure ()
                     else prependContext "()" $ fail "expected an empty array or object") j <|>
                 (withObject "()" $ \o ->
-                       if H.size o == 0
+                       if H.empty 0
                          then pure ()
                          else prependContext "()" $ fail "expected an empty array or object") j
     {-# INLINE parseJSON #-}
--- a/Data/Aeson/Types/FromJSON.hs
+++ b/Data/Aeson/Types/FromJSON.hs
@@ -1521,10 +1521,14 @@ instance FromJSON Ordering where
                   " (expected \"LT\", \"EQ\", or \"GT\")"
 
 instance FromJSON () where
-    parseJSON = withArray "()" $ \v ->
+    parseJSON j = (withArray "()" $ \v ->
                   if V.null v
                     then pure ()
-                    else prependContext "()" $ fail "expected an empty array"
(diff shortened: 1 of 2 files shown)


TomMD:
There is now a test for `decode "{}" == Just ()`.

tests/SerializationFormatSpec.hs
```diff
@@ -83,7 +83,7 @@ jsonExamples =
   , example "NonEmpty" "[1,2,3]"  (1 :| [2, 3] :: NonEmpty Int)
   , example "Seq" "[1,2,3]"  (Seq.fromList [1, 2, 3] ::  Seq.Seq Int)
   , example "DList" "[1,2,3]"  (DList.fromList [1, 2, 3] :: DList.DList Int)
-  , example "()" "[]"  ()
+  , Example "()" ["[]", "{}"]  () ()
```
phadej, line 86: The extra case should be added to `jsonDecodingExamples`. This example (which is `ndExample`) tells that there would be a system that `()` could encode as `{}`, which isn't true.

TomMD: I believe this issue is now fixed.

Data/Aeson/Types/FromJSON.hs
```diff
@@ -1521,10 +1521,14 @@ instance FromJSON Ordering where
...
 
 instance FromJSON () where
-    parseJSON = withArray "()" $ \v ->
+    parseJSON j = (withArray "()" $ \v ->
                   if V.null v
                     then pure ()
-                    else prependContext "()" $ fail "expected an empty array"
+                    else prependContext "()" $ fail "expected an empty array or object") j <|>
+                (withObject "()" $ \o ->
+                       if H.size o == 0
```
bergmark, line 1529:
```suggestion
                       if H.empty 0
```
i suppose! O(1) instead of O(n)

bergmark: added the commit

bergmark: ... proof i shouldn't edit code on github

Data/Aeson/Types/FromJSON.hs
```diff
@@ -1521,10 +1521,14 @@ instance FromJSON Ordering where
...
 
 instance FromJSON () where
-    parseJSON = withArray "()" $ \v ->
+    parseJSON j = (withArray "()" $ \v ->
                   if V.null v
                     then pure ()
-                    else prependContext "()" $ fail "expected an empty array"
+                    else prependContext "()" $ fail "expected an empty array or object") j <|>
+                (withObject "()" $ \o ->
+                       if H.empty 0
```
bergmark, line 1529:
```suggestion
                       if H.empty o
```

phadej:
See my comment: https://github.com/haskell/aeson/issues/788#issuecomment-939328524

PR #793, opened 2020-07-21
Bumps dlist upper bound

jkachmar:
This seems to compile without issue locally, but I haven't exhaustively compared the changelog with what `aeson` uses to see if there's something deeper going on.


jkachmar:
re. commercialhaskell/stackage#5524, a revision could be used to unblock `aeson-1.4.7.1` (which is the current version tracked by Stackage nightly) for the time being.

jkachmar:
CI's failing with a `curl` error, seems unrelated to this change fwiw.

PR #802, opened 2020-08-27
Support template-haskell 2.17.

galenhuntington:
The new version of `template-haskell` released with GHC 9 added a `Code` newtype (https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3358).  This package builds under GHC 9.0.1-alpha1 with this change.

I'm not able to run the tests because `tasty-quickcheck` doesn't build under GHC 9 because of other TH type changes, but this change is pretty straightforward (and should not affect end users).

### DIFF
--- a/Data/Aeson/Types/Internal.hs
+++ b/Data/Aeson/Types/Internal.hs
@@ -427,7 +427,7 @@ instance TH.Lift Value where
     lift (Object o) = [| Object (H.fromList . map (first pack) $ o') |]
       where o' = map (first unpack) . H.toList $ o
 #if MIN_VERSION_template_haskell(2,17,0)
-    liftTyped = TH.Code . TH.unsafeTExpCoerce . TH.lift
+    liftTyped = TH.unsafeCodeCoerce . TH.lift
 #elif MIN_VERSION_template_haskell(2,16,0)
     liftTyped = TH.unsafeTExpCoerce . TH.lift
 #endif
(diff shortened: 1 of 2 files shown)


phadej (approved):

Data/Aeson/Types/Internal.hs
```diff
@@ -426,7 +426,9 @@ instance TH.Lift Value where
       where a' = V.toList a
     lift (Object o) = [| Object (H.fromList . map (first pack) $ o') |]
       where o' = map (first unpack) . H.toList $ o
-#if MIN_VERSION_template_haskell(2,16,0)
+#if MIN_VERSION_template_haskell(2,17,0)
+    liftTyped = TH.Code . TH.unsafeTExpCoerce . TH.lift
```
phadej, line 430: TH.unsafeCodeCoerce . TH.lift

bergmark:
Thank you @galenhuntington!

Perhaps this will later be included in the th-abstraction package that we use?

Will try to get the next release out this weekend.

galenhuntington:
About that, I should mention that I had to build `th-abstraction` from https://github.com/glguy/th-abstraction/pull/83 for GHC 9, because of the (unrelated) `TyVarBndr` change in TH 2.17.  So the upper bound on `th-abstraction` may have to be changed to accommodate whatever that version winds up being when it is merged.  However, that PR does not cover this `Code` change.

PR #803, opened 2020-08-29
Replace criterion with gauge because the latter does not depend on aeson

ethercrow:
I think it would allow the aeson package to have benchmarks without jumping through as many hoops as there are now. The separate cabal.project, separate cabal file, maybe some parts of scripts will also become redundant.

`gauge` is both API-compatible and output format compatible with `criterion`, so `criterion-compare` tool seems to still work with it.

### DIFF
--- a/benchmarks/Suite.hs
+++ b/benchmarks/Suite.hs
@@ -7,7 +7,7 @@ import           Data.Aeson
 import           Prelude.Compat
 
 import           Control.DeepSeq      (NFData)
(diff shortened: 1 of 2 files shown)


phadej:
I avoid depending on packages by `gauge`s maintainer.

phadej:
More importantly, CSV output is needed to make comparisons of benchmark runs as `bench.sh` / `criterion-compare-txt` allows to compare different branches by actually calculating the differences, and not only "visually compared by human eye".

ethercrow:
I see.

PR #804, opened 2020-09-04
toEncoding & JSON keys as strings

acondolu:
Hello!

My scenario is JSON serialization with `toEncoding` (to `ByteString`) which results in JSON files
with lots of constructor labels (= lots of keys). I am using `deriveToJSON` to derive the `ToJSON` instances.

The current way of implementing `toEncoding` of objects is by supplying key-value pairs, where the key is an Haskell `String`.
Using strings when serializing to `ByteString`s seems inefficient, so I experimented with using `ByteString` for keys instead.
I only touched the Template Haskell part in such a way that, when deriving `toEncoding`, the labels of constructors are escaped at compile time, and also manually lifted at the toplevel of `toEncoding` and thus shared in subsequent calls to `toEncoding` (see dumped splices below).

I benchmarked an artificial example that builds a big binary tree and serializes it with `toEncoding`:

```haskell
data BinTree = TreeLeaf | TreeNode { leftTree :: BinTree, rightTree :: BinTree }
$(deriveToJSON defaultOptions ''BinTree)
```

The performance improvement is promising (it's more than twice as fast):

- Current `aeson`:
```
time                 1.031 s    (1.026 s .. 1.042 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.027 s    (1.026 s .. 1.029 s)
std dev              2.126 ms   (75.05 μs .. 2.567 ms)
variance introduced by outliers: 19% (moderately inflated)
```

```
70,748,400,904 bytes allocated in the heap
    62,214,640 bytes copied during GC
        158,952 bytes maximum residency (8 sample(s))
        54,320 bytes maximum slop
              0 MB total memory in use (0 MB lost due to fragmentation)
```

- This PR:
```
time                 331.7 ms   (329.8 ms .. 335.1 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 330.0 ms   (329.6 ms .. 331.2 ms)
std dev              903.6 μs   (19.72 μs .. 1.248 ms)
variance introduced by outliers: 16% (moderately inflated)
```

```
44,980,570,088 bytes allocated in the heap
    21,969,912 bytes copied during GC
        157,424 bytes maximum residency (6 sample(s))
        52,760 bytes maximum slop
              0 MB total memory in use (0 MB lost due to fragmentation)
```

(The code for this benchmark is here: https://github.com/acondolu/aeson-bintree)

This is the output of `-ddump-splices`:
```haskell
  deriveToJSON defaultOptions ''BinTree
======>
  instance ToJSON BinTree where
    toJSON = [...]
    toEncoding
      = let
          bs_TreeLeaf = pack [...] -- <-- "\"TreeLeaf\""
          bs_TreeNode = [...]
          bs_leftTree = [...]
          bs_rightTree = [...]
          bs_tag = [...]
        in
          \ value_acGC
            -> case value_acGC of
                  TreeLeaf
                    -> fromPairs
                        ((pair' (Encoding (shortByteString bs_tag)))
                            (Encoding (shortByteString bs_TreeLeaf)))
                  TreeNode arg1_acGD arg2_acGE
                    -> fromPairs
                        ((pair' (Encoding (shortByteString bs_tag)))
                            (Encoding (shortByteString bs_TreeNode))
                            <>
                              ((pair' (Encoding (shortByteString bs_leftTree)))
                                (toEncoding arg1_acGD)
                                <>
                                  (pair' (Encoding (shortByteString bs_rightTree)))
                                    (toEncoding arg2_acGE)))
```

Comparison using `aeson` benchmarks:
```
Benchmark                              master  bs-keys         
Examples/decode/github-issues/lazy    2.91e-3  2.78e-3   -4.73%
Examples/decode/github-issues/strict  3.55e-3  2.71e-3  -23.55%
Examples/decode/jp100/lazy            3.63e-3  3.47e-3   -4.43%
Examples/decode/jp100/strict          3.52e-3  3.51e-3   -0.42%
Examples/decode/twitter100/lazy       2.86e-3  2.77e-3   -3.20%
Examples/decode/twitter100/strict     2.58e-3  2.49e-3   -3.33%
zzz-geom-mean                         3.15e-3  2.93e-3   -6.96%
```

### Alternative solution
This PR only improves the performance when deriving instances with TH (unsatisfactory, for instance `genericToEncoding` is left out).

An alternative would be to leave the TH part unchanged, make `pair` to pack the escaped keys to `ByteString`s, and force GHC to do the lifting by itself, by making it realize that escaped keys are constant expressions after all. However, I was not able to do that:
maybe somebody more experienced with the GHC simplifier can help.

Let me know what you think!

### DIFF
--- a/Data/Aeson/TH.hs
+++ b/Data/Aeson/TH.hs
@@ -132,7 +132,7 @@ import qualified Data.ByteString.Short as S (pack)
 import qualified Data.ByteString.Lazy as BL (unpack)
 import Control.Monad (liftM2, unless, when)
 import Data.Foldable (foldr')
-import Data.IORef (IORef, newIORef, readIORef, modifyIORef)
+import Data.IORef (IORef, atomicModifyIORef, newIORef, readIORef)
 #if MIN_VERSION_template_haskell(2,8,0) && !MIN_VERSION_template_haskell(2,10,0)
 import Data.List (nub)
 #endif
@@ -151,11 +151,11 @@ import Text.Printf (printf)
 import qualified Data.Aeson.Encoding.Internal as E
 import qualified Data.Aeson.Encoding.Builder as EB
 import qualified Data.Foldable as F (all)
-import qualified Data.HashMap.Strict as H (difference, fromList, keys, lookup, toList)
+import qualified Data.HashMap.Strict as H (HashMap, difference, empty, fromList, insert, keys, lookup, toList)
 import qualified Data.List.NonEmpty as NE (length, reverse)
 import qualified Data.Map as M (fromList, keys, lookup , singleton, size)
 import qualified Data.Semigroup as Semigroup (Option(..))
-import qualified Data.Set as Set (empty, insert, member, toList)
+import qualified Data.Set as Set (empty, insert, member)
 import qualified Data.Text as T (Text, pack, unpack)
 import qualified Data.Vector as V (unsafeIndex, null, length, create, empty)
 import qualified Data.Vector.Mutable as VM (unsafeNew, unsafeWrite)
@@ -354,30 +354,39 @@ consToValue target jc opts instTys cons = do
               , let conName = constructorName con
               ]
         | otherwise -> [argsToValue target jc tvMap opts True con | con <- cons]
+    -- Bind the definitions of used JSON keys at top-level
     lets :: ToJSONFun -> ExpQ -> ExpQ
     lets Value e = e
     lets (Encoding r) e = do
-      e' <- e -- collect used keys
+      e' <- e -- collect used JSON keys
       labels <- runIO $ readIORef r
       let decs = map
-            (\k -> let str = BL.unpack $ B.toLazyByteString $ EB.string k
-                  in valD (varP $ varOfKey k) (normalB [| S.pack str |]) [])
-            (Set.toList labels)
+            (\(k,v) ->
+              let str = BL.unpack $ B.toLazyByteString $ EB.string k
+              in valD (varP v) (normalB [| S.pack str |]) [])
+            (H.toList labels)
       letE decs $ pure e'
 
-varOfKey :: String -> Name
-varOfKey l = mkName $ "bs_" ++ l
+-- | Map a JSON string to the name of the Haskell variable that
+-- holds that string. Used to gather all necessary JSON keys and
+-- bind them at top-level.
+-- By performing this manual floating, we avoid GHC to
+-- re-allocate the keys at each call of 'toJSON'/'toEncoding'.
+type KeyMap = IORef (H.HashMap String Name)
 
-type KeySet = IORef (Set String)
+newKeyMap :: Q KeyMap
+newKeyMap = runIO $ newIORef H.empty
 
-newKeySet :: Q KeySet
-newKeySet = runIO $ newIORef Set.empty
-
-keyRef :: KeySet -> String -> ExpQ
+-- | Name of the Haskell variable that is going to hold
+-- the supplied JSON string.
+keyRef :: KeyMap -> String -> Q Exp
 keyRef ref s = do
-  runIO $ modifyIORef ref $ Set.insert s
-  let var = varE $ varOfKey s
-  [|E.Encoding|] `appE` ([|B.shortByteString|] `appE` var)
+  vNew <- newName "key"
+  v <- runIO $ atomicModifyIORef ref $ \hm ->
+    case H.lookup s hm of
+      Nothing -> (H.insert s vNew hm, vNew)
+      Just v -> (hm, v)
+  [|E.Encoding|] `appE` ([|B.shortByteString|] `appE` varE v)
 
 -- | Name of the constructor as a quoted 'Value' or 'Encoding'.
 conStr :: ToJSONFun -> Options -> Name -> Q Exp
@@ -685,7 +694,7 @@ consFromJSON :: JSONClass
              -- ^ The FromJSON variant being derived.
              -> Name
              -- ^ Name of the type to which the constructors belong.
-             -> KeySet
+             -> KeyMap
              -> Options
              -- ^ Encoding options
              -> [Type]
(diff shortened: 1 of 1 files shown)


acondolu:
Hello,

any comment on this PR? Is this performance improvement desirable?

Let me know what I can do to move this PR in some direction.

Data/Aeson/TH.hs
```diff
@@ -349,11 +354,35 @@ consToValue target jc opts instTys cons = do
...
+      let decs = map
+            (\k -> let str = BL.unpack $ B.toLazyByteString $ EB.string k
+                  in valD (varP $ varOfKey k) (normalB [| S.pack str |]) [])
+            (Set.toList labels)
+      letE decs $ pure e'
+
+varOfKey :: String -> Name
+varOfKey l = mkName $ "bs_" ++ l
+
+type KeySet = IORef (Set String)
```
phadej, line 371: `IORef`? Why?

Data/Aeson/TH.hs
```diff
@@ -349,11 +354,35 @@ consToValue target jc opts instTys cons = do
...
+      e' <- e -- collect used keys
+      labels <- runIO $ readIORef r
+      let decs = map
+            (\k -> let str = BL.unpack $ B.toLazyByteString $ EB.string k
+                  in valD (varP $ varOfKey k) (normalB [| S.pack str |]) [])
+            (Set.toList labels)
+      letE decs $ pure e'
+
+varOfKey :: String -> Name
+varOfKey l = mkName $ "bs_" ++ l
```
phadej, line 369: Also I'm not a fan. This should be using `newName`, as it's local definition, isn't it?

phadej: Aa problem is that TH-generated definitions also cause shadowed-name etc warnings (as they should). And given that `aeson` is used "everywhere", let's not take any risks.

The `IORef` is another kind of a problem. This `TH` code is already quite difficult to follow. Let's not make it worse. The TH code itself doesn't have performance constraints, only the code we generate with it.

acondolu: Thank you, I've replaced `mkName` with `newName`.

Data/Aeson/TH.hs
acondolu: I use the `IORef` to store the JSON strings encountered during the generation of `toEncoding`. In the end, I gather the strings from the `IORef` and create the top-level bindings.

What doesn't look good about the `IORef`? If we don't want it, I guess the alternative is using a state monad style, but the refactoring seems quite invasive.

phadej: I feel uneasy.

There is a non-local state, and it's hidden. I don't know if changing `type KeyMap` to `type KeyMapIORef` and adding a note about the scheme will help.

(I'm not opposed to having `IORef` based `StateT s IO`-like monads, but then I'd like to have that monad in `newtype`, and the problem of refactoring is still there).

What others think, @bergmark @RyanGlScott

Lysxia: Do you really save much compile time with that optimization? The strings involved here are short and used only once, aren't they?

phadej: @Lysxia as you can see from benchmark in PR description, GHC is dumb, and doesn't float bytestring `Encoding`s. So this is welcome change.

Using `IORef` is *easy* way to introduce state into TH structure traversal, so we can do manual float-out. It's not a performance optimization, but a "as Q is IO, let's embrace it and make our lives easier but changing everything".
I understand that viewpoint. However I'm not sure if it makes our lives easier in long term.

Lysxia: Thanks for the clarification. I had indeed misunderstood the change.

I also found `IORef` a bit surprising but ultimately it doesn't seem worse than the alternative of `StateT KeyMap` (without `IORef`). The state is arguably quite "well-behaved" since it's only accessible when `r :: KeyMap` is in scope and all updates are done by `keyRef` where it's easy to see that no overwriting happens. It's completely contained in this file and invisible to users, so I think this situation is really far from the horror tales of global state programmers like to tell.

phadej:
I also wonder whether it makes sense to let-float the `shortByteString (pack ...)`, not only `pack`. Whether there is difference.

If I were to write something like that by hand, I'd float-out complete `Encoding 'Text` values for keys, i.e.

```
toEncoding =
  let keyA = Encoding.string "a" -- and these could be pre-calced a lot.
      keyB = Encoding.string "b" --
  in \v -> case v of
     ... -> fromPairs (pair' keyA ... <> pair' keyB ...)
```

In fact, the *in theory*, if we construct
https://hackage.haskell.org/package/bytestring-0.11.0.0/docs/Data-ByteString-Builder-Prim.html#t:FixedPrim
(as we know the length!), encoding should go as fast as it could:
one length check for whole key, and then just writing the key data into the
buffer. Nothing for GHC really to optimize there.

---

Could you experiment whether that is viable, and have an effect.
Reducing how much stuff GHC optimizer have to consider has effect
on compilation time (especially if people go crazy and use `{-# INLINE #-}`
everywhere).

If we have `FixedPrim` stuff, it would be up to GHC to decide whether they are cheap to inline, or keep floated-out.
And we should let GHC do that. Inlining is cheaper optimization to perform than CSE.

---

I can then think how to get rid of `IORef` in elegant way.
(I know how, but it's indeed not a one-line patch, though quite
straigh-forward anyway).

Data/Aeson/TH.hs
acondolu: I renamed `KeyMap` to `KeyMapIORef`.

acondolu:
Hi,
I've applied the change suggested above (floating the `Encoding`s instead of the `ByteString`s). I've carried again the benchmark and the performance gain remains, so GHC is still happy. 🎉

About the `FixedPrim` suggestion: I think what `shortByteString :: ShortByteString -> Builder` does is basically what mentioned by @phadej in the comment above? i.e. checking the limits of the buffer and then copying the underlying byte array (see https://hackage.haskell.org/package/bytestring-0.11.0.0/docs/src/Data.ByteString.Builder.Internal.html#shortByteString). (The important thing is ensuring that the `ShortByteString` is computed only once, and not at every call of `toEncoding`. Once it's computed, I agree that there's nothing to optimise and it's as fast as possible).

phadej:
I implemented this idea in https://github.com/haskell/aeson/pull/906

PR #805, opened 2020-09-12
Use Text.Encoding.decodeLatin1 for decoding ASCII strings

ethercrow:
According to the original decodeLatin1 PR https://github.com/haskell/text/pull/18 it's significantly faster than the currently used decodeUtf8 at decoding ASCII.

```
Benchmark                             baseline2  treatment2
Examples/decode/github-issues/lazy      2.35e-3     2.15e-3  -8.63%
Examples/decode/github-issues/strict    2.08e-3     2.04e-3  -2.18%
Examples/decode/jp100/lazy              2.51e-3     2.32e-3  -7.57%
Examples/decode/jp100/strict            2.40e-3     2.39e-3  -0.51%
Examples/decode/twitter100/lazy         1.76e-3     1.65e-3  -6.65%
Examples/decode/twitter100/strict       1.75e-3     1.63e-3  -6.60%
zzz-geom-mean                           2.12e-3     2.00e-3  -5.40%
```

### DIFF
--- a/Data/Aeson/Parser/Internal.hs
+++ b/Data/Aeson/Parser/Internal.hs
@@ -322,13 +322,19 @@ jstring_ = do
   -- not sure whether >= or bit hackery is faster
   -- perfectly, we shouldn't care, it's compiler job.
   s <- A.takeWhile (\w -> w /= DOUBLE_QUOTE && w /= BACKSLASH && w >= 0x20 && w < 0x80)
-  let txt = TE.decodeLatin1 s
+  let txt = unsafeDecodeASCII s
   mw <- A.peekWord8
   case mw of
     Nothing           -> fail "string without end"
     Just DOUBLE_QUOTE -> A.anyWord8 $> txt
     Just w | w < 0x20 -> fail "unescaped control character"
     _                 -> jstringSlow s
+
+-- | The input is assumed to contain only 7bit ASCII characters (i.e. @< 0x80@).
+--   We use TE.decodeLatin1 here because TE.decodeASCII is currently (text-1.2.4.0)
+--   deprecated and equal to TE.decodeUtf8, which is slower than TE.decodeLatin1.
+unsafeDecodeASCII :: B.ByteString -> Text
+unsafeDecodeASCII = TE.decodeLatin1
 
 jstringSlow :: B.ByteString -> Parser Text
 {-# INLINE jstringSlow #-}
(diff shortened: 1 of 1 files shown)


ethercrow:
Looks like this was already considered in https://github.com/bos/aeson/pull/753#discussion_r408955626

ethercrow:
@phadej what do you think?

Data/Aeson/Parser/Internal.hs
```diff
@@ -322,7 +322,7 @@ jstring_ = do
   -- not sure whether >= or bit hackery is faster
   -- perfectly, we shouldn't care, it's compiler job.
   s <- A.takeWhile (\w -> w /= DOUBLE_QUOTE && w /= BACKSLASH && w >= 0x20 && w < 0x80)
-  let txt = TE.decodeUtf8 s
+  let txt = TE.decodeLatin1 s
```
phadej, line 325: could you make a local definition, and name it `unsafeDecodeAscii` with a comment that

```
-- | The input is assumed to contain only 7bit ASCII characters (i.e. @< 0x80@).
unsafeDecodeAscii :: ByteString -> Text
```

That name will communicate better what we try to do. In fact, I think we can win more with custom FFI (`+cffi`) i.e. do what `decodeLatin1` does, but with even less checks (as we have checked that there is no `0x80` bytes).

Perfectly, `unsafeDecodeAscii` would be in `text`.

phadej (changes requested):
This is good. Lets add a bit of internal documentation though.

phadej (approved):

Data/Aeson/Parser/Internal.hs
ethercrow: I took a stab at accelerating `decodeLatin1` https://github.com/haskell/text/pull/297

ethercrow:
@phadej @bergmark Could you clarify what's the process, if any exists, of aeson releases and PR merging? Is this PR waiting for something or is just everyone super busy these days?

bergmark:
@ethercrow it's blocked on me being busy, unfortunately. I'll try to get a release out this weekend. Hoping to move this repo to the haskell org soon to raise the bus factor.

bergmark:
and thanks for the patch!

ethercrow:
@bergmark thank you for the explanation, take care!

PR #809, opened 2020-09-30
Allow th-abstraction-0.4

phadej:
This is draft, as `bifunctors` (atm) prevent `th-abstraction-0.4` from entering the plan.


phadej:
`bifunctors-5.5.8` is on Hackage, so it should be picked up (and `th-abstraction-0.4`) by CI. Lets see.

bergmark:
yep travis log LG, thanks!

phadej:
@bergmark thanks for merging!

Could you make a release soon-ish, as there are packages bumping lower bound of `th-abstraction` to `>=0.4`, as it contains helpers to have CPP (for GHC-9.0 support) only in one place (which luckily is not the case with `aeson`)

bergmark:
done!

PR #817, opened 2020-12-08
Change urls in .cabal files

phadej:


Lysxia:
Just out of curiosity, when did that transfer of aeson to the haskell account happen?

phadej:
@Lysxia based on the date of this PR, around ten days ago. 

The packages were moved to haskell organization, so e.g. adding maintainers or enabling external tools 
 (e.g. CI, which was the case for e.g .`attoparsec`) wouldn't be blocked on Bryan.
 
I however assume that there is nothing else implied. E.g. the maintainers of packages are still their current maintainers,
and e.g. https://github.com/haskell/text-format is unmaintained, so, I guess, free for grabs.
 
Thanks to @bgamari by poking people to make this happen.

PR #822, opened 2021-01-11
Use GitHub Actions

phadej:


phadej:
Travis has served us well. Hopefully GHA will too.

PR #828, opened 2021-01-18
Add sum encoding TaggedFlatObject

poscat0x04:
WIP, currently only supports generic deriving.


poscat0x04:
I'm having serious troubles writing Template Haskell without splicing :(

Data/Aeson/Types/FromJSON.hs
```diff
@@ -10,6 +10,7 @@
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
```
phadej, line 13: Cannot be used. We support pre GHC-8.0 compilers.

Data/Aeson/Types/FromJSON.hs
```diff
@@ -158,6 +159,8 @@ import qualified Data.Primitive.Types as PM
 import qualified Data.Primitive.PrimArray as PM
 
 import Data.Coerce (Coercible, coerce)
+import GHC.TypeNats
+import Data.Kind (Type)
```
phadej, line 163: ditto

poscat0x04:
It should now work with GHC 7.*

poscat0x04:
For some reason, the `INCOHERENT` pragma did not work for ghc-7.8.4.

poscat0x04:
@phadej Is this ok now?

Data/Aeson/TH.hs
```diff
@@ -421,6 +421,7 @@ sumToValue target opts multiCons nullary conName value pairs
                 content = pairs contentsFieldName
             in fromPairsE $
               if nullary then tag else infixApp tag [|(Monoid.<>)|] content
+          TaggedFlatObject {} -> error "impossible"
```
bergmark, line 424: Is there a bug in aeson if this triggers?

Data/Aeson/Types/Internal.hs
```diff
@@ -682,6 +682,8 @@ data SumEncoding =
     -- by the encoded value of that field! If the constructor is not a
     -- record the encoded constructor contents will be stored under
     -- the 'contentsFieldName' field.
+  | TaggedFlatObject { tagFieldName :: String }
```
bergmark, line 685: What happens if the tagFieldName is also a record field?

Data/Aeson/Types/Internal.hs
```diff
@@ -682,6 +682,8 @@ data SumEncoding =
...
+    -- ^ A constructor will be encoded to an object with a field
```
bergmark, line 686: Please clarify this, it's not clear to me what "an object with a field" means

bergmark (changes requested):
> Encode single constructor data types with field names as a plain record (This is the same as TaggedObject)

"single constructor data types with field names" = record?
No `$tagName` here, or what does "plain record" mean? If there's precedent for this from TaggedObject then perhaps we should keep this, but why does TaggedObject work like that? :-S


> Encode multi constructor data types without fields names as a record whose keys are the positions of the fields(starting from 1) and values are the fields encoded as JSON with an additional KV pair "$tagFieldName": "$constructorName". For example A 1 2 will be encoded as {"tag": "A", "1": 1, "2": "2" }.

I think starting at 0 would make more sense as it matches the indices when we use arrays. I would also make the indices ints so that this is closer to an array. Or is there precedent for this elsewhere that I missed?

> Note that the tag will overwrite any fields that have the same name.

Can we give an error if this occurs instead?


I think there should be some regression tests on all the functionality. We have some in NullaryConstructors.hs now but It's hard to tell how this will behave with other shapes of data types.

Data/Aeson/Types/Internal.hs
bergmark: Saw this answered elsewhere - This should be documented here I think.

poscat0x04:
>No $tagName here, or what does "plain record" mean?

Plain record just means a record without any additional field.

> I think starting at 0 would make more sense as it matches the indices when we use arrays.

Yeah, that makes sense. TBH the current choice of encoding constructors without field names is a bit arbitrary. I initially did not plan to support them but then figured out it would affect other `SumEncoding` variants because we currently cannot distinguish different `SumEncodings` at the type level ([relevant code](https://github.com/haskell/aeson/blob/8f92dbf212a35c80855f33d87456edf0559d17bd/src/Data/Aeson/Types/FromJSON.hs#L1105-L1109)).

> I would also make the indices ints so that this is closer to an array.

I'm not sure if I understand. JSON keys must be strings, right?

> Can we give an error if this occurs instead, at least in TH? Or will it be more confusing if TH/Generics differ here?

Yeah, I think they should have the same behavior.

bergmark:
> I'm not sure if I understand. JSON keys must be strings, right?

Guess who has been doing too much JavaScript lately *runs away*, please disregard

bergmark (approved):
Nice :-)

I actually need this now for a rust API I'm testing!! Should be usable with `#[serde(tag = ...)]`

bergmark:
I managed to hit some github hot key that posted my review for me... @phadej do you want to look over this again?

bergmark:
Eh it looks like the tests don't compile, but only on GHC 7.8 :-/

phadej:
The issue test-failure issue might be something which will get fixed with #839. I think I can wipe out all use over overlapping instances for good. Just need to get to it. (And get courage to commit a breaking change).

bergmark:
You can do it!

lf-:
Hi! What's the status on this PR? I would like to delete some of my manually written instances :)

tadfisher:
What's the status of this PR? It would be a useful feature for our codebase, FWIW.

phadej:
My current opinion is to throw `Options` into `/dev/null` and rather provide a bunch of newtypes to be used with deriving via, for different encoding schemes.

Current generics and template haskell code is very much unmaintainable (as it tries to do evertything), so this PR is probably not going anywhere anytime soon.

PR #830, opened 2021-02-14
Subsrc dirs

phadej:


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Move code into subdirectories
+6230e4ee376ce01ab05b8e48a3a7484f8d602b9e
```
bergmark, line 2: TIL git 🤩

phadej: It's relatively new addition https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame

I'm not sure whether this file is built in or still need to be configured. But yes, one reason to avoid moving stuff around is removed.

PR #838, opened 2021-02-18
Fix #837: Stable (in key order) Show Value instance

### LINKED ISSUE #837: hashable 1.3.1.0 changes aesons property ordering(?)
At least I think this is the case, as a result of https://github.com/haskell-unordered-containers/hashable/pull/196 so sorry if this turns out to be incorrect. Thought i'd file this now in case we want to do something about it quickly before a lot of downstream libs have been updated.

aeson doesn't guarantee ordering of fields, which makes sense. But it is unfortunate if this change to hashable breaks test suites, esp. if it's because of a dependency version changing. Many people are matching aeson output against strings. It could affect "actual" code as well I suppose.

I'm not sure what the best way forward is

1. restrict aeson to an older version of hashable until we make a major release
2. make the latest hashable release unbuildable and publish a major version for it instead.
3. do nothing.

We should probably recommend use of Data.Aeson.QQ(.Simple) in test suites either way. It would have prevented the noise.

@juhp do you have examples of packages with test failures?

phadej:
- Add test that read . show roundtrips
- Add test that Value To/FromJSON instances roundtrip (as we got
  Arbitrary Value instance in tests)
- Cleanup some doctest looking comments

### DIFF
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -1,5 +1,5 @@
 name:            aeson
-version:         1.5.5.1
+version:         1.5.6.0
 license:         BSD3
 license-file:    LICENSE
 category:        Text, Web, JSON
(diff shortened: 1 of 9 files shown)


bergmark (approved):
i assumed soon-ish was weeks not minutes :D

PR #855, opened 2021-06-21
make sure language is detected as haskell

silky:
:)


bergmark:
Thanks! I laughed at this the other day

silky:
@bergmark A great joke from one of my colleagues was that an alternative would be to change the tagline to "A swift Haskell JSON library" 😀

PR #859, opened 2021-07-31
Support decimal output for rational

newhoggy:
The `FromJSON` instance is already capable of handling the decimal representation.  This change supports outputting the decimal representation when that is more compact.

### DIFF
--- a/src/Data/Aeson/Types/ToJSON.hs
+++ b/src/Data/Aeson/Types/ToJSON.hs
@@ -82,7 +82,7 @@ import Data.List (intersperse)
 import Data.List.NonEmpty (NonEmpty(..))
 import Data.Proxy (Proxy(..))
 import Data.Ratio (Ratio, denominator, numerator)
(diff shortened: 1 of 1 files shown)


newhoggy:
```
λ> toJSON (1 % 10)
Number 0.1
λ> toJSON (1 % 3)
Object (fromList [("denominator",Number 3.0),("numerator",Number 1.0)])
```

dcoutts (commented):
LGTM.

https://hackage.haskell.org/package/scientific-0.3.7.0/docs/Data-Scientific.html#v:fromRationalRepetend

I agree that matching on `Right (s, Nothing)` looks like the right choice.

bergmark:
Thanks!

I think I need some convincing for this. It is a breaking change for anyone using this instance. The reason to only adapt FromJSON may have been because we do not consider that breaking.

I would lean towards having this in a newtype.

newhoggy:
What would you choose as the name of the newtype?

PR #861, opened 2021-08-08
Fix FromJSON1 instance of Sum

### LINKED ISSUE #860: Very minor oopsie in FromJSON1 instance of Sum
In the `InR` case, the `Key` is still `inl`.

```
instance (FromJSON1 f, FromJSON1 g) => FromJSON1 (Sum f g) where
    liftParseJSON p pl (Object (H.toList -> [(key, value)]))
        | key == inl = InL <$> liftParseJSON p pl value <?> Key inl
        | key == inr = InR <$> liftParseJSON p pl value <?> Key inl
...
```

Patil2099:
Fixes  #860

### DIFF
--- a/src/Data/Aeson/Types/FromJSON.hs
+++ b/src/Data/Aeson/Types/FromJSON.hs
@@ -1801,7 +1801,7 @@ instance (FromJSON1 f, FromJSON1 g, FromJSON a) => FromJSON (Product f g a) wher
 instance (FromJSON1 f, FromJSON1 g) => FromJSON1 (Sum f g) where
     liftParseJSON p pl (Object (H.toList -> [(key, value)]))
         | key == inl = InL <$> liftParseJSON p pl value <?> Key inl
-        | key == inr = InR <$> liftParseJSON p pl value <?> Key inl
+        | key == inr = InR <$> liftParseJSON p pl value <?> Key inr
       where
         inl, inr :: Text
         inl = "InL"


phadej (changes requested):
Please add test(s) to SerializationFormatSpec.hs

Vlix:
The `... <?> Key x` part of the `Parser` doesn't influence serialization, right?
So the test should be a failed parsing, and make sure the error message is correct?

bergmark:
> The `... <?> Key x` part of the `Parser` doesn't influence serialization, right?
> So the test should be a failed parsing, and make sure the error message is correct?

Right. I think the right place for the test is the golden tests in tests/ErrorMessages.hs (`simple` / `output`)

src/Data/Aeson/Types/FromJSON.hs
```diff
@@ -1801,7 +1801,7 @@ instance (FromJSON1 f, FromJSON1 g, FromJSON a) => FromJSON (Product f g a) wher
 instance (FromJSON1 f, FromJSON1 g) => FromJSON1 (Sum f g) where
     liftParseJSON p pl (Object (H.toList -> [(key, value)]))
         | key == inl = InL <$> liftParseJSON p pl value <?> Key inl
-        | key == inr = InR <$> liftParseJSON p pl value <?> Key inl
+        | key == inr = InR <$> liftParseJSON p pl value <?> Key inr
```
taasan, line 1804: Wouldn't something like this be less error prone? (Change `f` and `g` to more meaningful names)

```haskell
    liftParseJSON p pl (Object (H.toList -> [(key, value)]))
        | key == inl = f InL
        | key == inr = f InR
      where
        f g = g <$> liftParseJSON p pl value <?> Key key
```

phadej:
I'm sorry. I completely misunderstood the PR. if it's fixing the `Key` only we don't necessarily need a test.

PR #862, opened 2021-08-30
Allow building with base-compat-batteries-0.12.*

RyanGlScott:


phadej:
I published the revisions

PR #865, opened 2021-09-13
Make Object use an opaque Textmap interface

Boarders:


Boarders:
I still didn't add the `Lifted` instance as I do not know Template Haskell well enough to know what to do. Please let me know what else needs to change? I can add documentation for TextMap.

src/Data/Aeson/Types/ToJSON.hs
```diff
@@ -1737,17 +1738,39 @@ instance (ToJSON a) => ToJSON (HashSet.HashSet a) where
...
 instance (ToJSON v, ToJSONKey k) => ToJSON (H.HashMap k v) where
+    toJSON = toJSON1
+    toEncoding = toEncoding1
+
+-------------------------------------------------------------------------------
+-- Data.Aeson.TextMap
+-------------------------------------------------------------------------------
+
+instance ToJSON1 TM.TextMap where
+    liftToJSON g _ = case toJSONKey of
```
phadej, line 1761: this doesn't look right. You don't need to look at `toJSONKey` (of `Text`), you can directly do `Object . TM.map g`, I think.

Boarders: Cool, changed it.

src/Data/Aeson/TextMap.hs
```diff
@@ -0,0 +1,102 @@
...
+import Prelude hiding (lookup)
+import Data.Bifunctor (first)
+
+
+import qualified Data.HashMap.Strict as H
+import qualified Language.Haskell.TH.Syntax as TH
+
+newtype TextMap v = TextMap { unTextMap :: HashMap Text v }
+  deriving (Eq, Ord, Typeable, Data)
+  deriving newtype (Read, Show, Functor, Hashable, NFData)
```
phadej, line 44: `DerivingStrategies` is GHC-8.2 only. We should write these instance by hand for now. (stock deriving for `Functor` is fine. We need `Foldable` and `Traversable` too, latter cannot be `newtype` derivied anyway: https://ryanglscott.github.io/2018/06/22/quantifiedconstraints-and-the-trouble-with-traversable/).

Boarders:
Do you think there is any value in also making this abstract in terms of the `Text` type so that could be switched out or save that for later?

phadej:
Let's do with concrete `Text` for now. One step at the time.

phadej:
FWIW, feel free to squash my commit once things starts to look ok.

src/Data/Aeson/TextMap.hs
```diff
@@ -0,0 +1,102 @@
...
+
+instance TH.Lift v => TH.Lift (TextMap v) where
+  lift (TextMap m) = [| TextMap (H.fromList . map (first pack) $ m') |]
+      where m' = map (first unpack) . H.toList $ m
```
phadej, line 48: add `liftTyped = error "TODO"`. I'll fix it afterwards.

Let's make GHC-9.0 & 8.10 jobs happy, I'll take care of the rest + benchmark jobs.

phadej: ... or you can look at `Lift Value`, and do the same: https://hackage.haskell.org/package/aeson-1.5.6.0/docs/src/Data.Aeson.Types.Internal.html#line-440

Boarders:
Apologies if this led to a bunch of CI spam.

src/Data/Aeson/Internal/Functions.hs
```diff
@@ -11,22 +11,31 @@
...
 import qualified Data.Map as M
+import qualified Data.Text as T
 
 -- | Transform a 'M.Map' into a 'H.HashMap' while transforming the keys.
 mapHashKeyVal :: (Eq k2, Hashable k2) => (k1 -> k2) -> (v1 -> v2)
               -> M.Map k1 v1 -> H.HashMap k2 v2
 mapHashKeyVal fk kv = M.foldrWithKey (\k v -> H.insert (fk k) (kv v)) H.empty
 {-# INLINE mapHashKeyVal #-}
 
+-- | Transform a 'M.Map' into a 'H.HashMap' while transforming the keys.
```
masaeedu, line 33: Probably missing something but figured it can't hurt to ask. Should this maybe be:

```suggestion
-- | Transform a 'M.Map' into a 'TM.TextMap' while transforming the keys.
```

?

Boarders: Thanks!

src/Data/Aeson/TH.hs
```diff
@@ -147,7 +148,7 @@ import Language.Haskell.TH.Syntax (mkNameG_tc)
 import Text.Printf (printf)
 import qualified Data.Aeson.Encoding.Internal as E
 import qualified Data.Foldable as F (all)
-import qualified Data.HashMap.Strict as H (difference, fromList, keys, lookup, toList)
+--import qualified Data.HashMap.Strict as H (difference, fromList, keys, toList)
```
phadej, line 151: delete

src/Data/Aeson/TextMap.hs
```diff
@@ -0,0 +1,210 @@
...
+    readPrec = parens $ prec 10 $ do
+      Ident "fromList" <- lexP
+      xs <- readPrec
+      return (fromList xs)
+
+    readListPrec = readListPrecDefault
+
+instance Show v => Show (TextMap v) where
+    showsPrec d m = showParen (d > 10) $
+      showString "fromList " . shows (toList m)
```
phadej, line 68: `toAscList` (to avoid non-detereminism of hashmap)

src/Data/Aeson/TextMap.hs
```diff
@@ -0,0 +1,210 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+-- |
+-- An abstract interface for maps from Textual keys to values.
+
+module Data.Aeson.TextMap (
+    TextMap,
```
phadej, line 10: I see you wrote some haddocks, great!

It would be great to have sections like in https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Map-Strict.html: Map type, construction, insertion, etc.

src/Data/Aeson/Types/ToJSON.hs
```diff
@@ -1737,17 +1738,36 @@ instance (ToJSON a) => ToJSON (HashSet.HashSet a) where
...
+
+-------------------------------------------------------------------------------
+-- Data.Aeson.TextMap
+-------------------------------------------------------------------------------
+
+instance ToJSON1 TM.TextMap where
+    liftToJSON g _ = Object . fmap g
+
+    -- liftToEncoding :: forall a. (a -> Encoding) -> ([a] -> Encoding) -> TM.HashMap k a -> Encoding
+    liftToEncoding g _ = case toJSONKey of
```
phadej, line 1764: the same issue as previously with `liftToJSON`.

phadej (commented):
Starting to look great!

Boarders:
@phadej any ideas what to do with the panic on GHC 8.0.2 - I don't have access to a local copy and failed to quickly build it.

phadej:
ignore it for now

phadej:
Ah, `Parser.Internal` already has

```haskell
#if __GLASGOW_HASKELL__ <= 710 && __GLASGOW_HASKELL__ >= 706
-- Work around a compiler bug
{-# OPTIONS_GHC -fsimpl-tick-factor=300 #-}
#endif
```

bump that to 400, let's see if that's enough :)

src/Data/Aeson/Types/ToJSON.hs
Boarders: You'll have to let me know if what I wrote is the expected behaviour.

src/Data/Aeson/Parser/Internal.hs
```diff
@@ -4,7 +4,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 #if __GLASGOW_HASKELL__ <= 710 && __GLASGOW_HASKELL__ >= 706
 -- Work around a compiler bug
-{-# OPTIONS_GHC -fsimpl-tick-factor=300 #-}
+{-# OPTIONS_GHC -fsimpl-tick-factor=400 #-}
```
phadej, line 7: It looks like `300` is enough, but we need to include GHC-8.0.2, i.e. change `__GLASGOW_HASKELL__ <= 710` to `__GLASGOW_HASKELL__ <= 800`.

src/Data/Aeson/TextMap.hs
```diff
@@ -0,0 +1,230 @@
...
+difference :: TextMap v -> TextMap v' -> TextMap v
+difference tm1 tm2 = TextMap (H.difference (unTextMap tm1) (unTextMap tm2))
+
+-- | Return a list of this map's keys.
+keys :: TextMap v -> [Text]
+keys = H.keys . unTextMap
+
+-- | Convert a 'TextMap' to a 'HashMap'.
+toHashMap :: TextMap v -> HashMap Text v
+toHashMap = unTextMap
```
SamuelSchlesinger, line 218: Why no inline here?

Boarders: I didn't add inline anywhere, I think, in general, it is best practice to add inline when a benchmark actually proves it is worthwhile and not before, but I am open to changing it if others agree.

phadej: I'm thinking how get rid (or at least discourage the usage) of this function. We don't want to have `HashMap` intermediates, if they are not necessary i.e. we are literally parsing into `HashMap` / serializing from `HashMap`...

Boarders: Let me investigate if these functions are really needed. For the instances for HashMap itself we can probably use a fold which will be more representative of an abstract interface.

phadej: They are fine for now.

phadej: To clarify: we would like to avoid rebuilding a *Map when the representation is the same. nice to have.

Boarders: There might be some way to not expose it as part of the interface but add a re-write rule for this representation or something like that.

PR #873, opened 2021-10-09
Change +inf/-inf :: Double/Float serialization

phadej:


guibou:
Where can I find rational / discussion about this change?

phadej:
https://github.com/haskell/aeson/issues/679

PR #894, opened 2021-11-19
Support text-2.0

### LINKED ISSUE #893: Support text-2.0
The biggest obstacle is 

https://github.com/haskell/aeson/blob/07cea5de4abbc53accb3e7be0d14a5a5596115f5/src/Data/Aeson/Parser/Unescape.hs#L2-L11

Both `Data.Aeson.Parser.UnescapeFFI` and `Data.Aeson.Parser.UnescapePure` are fairly low-level modules, which AFAIU in one go convert UTF8 buffer to UTF16 and replace `\uXXXX` sequences by correspondent code points. I can migrate them, but essentially it will be a complete rewrite.

Does it have to be that difficult? Could `aeson` first `decodeUtf8` and then replace `\uXXXX` sequences in a separate pass?

Bodigrim:
Closes #893. 

As suggested in `Data.Aeson.Parser.UnescapePure`, CC @eskimor @jprider63.

Tested with 

```cabal
with-compiler: ghc
packages: .
packages: attoparsec-iso8601
packages: examples
tests: true
-- constraints: aeson +cffi

packages: https://hackage.haskell.org/package/text-2.0/candidate/text-2.0.tar.gz

allow-newer:
  quickcheck-instances:text,
  uuid-types:text,
  text-short:text,
  strict:text,
  scientific:text

source-repository-package
  type: git
  location: https://github.com/haskell/attoparsec
```

### DIFF
--- a/attoparsec-iso8601/attoparsec-iso8601.cabal
+++ b/attoparsec-iso8601/attoparsec-iso8601.cabal
@@ -40,7 +40,7 @@ library
     base >= 4.9 && < 5,
     base-compat-batteries >= 0.10.0 && < 0.13,
     time-compat >= 1.9.4 && < 1.10,
-    text >= 1.2.3.0 && < 1.3.0.0,
+    text >= 1.2.3.0 && < 1.3.0.0 || >= 2.0 && <2.1,
     time >= 1.6.0.1 && < 1.13
 
   if flag(fast)
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -103,7 +103,7 @@ library
     , deepseq           >=1.4.2.0  && <1.5
     , ghc-prim          >=0.5.0.0  && <0.9
     , template-haskell  >=2.11.0.0 && <2.19
-    , text              >=1.2.3.0  && <1.3
+    , text              >=1.2.3.0  && <1.3 || >=2.0 && <2.1
     , time              >=1.6.0.1  && <1.13
 
   -- Compat
@@ -146,6 +146,7 @@ library
     cpp-options:    -DCFFI
     hs-source-dirs: src-ffi
     other-modules:  Data.Aeson.Parser.UnescapeFFI
+    build-depends:  text < 2.0
 
   if flag(ordered-keymap)
     cpp-options: -DUSE_ORDEREDMAP=1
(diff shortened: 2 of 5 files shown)


phadej:
@Bodigrim How you make @bgamari merge stuff to `attoparsec`? He didn't notice my https://github.com/haskell/attoparsec/pull/192 nor https://github.com/haskell/attoparsec/issues/193

Bodigrim:
Rebased and split into comments for easier reviewing.

attoparsec-iso8601/attoparsec-iso8601.cabal
```diff
@@ -40,7 +40,7 @@ library
     base >= 4.9 && < 5,
     base-compat-batteries >= 0.10.0 && < 0.13,
     time-compat >= 1.9.4 && < 1.10,
-    text >= 1.2.3.0 && < 1.3.0.0,
+    text >= 1.2.3.0 && < 1.3.0.0 || >= 2.0 && <2.1,
```
Lysxia, line 43: Why not `>= 1.2.3.0 && < 2.1`?

Bodigrim: I don't think there will be any `text-1.3` releases, but just to be on the safe side.

Bodigrim:
`decodeASCII` has been undeprecated in https://github.com/haskell/text/pull/387.

Bodigrim:
@phadej how does it look?

phadej:
This looks good, but not perfect. I'm looking at an alternative implementation for `UnescapePure`, but this is still valuable to test new thing for correctness. My next "work on OSS" slot is next weekend though, hopefully then, no promises though.

jprider63:
I don't really have this cached in. Could you explain the changes in `text-2.0`? It seems a bit odd that the state machine needs to change, especially for the `text-1.*` version.

This is personal preference, but I'd prefer to keep  the`write` function since it'd be useful for eventually adding support for Liquid Haskell to ensure there are no out of bound writes. I may actually have time to work on this later this month.

phadej:
> Could you explain the changes in text-2.0?

Internal encoding is changed to UTF8

> It seems a bit odd that the state machine needs to change

You cannot write the hi surrogate immediately, but we need to save it as we need to know whole codepoint to encode as utf8.

phadej:
Thanks. I merged commits from this PR in two other ones. In I https://github.com/haskell/aeson/pull/901 also made benchmarks compile (you need to sdist `aeson` and put it somewhere so `criterion` could compile).

I also noticed that `buffer-builder` is not compatible with `text` (and doesn't have bounds).

Bodigrim:
Thanks @phadej this is greatly appreciated.
I filed https://github.com/chadaustin/buffer-builder/pull/19

PR #902, opened 2021-12-03
Correct Haddock comment for `fromMap`

vaibhavsagar:


phadej (approved):
:facepalm: Thanks for spotting & fixing.

Issue #911, opened 2021-12-11
cannot build with bytestring-0.11.2.0: Ambiguous occurrence ‘scientific’


jwaldmann:
I was trying to use bytestring-0.11.2.0 but this gives
```
...
aeson             > /tmp/stack-4c186e59af6847f0/aeson-2.0.2.0/src/Data/Aeson/Encoding/Builder.hs:69:30: error:
aeson             >     Ambiguous occurrence ‘scientific’
aeson             >     It could refer to
aeson             >        either ‘B.scientific’,
aeson             >               imported from ‘Data.ByteString.Builder’ at src/Data/Aeson/Encoding/Builder.hs:47:1-35
aeson             >               (and originally defined in ‘Data.ByteString.Builder.RealFloat’)
aeson             >            or ‘Data.Aeson.Encoding.Builder.scientific’,
aeson             >               defined at src/Data/Aeson/Encoding/Builder.hs:143:1
aeson             >    |
aeson             > 69 | encodeToBuilder (Number n) = scientific n
aeson             >    |                              ^^^^^^^^^^
```
because `scientific` is a fresh export https://hackage.haskell.org/package/bytestring-0.11.2.0/docs/Data-ByteString-Builder-RealFloat.html#v:scientific

phadej:
oops

jwaldmann:
simply this?
```
-import Data.ByteString.Builder as B
+import Data.ByteString.Builder as B hiding (scientific)
```

phadej:
@jwaldmann No. Explicit import list or qualified import.

phadej:
I made revisions to `aeson-2` releases. Will make revisions for older `aeson`s later today.

jwaldmann:
Still, we'd want to use the new bytestring together with aeson? In my current project, I need aeson, but its performance isn't critical.  I also need bytestring, and it *is* critical. So I want to pick up their improvements. This would be blocked if aeson restricts bytestring to an earlier version.

phadej:
resolved in 2.0.3.0 release

coderfromhere:
Could you backport it to the 1.5 branch too? Stack from Nix is now blocked on it, and I have to rebuild it from a fork - https://github.com/coderfromhere/aeson/pull/1 until the most of the ecosystem moves to v2.

PR #913, opened 2021-12-13
introduce qualified import, for #911

jwaldmann:
I very naively made the import qualified, and replaced Builder with B.Builder everywhere in this module.


phadej:
Thanks, I fixed it in https://github.com/haskell/aeson/pull/912 differently. And also went through other unqualified imports, so we only import `Prelude.Compat`, `GHC.Generics` and template-haskell stuff unqualified.

PR #914, opened 2021-12-13
Adds 'encodeStrict'

dmjio:
Added for convenience.

### DIFF
--- a/src/Data/Aeson.hs
+++ b/src/Data/Aeson.hs
@@ -49,6 +49,7 @@ module Data.Aeson
     , eitherDecodeFileStrict
     , eitherDecodeStrict'
     , eitherDecodeFileStrict'
+    , encodeStrict
     -- * Core JSON types
     , Value(..)
     , Encoding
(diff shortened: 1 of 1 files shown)


src/Data/Aeson.hs
```diff
@@ -162,6 +163,12 @@ import qualified Data.ByteString.Lazy as L
 encode :: (ToJSON a) => a -> L.ByteString
 encode = encodingToLazyByteString . toEncoding
 
+-- | Efficiently serialize a JSON value as a strict 'B.ByteString'.
```
phadej, line 166: I don't like this. This is not *efficiently*. This is a helper, which perfectly shouldn't be needed, e.g. if you serialise web api response, `wai` takes lazy bytestring.

dmjio: I can change the comment to 

```haskell
-- | Serialize a JSON value as a strict 'B.ByteString'.
```

My use case is not with `wai`, but constructing HTTP client requests for use in tests.

dmjio: Also, convenience functions are very common in libraries and ease the developer burden. `forM_ = flip mapM_`, for example.

dmjio: Adds symmetry to the API as well, `decode`, `decodeStrict` now `encode`, `encodeStrict`

PR #916, opened 2021-12-31
Add QuickCheck instances

phadej:


src/Data/Aeson/Types/Internal.hs
```diff
@@ -388,6 +391,92 @@ instance Show Value where
...
+-- | @since 2.0.2.0
+instance QC.Arbitrary Value where
+    arbitrary = QC.sized arbValue
+
+    shrink Null       = []
+    shrink (Bool b)   = Null : map Bool (QC.shrink b)
+    shrink (String x) = Null : map (String . T.pack) (QC.shrink (T.unpack x))
+    shrink (Number x) = Null : map Number (shrScientific x)
+    shrink (Array x)  = V.toList x ++ map (Array . V.fromList) (QC.shrink (V.toList x))
+    shrink (Object x) = KM.elems x ++ map (Object . KM.fromList) (QC.shrink (KM.toList x))
```
bergmark, line 403: why shrink some values to null?

phadej: Why not? Maybe it's irrelevant whether it's `Bool True` or `Null` for whatever failure, and `Array [Null, Null, Null]` is simpler then `Array[Bool True, String "", Scientific 0]`, if for example failure case requires array of three any values.

bergmark: sorry for the short comment yesterday. i'm thinking that eg `Maybe [a]` is pretty common to eg introduce a new field or for patch updates. it may shrink nicely.

phadej: @bergmark I don't understand what you mean?

bergmark (approved):

src/Data/Aeson/Types/Internal.hs
bergmark: am i reading the code right, that atomic values shrink to null but objects and arrays never do? if Bool True and Null are interchangeable, why not shrink Array [Null] to Null as well?

phadej: I see.

Not `[Null]` will shrink to `Null`:

```haskell
["foo"]        -- shrinks to
"foo", [], ... -- shrinks to
null, ...
```

but e.g. empty array wont.

I guess it won't hurt to add `Null` to the from of every case.

phadej: Done

NorfairKing:
I'm a tad sad about this.
I wished my prod executables wouldn't depend on QuickCheck.

PR #922, opened 2022-01-12
Add explicit toEncoding method to ToJSON QuarterOfYear

TeofilC:
It seems to me like this was missed out when adding this instance.
A function was defined to create an `Encoding` but it wasn't used in the instance.


phadej:
oops. Thanks for spotting.

PR #930, opened 2022-02-16
Require years with at least 4 digits, and document the time format

### LINKED ISSUE #929: `Day` parser is more lenient than docs suggest
The docs state that `day` parses years in `YYYY` format:
https://github.com/haskell/aeson/blob/65fca856304c0b323d538cc783493d8d7a145114/attoparsec-iso8601/src/Data/Attoparsec/Time.hs#L44-L48
In reality it can parse any number of digits as a year since it uses `decimal` and does no further checking.

Lysxia:
Closes #929


attoparsec-iso8601/src/Data/Attoparsec/Time.hs
```diff
@@ -72,6 +82,17 @@ quarter = do
       <|> Q3 <$ char '3'
       <|> Q4 <$ char '4'
 
+-- | Parse a year @YYYY@, with at least 4 digits. Does not include any sign.
```
phadej, line 85: It might be worth noting that `Year` is just a `type Year = Integer`, not `newtype`. And in relatively recent `time` (`1.11`). `time-compat`  defines such an alias for older `time`s too.

attoparsec-iso8601/src/Data/Attoparsec/Time.hs
```diff
@@ -72,6 +82,17 @@ quarter = do
...
+year :: Parser Year
+year = do
+  ds <- takeWhile1 isDigit
+  if T.length ds < 4 then
+    fail "expected year with at least 4 digits"
+  else
+    pure (T.foldl' step 0 ds)
```
phadej, line 92: This is bad. See https://github.com/haskell/aeson/issues/673 and its fix.

phadej (changes requested):
Add tests (`SerializationFormatSpec.hs`).

attoparsec-iso8601/src/Data/Attoparsec/Time.hs
Lysxia: uhoh. Thanks for catching this.

Lysxia: BTW `decimal` is also used here:

https://github.com/haskell/aeson/blob/65fca856304c0b323d538cc783493d8d7a145114/src/Data/Aeson/Parser/Internal.hs#L498

that  should be removed too right?

phadej: `scientific :: Integer -> Int -> Scientific` so that `decimal` is parsing `Int` and is fine.

Lysxia: It won't run in quadratic time, so no DoS, but it will still overflow with a wrong result, which is undesirable.

Lysxia:
The time parsers work on `Text` which is a bit annoying because `bsToInteger` from `Data.Aeson.Parser.Internal` takes a bytestring, but the algorithm internally works on lists anyway so it wasn't too hard to adapt.

phadej (approved):

phadej:
Squased, merged (and added changelogs + version bump) in https://github.com/haskell/aeson/pull/945. Thanks!

PR #931, opened 2022-02-22
Correct `@since` version for `KeyMap.elems`

ysangkok:
Seems like the wrong version was added in 02702b4bc0a2077a59b99feca00c28fc79072024.


phadej:
thanks

PR #936, opened 2022-03-29
Remove SCC pragmas

TeofilC:
This patch removes explicit SCC pragmas from aeson.

### Motivation
When profiling an application that uses aeson heavily, profiles become dominated with cost centres from aeson. This can partly be alleviated by passing `-fno-prof-auto`, etc, in aeson's ghc-options. Yet, there are still some explicit SCC pragmas heavily trafficked parts of aeson. These still dominate. 

I think users would benefit from not having these cost centres cluttering their profiles, especially because afaik cost centres can inhibit optimisations, and these SCC pragmas live in functions marked INLINE.


phadej:
Yes, let's remove them. I don't think they add value for us profiling aeson either.

TeofilC:
Thanks!

PR #937, opened 2022-04-28
fix text-short FromJSON withText typo

raehik:
text-short would report `Lazy Text` for a parse failure. This replaces that with `ShortText`.


phadej:
Thanks

Issue #933, opened 2022-03-12
RFC: Split generic deriving into specific cases, powered by DerivingVia.


phadej:
A problem I find in improving `aeson` is dealing with generic and TH code. It's monolithic, trying to cater for all needs at once.
Issues like https://github.com/haskell/aeson/issues/824 and https://github.com/haskell/aeson/issues/827 is hard to incorporate as everything is coupled. Compiling generics code is also slow (https://github.com/haskell/aeson/issues/810) as simplifier needs to work out a lot of code.

I propose a long term strategy of splitting various kinds of encoding formats into separate newtype wrappers to be used with `DerivingVia`
- The https://github.com/haskell/aeson/issues/909 issue would be a good start, to provide the "correct" Generic deriving. (would resolve https://github.com/haskell/aeson/issues/660)
- The following step would be to provide deriving newtypes for product types. We might introduce `To/FromJSONObject` classes here.
- I propose to not touch the current deriving code for a while, until the alternatives are sufficiently expressive.
- ... but when they are, remove the current generics code and default implementations. That would be a major breaking change at that point, but on the other hand, any further changes to generics code will be much more localized.
    - That also will allow to move generics code out of `ToJSON` & `FromJSON` modules, making working on `aeson` slightly more pleasant.

"simplifying" generics story would also help to improve other aspects of aeson, e.g. https://github.com/haskell/aeson/pull/839 is somewhat stuck as getting the change through current generics code is very hard.

Note: TH code should be adopted to be configurable similarly to DerivingVia code, if people would prefer to use TH. This shouldn't be difficult.

In summary: 1. Add an alternative generics code to be used with `DerivingVia`. 2. Remove old generics code. 3. Profit.

ping @bergmark @Lysxia

bergmark:
Sounds good!

Does DerivingVia compose well? E.g. if I want to use both `fieldLabelModifier` and `rejectUnknownFields` together, how would that look?

phadej:
> Sounds good!
> 
> Does DerivingVia compose well? E.g. if I want to use both `fieldLabelModifier` and `rejectUnknownFields` together, how would that look?

It doesn't. Each "newtype" has to implement the features, but can use the same auxiliary code for that.

treeowl:
> That also will allow to move generics code out of ToJSON & FromJSON modules, making working on aeson slightly more pleasant.

That approach tends to lead to orphan instances. Are those considered okay here?

phadej:
[...]

PR #939, opened 2022-05-09
Introduce a flag to default toEncoding to Generic rather than an implementation in terms of toJSON

TeofilC:
This adds a flag (alternative names appreciated) that defines the default toEncoding method of ToJSON in terms of Generics rather than the toJSON method. This introduces symmetry between the toJSON and toEncoding methods, and makes it a lot easier to try to eliminate uses of toJSON from a codebase.

This change is placed behind a flag as it would break code that doesn't use Generic deriving and instead only gives an implementation of toJSON.

### DIFF
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -57,6 +57,11 @@ flag ordered-keymap
   default:     True
   manual:      True
 
+flag default-toEncoding-to-toJSON
+  description: Use the toJSON method to give a default value for toEncoding
+  default: False
+  manual: True
+
 library
   default-language: Haskell2010
   hs-source-dirs:   src attoparsec-iso8601/src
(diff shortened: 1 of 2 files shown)


TeofilC:
This flag would be off by default. But if your preference is to go for DerivingVia that makes sense.

I would still like a flag that disables the default instance of `toEncoding` in terms of `toJSON` though as it makes it viable to find cases where a `toEncoding` instance was missed. This would be off be default. What do you think of that?

phadej:
My (#933) long term plan is to remove default implementation all-together.

TeofilC:
Sure that makes sense and I agree is the right approach. 
If you're not interested in having a flag for this in the interm I'll close this PR.

PR #941, opened 2022-05-21
Expose KeyMap and Key constructors in Internal modules

thomasjm:
Aeson 2 introduced the opaque `Key` and `KeyMap` types. It's good that they are opaque, but it prevents the user from deriving `newtype` instances for them. Since the wrapped types are simply `HashMap`, `Map`, and `Text`, useful instances for them exist throughout the ecosystem and it would be nice to be able to leverage them.

For example, when working with the `flat` serialization library, I'd like to be able to write

```haskell
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
deriving newtype instance Flat A.Key
deriving newtype instance (Flat a) => Flat (A.KeyMap a)
```

These `deriving` statements only work if the constructors are visible, so this PR exposes them in `Data.Aeson.Key.Internal` and `Data.Aeson.KeyMap.Internal` respectively.

### DIFF
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -67,7 +67,9 @@ library
     Data.Aeson.Internal
     Data.Aeson.Internal.Time
     Data.Aeson.Key
+    Data.Aeson.Key.Internal
     Data.Aeson.KeyMap
+    Data.Aeson.KeyMap.Internal
     Data.Aeson.Parser
     Data.Aeson.Parser.Internal
     Data.Aeson.QQ.Simple
@@ -81,7 +83,6 @@ library
     Data.Aeson.Internal.Functions
     Data.Aeson.Internal.Text
     Data.Aeson.Internal.TH
-    Data.Aeson.KeyMap.Internal
     Data.Aeson.Parser.Time
     Data.Aeson.Parser.Unescape
     Data.Aeson.Types.Class
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -81,6 +81,7 @@ library
     Data.Aeson.Internal.Functions
     Data.Aeson.Internal.Text
     Data.Aeson.Internal.TH
+    Data.Aeson.KeyMap.Internal
     Data.Aeson.Parser.Time
     Data.Aeson.Parser.Unescape
     Data.Aeson.Types.Class
--- a/src/Data/Aeson/Internal.hs
+++ b/src/Data/Aeson/Internal.hs
@@ -21,10 +21,8 @@ module Data.Aeson.Internal
     , formatError
     , ifromJSON
     , iparse
-    , KeyMap(..)
     ) where
 
 
-import Data.Aeson.KeyMap.Internal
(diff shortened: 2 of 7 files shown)


phadej:
The types are opaque on purpose, `KeyMap` may be implemented by HashMap or Map, and you cannot know. The same applies for `Key` (in future it may be optionally backed up by `ShortText` instead).

thomasjm:
Of course, I totally understand why they're opaque. I haven't made them non-opaque, I've just exposed them in a `.Internal` module. 

Aeson already has several such modules, containing explicitly unstable exports. What's the harm in exposing a couple more constructors this way, so that users who are willing to risk breakage can use them to derive `newtype` instances?

Without some kind of compromise like this, Aeson 2 makes it impossible to derive these instances which were possible in Aeson 1. This has ramifications for working with other libraries in the ecosystem (for example Morpheus GraphQL, which makes heavy use of `Generic`).

Lysxia:
+1 Exposing the constructors in `.Internal` seems like a fair compromise to me.

Boarders:
I am torn on the question as it is very annoying not being able to use `DerivingVia` due to a constructor not being exposed by a library. On the other hand, can the issue be solved by proving `KeyMap` and `Key` with `Generic` instances? Perhaps that is enough to solve your particular problem. It probably is a bad sign if wanting to derive instances directly from `Key` or `KeyMap` though but if you have some compelling use cases that can't be fixed with the addition of more instances in `aeson` then maybe that would sway the balance.

phadej:
The implementation can be changed with a package flag, then your code using `DerivingVia` will break. The `KeyMap` is opaque. Neither it can have reasonable `Generic` instance, as there is no structure to expose.

phadej:
Use https://hackage.haskell.org/package/aeson-2.0.3.0/docs/Data-Aeson-KeyMap.html#v:coercionToHashMap and https://hackage.haskell.org/package/aeson-2.0.3.0/docs/Data-Aeson-KeyMap.html#v:coercionToMap to still be able to do `coerce`. The instance definitions won't be as convenient, but they will be resilient against configuration changes.

thomasjm:
> The implementation can be changed with a package flag, then your code using DerivingVia will break.

Not necessarily -- in the case I mentioned with the `flat` library, there are instances for both `HashMap` and `Map`, as well as other common container types. Since `Key` and `KeyMap` tend to be wrappers for common types, I would expect this often to be the case. (In any case, breakage is a risk I'd be willing to take when using an `Internal` module.)

> can the issue be solved by proving `KeyMap` and `Key` with `Generic` instances?

That wouldn't help with my `flat` use-case unfortunately.

> [use] `coercionToHashMap` etc.

This doesn't sound too good to me, because implementing these instances using `Coercion` pattern matching will result in sort of a "partial" instance with a potential runtime crash if the underlying type doesn't turn out to be one of the ones I handle. I'd actually prefer a simple compile-time breakage when Aeson changes the underlying type.

PR #942, opened 2022-05-26
Data.Aeson.KeyMap re-exports key functions from Data.Aeson.Key

friedbrice:
My doctor says that this will help lower my blood pressure.

### DIFF
--- a/src/Data/Aeson/KeyMap.hs
+++ b/src/Data/Aeson/KeyMap.hs
@@ -88,8 +88,6 @@ module Data.Aeson.KeyMap (
     
     -- * Key Type
     Key,
-    fromText,
-    toText,
 ) where
 
 -- Import stuff from Prelude explicitly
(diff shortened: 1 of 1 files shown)


src/Data/Aeson/KeyMap.hs
```diff
@@ -85,6 +85,11 @@ module Data.Aeson.KeyMap (
     filterWithKey,
     mapMaybe,
     mapMaybeWithKey,
+    
+    -- * Key Type
+    Key,
+    fromText,
```
phadej, line 91: I'm fine with re-exporting `Key`, but `toText` and `fromText` read wrong:

```haskelll
key = KeyMap.fromText t
```

phadej (changes requested):

src/Data/Aeson/KeyMap.hs
friedbrice: Done!

phadej:
squashed and merged in https://github.com/haskell/aeson/pull/946, thanks!

PR #949, opened 2022-06-10
Resolve #927: Add MonadFix instance

### LINKED ISSUE #852: Add a way to fail, showing path to field, in `fromJSON`
Currently, AFAIU, `decode`-like functions always add the path to field into the error message, while `fromJSON` never does this. And there is no way to manage that without using the internal modules.

My use case - deferred parsing of some datatype piece. So that I apply `eitherDecode` to the big datatype, getting `Value` in one of its fields and later running `fromJSON` on it.

### LINKED ISSUE #927: MonadFix instance for Parser
I'm parsing JSON into a cyclic graph. Right now I'm producing `IntMap Foo -> Foo` and then taking a fixpoint to make `IntMap Foo`. If I had `MonadFix` available to me, I could eliminate the extra step, which would be very convenient.

It looks like the parser is a simple two continuation implementation, so a `MonadFix` instance taking a fixpoint over all branches shouldn't be too hard to implement. I'd be happy to submit a PR if the change would be welcome.

Thank you!

phadej:


src/Data/Aeson/Types/Internal.hs
```diff
@@ -306,6 +308,15 @@ instance Monad.Monad Parser where
     {-# INLINE fail #-}
 #endif
 
+-- |
+--
+-- @since 2.1.0.0
+instance MonadFix Parser where
+    mfix f = let p = f (lower [] p) in p
```
Lysxia, line 315: The fixed point should not be a function and the `path` should be taken from the `Parser` context.

```
-- I haven't typechecked this
mfix f = Parser $ \path kf ks -> let x = runParser (f (fromISuccess x)) path IError ISuccess in
  case x of
    IError p e -> kf p e
    ISuccess y -> ks y

fromISuccess :: IResult a -> a
```

phadej: But what's the difference. Can we write a test which differentiates this and the current impl?

Lysxia: `take 3 <$> mfix (\x -> pure (trace "x" (1 : x)))` should trace only once.

phadej: Both implementations trace only once, AFAICT. (By trying in ghci).

phadej: Anyway, yours has `path` thing, so it's better. Let's go with it.

Lysxia: Oops it's a bit more subtle than I thought, but here's a working litmus test:

```
>>> parse (\_ -> take 3 <$> mfix (\x -> return () >>= \_ -> pure (trace "THIS SHOULD PRINT ONLY ONCE" (() : x)))) ()
Success THIS SHOULD PRINT ONLY ONCE
[()THIS SHOULD PRINT ONLY ONCE
,(),()]
```

PR #950, opened 2022-06-10
Support for datatypes with no constructors in TH and generics

evgeny-osipenko:
This PR adds support for datatypes with no constructors in TH generators (`mkToJSON`, `mkToEncoding` and `mkParseJSON`) and generic instances (`GToJSON'`, `GFromJSON'`).

For example, this PR enables this code:
```hs
data NoConstructors
    deriving (Generic)
instance ToJSON NoConstructors
instance FromJSON NoConstructors

data MyVoid
deriveJSON defaultOptions ''MyVoid
```

On the current `master`, both variants fail: the generics are missing instances of `GToJSON' Encoding Zero V1` and `GFromJSON' Zero V1`, and TH makes an explicit call to `error`.

The actual behavior that is added:

- `toJSON x = seq x (error "case: V1")`
- `toEncoding x = seq x (error "case: V1")`
- `parseJSON _ = fail "Attempted to parse empty type"`

The main use case is to allow substituting such types into polymorphic sums, where some variants are not used for a particular instantiation, and we need to serialize/deserialize the sum; though, I think it's a natural generalization by itself. As an example of such use case:
```hs
class
    ( ToJSON (ResultType lang)
    , ToJSON (FailureType lang)
    , FromJSON (ResultType lang)
    , FromJSON (FailureType lang)
    ) =>
    RecordableResult (lang :: k)
  where
    data ResultType lang
    data FailureType lang

data RequestOutcome lang
    = OutcomeSuccess (ResultType lang)
    | OutcomeFailure (FailureType lang)
    deriving (Generic)
instance (RecordableResult lang) => ToJSON (RequestOutcome lang)
instance (RecordableResult lang) => FromJSON (RequestOutcome lang)

instance RecordableResult MonadDatabase where
    data ResultType MonadDatabase = DatabaseResult Aeson.Value
        deriving (Generic, ToJSON, FromJSON)
    data FailureType MonadDatabase = DatabaseError DatabaseException
        deriving (Generic, ToJSON, FromJSON)

instance (ToJSON r, FromJSON r) => RecordableResult (MonadReader r) where
    data ResultType (MonadReader r) = ReaderResult r
        deriving (Generic, ToJSON, FromJSON)
    data FailureType (MonadReader r) -- MonadReader never fails
        deriving (Generic, ToJSON, FromJSON)
        -- but we still need ToJSON/FromJSON for the surrounding machinery to work
```

### DIFF
--- a/src/Data/Aeson/TH.hs
+++ b/src/Data/Aeson/TH.hs
@@ -327,8 +327,8 @@ consToValue :: ToJSONFun
             -- ^ Constructors for which to generate JSON generating code.
             -> Q Exp
 
-consToValue _ _ _ _ [] = error $ "Data.Aeson.TH.consToValue: "
-                             ++ "Not a single constructor given!"
+consToValue _ _ _ _ [] =
+    [| \x -> x `seq` error "case: V1" |]
 
 consToValue target jc opts instTys cons = autoletE liftSBS $ \letInsert -> do
     value <- newName "value"
@@ -688,8 +688,8 @@ consFromJSON :: JSONClass
              -- ^ Constructors for which to generate JSON parsing code.
              -> Q Exp
 
-consFromJSON _ _ _ _ [] = error $ "Data.Aeson.TH.consFromJSON: "
-                                ++ "Not a single constructor given!"
+consFromJSON _ _ _ _ [] =
+    [| \_ -> fail "Attempted to parse empty type" |]
 
 consFromJSON jc tName opts instTys cons = do
   value <- newName "value"
@@ -1154,7 +1154,7 @@ instance {-# OVERLAPPABLE #-} LookupField a where
 
 instance {-# INCOHERENT #-} LookupField (Maybe a) where
     lookupField pj _ _ = parseOptionalFieldWith pj
- 
+
 #if !MIN_VERSION_base(4,16,0)
 instance {-# INCOHERENT #-} LookupField (Semigroup.Option a) where
     lookupField pj tName rec obj key =
--- a/src/Data/Aeson/TH.hs
+++ b/src/Data/Aeson/TH.hs
@@ -1,5 +1,9 @@
+{-# OPTIONS_GHC -Wno-incomplete-patterns #-}
+-- The function `emptyCase` emits the warning, but we should only use it for
+-- Void-like types, where there are no patterns to match on
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE EmptyCase #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE NoImplicitPrelude #-}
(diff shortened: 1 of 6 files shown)


src/Data/Aeson/TH.hs
```diff
@@ -327,8 +327,8 @@ consToValue :: ToJSONFun
             -- ^ Constructors for which to generate JSON generating code.
             -> Q Exp
 
-consToValue _ _ _ _ [] = error $ "Data.Aeson.TH.consToValue: "
-                             ++ "Not a single constructor given!"
+consToValue _ _ _ _ [] =
+    [| \x -> x `seq` error "case: V1" |]
```
phadej, line 331: use `EmptyCase`.

src/Data/Aeson/Types/ToJSON.hs
```diff
@@ -829,6 +829,12 @@ instance ( ToJSON1 f
 --------------------------------------------------------------------------------
 -- Generic toEncoding
 
+instance GToJSON' Encoding arity V1 where
+    -- Empty values do not exist, which makes the job of formatting them
+    -- rather easy:
+    gToJSON _ _ x = x `seq` error "case: V1"
```
phadej, line 835: `EmptyCase`

phadej (changes requested):

src/Data/Aeson/Types/ToJSON.hs
evgeny-osipenko: Done

src/Data/Aeson/TH.hs
```diff
@@ -328,7 +332,9 @@ consToValue :: ToJSONFun
             -> Q Exp
 
 consToValue _ _ _ _ [] =
-    [| \x -> x `seq` error "case: V1" |]
+    -- If we put the expression into the TH-generated code directly,
+    -- that makes it require EmptyCase at use-sites too
+    [| emptyCase |]
```
phadej, line 337: why not `[| \x -> case x of {} |]`?

It's fine to require users to enable `EmptyCase`.

src/Data/Aeson/TH.hs
evgeny-osipenko: Done here as well, however, I moved the expression into a separate function, otherwise the empty-case expression in the splice will require enabling `EmptyCase` at the use site too

src/Data/Aeson/TH.hs
evgeny-osipenko: Is there any benefit in doing that?

phadej: Yes, `-ddump-splices` doesn't look suspicious.

evgeny-osipenko: Done

evgeny-osipenko:
None of the instances were actually used, so I just removed them.

phadej:
Merged squashed in https://github.com/haskell/aeson/pull/971 (with some additions). Thanks!

PR #963, opened 2022-08-30
Implement KeyMap.!? operator.

Javran:
This is to match the counterpart in HashMap, making it smoother migration to aeson-2+


phadej:
That's still not correct.

phadej:
Thanks! merged in https://github.com/haskell/aeson/pull/964

Issue #952, opened 2022-06-14
Add KeyMap.insertWith


TeofilC:
KeyMap currently doesn't export a function that corresponds to {,Hash}Map.insertWith.

I would be interested in having this function. If this is useful for others as well, I'd be happy to write an MR.

TeofilC:
Thanks!

PR #965, opened 2022-09-01
Add an insertWith function.

philderbeast:
See #952.


phadej (changes requested):
Same problem as in #963: ou forgot there is also a HashMap implementation.

Also you should use `insertWith` from `(unordered-)containers`, and not reinvent it.

PR #968, opened 2022-09-16
Use unsafeDupablePerformIO in unsafePackLenLiteral

phadej:
Does @RyanGlScott have any ideas why the issue is happening in the first place, and this fixes it?

### DIFF
--- a/src/Data/Aeson/Internal/ByteString.hs
+++ b/src/Data/Aeson/Internal/ByteString.hs
@@ -13,8 +13,8 @@ import Data.Word (Word8)
 import Foreign.ForeignPtr (ForeignPtr)
 import Data.ByteString.Short (ShortByteString, fromShort)
 import GHC.Exts (Addr#, Ptr (Ptr))
-import Data.ByteString.Internal (accursedUnutterablePerformIO)
 import Data.ByteString.Short.Internal (createFromPtr)
+import System.IO.Unsafe (unsafeDupablePerformIO)
 
 import qualified Data.ByteString as BS
 import qualified Language.Haskell.TH.Lib as TH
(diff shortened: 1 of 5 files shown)


src/Data/Aeson/Internal/ByteString.hs
```diff
@@ -84,4 +84,4 @@ liftSBS sbs = withBS bs $ \_ len -> [| unsafePackLenLiteral |]
 
 unsafePackLenLiteral :: Int -> Addr# -> ShortByteString
 unsafePackLenLiteral len addr# =
-    accursedUnutterablePerformIO $ createFromPtr (Ptr addr#) len
+    unsafeDupablePerformIO $ createFromPtr (Ptr addr#) len
```
RyanGlScott, line 87: It might be worth pointing out that this implementation was taken almost verbatim from `bytestring`. This would serve as a pointer to those interested in why `unsafeDupablePerformIO` in particular was chosen.

RyanGlScott (commented):
If I understand correctly, the previous `accursedUnutterablePerformIO`-based version was misbehaving due to the issue fixed in haskell/bytestring#542? If so, then the fix chosen here looks sensible. It might be worth updating the title of this PR, which appears to be out of date with respect to the current code.

src/Data/Aeson/Internal/ByteString.hs
phadej: Done

phadej:
@RyanGlScott thanks for the review.

phadej:
For interested:

```haskell
unsafeDupablePerformIO  :: IO a -> a
-- See Note [unsafePerformIO and strictness]
unsafeDupablePerformIO (IO m) = case runRW# m of (# _, a #) -> lazy a

accursedUnutterablePerformIO :: IO a -> a
accursedUnutterablePerformIO (IO m) = case m realWorld# of (# _, r #) -> r
```

That's why both fixes, using `unsafeDupablePerformIO` and (previous try of) making the let bindings lazy worked. `unsafeDupablePerformIO` is more principled.

Issue #882, opened 2021-10-11
Add "since" information to documentation


andreasabel:
E.g. module `Data.Aeson.Key` is new.  A `Since: 2.0.0.0` would be helpful.
Likewise for other new functionality.

PR #973, opened 2022-10-16
Add @since documentation (#882)

blackheaven:


phadej (changes requested):
Adding since for `Key` and `KeyMap` is fine. Everything else seems to be more noise than useful information nowadays. (You really shouldn't use anything older than `aeson-2`).

PR #981, opened 2022-11-08
Improve `Arbitrary` instance.

### LINKED ISSUE #980: Arbitrary instance seems low quality
Here are some samples I get from it (aeson-2.0.3.0, but it looks like the instance hasn't changed):

```
ghci> let mySample n = do { vals <- sample' $ resize n (arbitrary @Value); mapM_ (putStrLn . read . show . encode) vals; putStrLn "" }

ghci> mySample 0
""
false
""
true
null
null
""
null
0
0
null

ghci> mySample 1
[]
[]
{}
{}
{}
[]
{}
[]
{}
[]
{}

ghci> mySample 2
[[]]
[[]]
[[]]
[[]]
[[]]
{"":[]}
[[]]
{">X":{}}
[[]]
[{}]
{"ó±¦¤ð":{}}

ghci> mySample 3
{"\u000fW":{},"q":[]}
[[],[]]
[[{}]]
{"\\C":{"2":[]}}
[{},[]]
[[],{}]
[{"":{}}] ±":[]}
[{},{}],"á
[{},{}]
[[[]]]
```

The current implementation is:

```haskell
instance QC.Arbitrary Value where
    arbitrary = QC.sized arbValue

arbValue :: Int -> QC.Gen Value
arbValue n
    | n <= 0 = QC.oneof
        [ pure Null
        , Bool <$> QC.arbitrary
        , String <$> arbText
        , Number <$> arbScientific
        ]

    | otherwise = QC.oneof
        [ Object <$> arbObject n
        , Array <$> arbArray  n
        ]

arbText :: QC.Gen Text
arbText = T.pack <$> QC.arbitrary

arbScientific :: QC.Gen Scientific
arbScientific = Sci.scientific <$> QC.arbitrary <*> QC.arbitrary

arbObject :: Int -> QC.Gen Object
arbObject n = do
    p <- arbPartition (n - 1)
    KM.fromList <$> traverse (\m -> (,) <$> QC.arbitrary <*> arbValue m) p

arbArray :: Int -> QC.Gen Array
arbArray n = do
    p <- arbPartition (n - 1)
    V.fromList <$> traverse arbValue p

arbPartition :: Int -> QC.Gen [Int]
arbPartition k = case compare k 1 of
    LT -> pure []
    EQ -> pure [1]
    GT -> do
        first <- QC.chooseInt (1, k)
        rest <- arbPartition $ k - first
        QC.shuffle (first : rest)
```

And I think there are two problems here. The first is that `arbText` and `arbScientific` are only called when the generator size is 0, so they only produce `""` and `0`. The second is that `arbPartition` never produces a list containing 0, so it never puts `Null`/`Bool`/`String`/`Number` inside an `Object` or `Array`.

I expect that simply allowing the first four constructors to be produced with a positive generator size would be sufficient here, though having `arbPartition` sometimes generate a 0 might also be a mild improvement.

ChickenProp:
Fixes #980. As discussed there, simply changing `n <= 0` to `n <= 1`
gives good generation of strings, bools, numbers and null. Turns out
it also forbids `[]` and `{}` from being generated, so I added those
explicitly to that case.

My understanding is this requires a major version bump, as a change in
behavior of an exported function.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 For the latest version of this document, please see [https://github.com/haskell/aeson/blob/master/changelog.md](https://github.com/haskell/aeson/blob/master/changelog.md).
 
+### Upcoming
+
+* Improve `Arbitrary` instance (see #980, requires major version bump).
+
 ### 2.1.2.0
 
 * Support deriving for empty datatypes (such as `Void` and `V1`)
--- a/src/Data/Aeson/Types/Internal.hs
+++ b/src/Data/Aeson/Types/Internal.hs
@@ -453,11 +453,13 @@ type RepValue
 
 arbValue :: Int -> QC.Gen Value
 arbValue n
-    | n <= 0 = QC.oneof
+    | n <= 1 = QC.oneof
         [ pure Null
         , Bool <$> QC.arbitrary
         , String <$> arbText
         , Number <$> arbScientific
+        , pure emptyObject
+        , pure emptyArray
         ]
 
     | otherwise = QC.oneof


Lysxia (approved):
LGTM

phadej:
Looks good indeed, Yet I'll write some tests to we can actually claim anything about quality of generated values.

PR #989, opened 2023-01-20
Add encodingToStrictByteString

phadej:
This runs Builder to produce Strict ByteString directly, by making a mutable buffer and growing it exponentially.

This might be good or bad, better or worse than
LBS.toStrict . encodingToLazyByteString.
Latter allocates many small chunks, and copies once; encodingToStrictByteString makes a buffer exponentially, but copies data everytime.

TL;DR `bytestring`s `Builder` implementation allows to execute it into a contiguous buffer, i.e. strict bytestring.

(We could almost reuse the same trick to produce strict text as well, but we'd need to copy to unpin resulting bytearray)

### DIFF
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -81,6 +81,7 @@ library
     Data.Aeson.Internal.Functions
     Data.Aeson.Internal.Prelude
     Data.Aeson.Internal.Scientific
+    Data.Aeson.Internal.StrictBuilder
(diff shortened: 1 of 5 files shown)


phadej:
Please don't merge this. This wasn't any more efficient than serializing to lazy bytestring and then converting to strict one if needed.

PR #1006, opened 2023-02-27
Support building with `th-abstraction-0.5.*`

RyanGlScott:
This introduces a `TypeData` constructor for `DatatypeVariant` that characterizes `type data` declarations (introduced in GHC 9.6). Because `type data` data constructors only exist at the type level, not the value level, it doesn't make sense to define `ToJSON` or `FromJSON` instances for a `type data` declaration. As a result, this patch makes it an error to use `aeson`'s TH machinery in conjunction with a `type data` declaration.


src/Data/Aeson/TH.hs
```diff
@@ -1946,6 +1949,17 @@ existentialContextError conName = error
   . showString "‘ must be truly polymorphic in the last argument(s) of the data type"
   $ ""
 
+#if MIN_VERSION_th_abstraction(0,5,0)
+-- | We cannot implement class methods at the term level for @type data@
+-- declarations, which only exist at the type level.
+typeDataError :: Name -> Q a
+typeDataError dataName = fail
```
phadej, line 1956: What GHC-9.6 will tell if I try to do

```hasell
type data Foo = Foo

class ToJSON Foo
```

or just

```haskell
type data Foo = Foo
  deriving (Ord)
```

phadej: Looks like GHC is very much happy if I do:

```haskell
{-# LANGUAGE TypeData #-}
type data Foo = MkFoo
instance Show Foo where showsPrec _ _ = showString "MkFoo"
```

```haskell
Prelude> undefined :: Foo
MkFoo
```

So, I don't really know what to do with `type data` stuff.

I think TH should just generate the code and let GHC barf (as constructors are not in the term name space).

RyanGlScott: > I think TH should just generate the code and let GHC barf (as constructors are not in the term name space).

I was originally worried that this would generate incomprehensibly bad error messages, it's actually not as bad as I imagined it would be:

```
error:
    • Type constructor ‘MkFoo’ used as a constructor-like thing
    • In the pattern: MkFoo
      In a case alternative: MkFoo -> ...
```

In that case, I'm fine to defer the error to GHC's end. I've pushed a new version of the code that simply passes `type data` declarations through without any further scrutiny.

PR #1021, opened 2023-05-02
docs: use ssh protocol to clone

xieyuschen:
When using the git protocol to pull the repo, it requires additional setup work. The repo only has `40.43 MiB` when cloning, we could use `ssh` to clone it directly I believe.

Regards


xieyuschen:
Hi @phadej , could you kindly check my pr?

PR #1022, opened 2023-05-20
Warn about possible exceptions during file decoding

deemp:
In my project, I spent a couple of hours trying to determine the reason for an exception `withBinaryFile: does not exist (No such file or directory)`.
After reading the current `aeson` docs, I expected that the case when the file is missing is also reported as a String error.

I see that currently, this is not true. I'm not sure these functions should handle missing files so I suggest to just warn about possible exceptions.

### DIFF
--- a/src/Data/Aeson.hs
+++ b/src/Data/Aeson.hs
@@ -199,6 +199,8 @@ encodeFile fp = L.writeFile fp . encode
 --
 -- This function parses immediately, but defers conversion.  See
 -- 'json' for details.
+--
+-- Throws an 'Exception' when the file is missing.
 decodeFileStrict :: (FromJSON a) => FilePath -> IO (Maybe a)
 decodeFileStrict = fmap decodeStrict . B.readFile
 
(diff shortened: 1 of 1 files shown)


Lysxia:
Thanks!

src/Data/Aeson.hs
```diff
@@ -199,6 +199,8 @@ encodeFile fp = L.writeFile fp . encode
 --
 -- This function parses immediately, but defers conversion.  See
 -- 'json' for details.
+--
+-- Throws an 'Exception' when the file is missing.
```
phadej, line 203: This is misleading. `decodeFileStrict` (and others) throw all the errors `B.readFile` may throw, not only missing file. More correctly it should say "Doesn't modify exceptions thrown by `Data.ByteString.readFile` used to read the file". FWIW, `readFile` doesn't document any exception behavior either.

```
*Data.Aeson> decodeFileStrict "foo" :: IO (Maybe Int)
*** Exception: foo: withBinaryFile: permission denied (Permission denied)
```

Lysxia: It's a bit strong to call this "misleading", but feel free to make it more precise.

PR #1023, opened 2023-05-26
Type-directed optional fields

friedbrice:
This PR makes a backwards-compatible change to `ToJSON` and `FromJSON` that allows the instance to configure the optionality of record fields of that type when using Generic deriving or Template Haskell deriving.

Specifically, we make the following use-facing changes:

```haskell
class ToJSON a where
    ...
    omitField :: a -> Bool
    omitField _ = False
```

and

```haskell
class FromJSON a where
    ...
    omittedField :: Maybe a
    omittedField = Nothing
```

The new methods have default implementations that preserve the current behavior of Generically-derived and Template Haskell created instances.
Instance authors can use these methods to configure how to represent fields of this type as omitted from a JSON record.
As an added benefit, this PR eliminates the needs for special-case handling of `Maybe` and `Option`, significantly simplifying the Generic and Template Haskell code.

This PR solves https://github.com/haskell/aeson/issues/646 and https://github.com/haskell/aeson/issues/792, and it's similar to https://github.com/haskell/aeson/pull/839 by @phadej.
Please let me know what you think, if this PR is a suitable addition to Aeson, and if there's anything about it you'd like me to fix or change.

Thanks much!

### DIFF
--- a/src/Data/Aeson/Types/FromJSON.hs
+++ b/src/Data/Aeson/Types/FromJSON.hs
@@ -70,7 +70,6 @@ module Data.Aeson.Types.FromJSON
     , (.:?)
     , (.:!)
     , (.!=)
-
     -- * Internal
     , parseOptionalFieldWith
     ) where
@@ -378,6 +377,33 @@ class FromJSON a where
           zipWithM (parseIndexedJSON parseJSON) [0..]
         . V.toList
         $ a
+
+    -- | Default value for optional fields.
+    --
+    -- Defining @omittedField = 'Just' x@ makes object fields of this type optional.
+    -- When the field is omitted, the default value @x@ will be used.
+    --
+    -- @
+    --   newtype A = A Int deriving (Generic)
+    --   instance FromJSON A where omittedField = Just (A 0)
+    --
+    --   data R = R { a :: A, b :: Int } deriving ('Generic', 'FromJSON')
+    --
+    --   decode "{\"b\":1}" -- Just (R (A 0) 1)
+    -- @
+    --
+    -- Defining @omittedField = 'Nothing'@ makes object fields of this type required.
+    --
+    -- @
+    --   omittedField :: Maybe Int -- Nothing
+    --   decode "{\"a\":1}" -- Nothing
+    -- @
+    --
+    -- The default implementation is @omittedField = Nothing@.
+    --
+    -- @since x.x.x.x
+    omittedField :: Maybe a
+    omittedField = Nothing
 
 -- | @since 2.1.0.0
 instance (Generic a, GFromJSON Zero (Rep a)) => FromJSON (Generically a) where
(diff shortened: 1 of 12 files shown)


tests/UnitTests/OptionalFields.hs
```diff
@@ -0,0 +1,122 @@
...
+optionalFields :: [TestTree]
+optionalFields = [omitGenerics, omitTH, proofOfConcept]
+
+
+-- c.f. https://github.com/haskell/aeson/pull/839#issuecomment-782453060
+data POC = POC
+  { x :: Nullable Int -- Field is required, but can be null.
+  , y :: Undefineable Int -- Field is optional, but cannot be null.
+  , z :: NullOrUndefineable Int -- Field is optional, and can be null.
+  }
```
friedbrice, line 20: These tests verify that the PR supports your desired behavior (https://github.com/haskell/aeson/pull/839#issuecomment-782453060), @bergmark :)

src/Data/Aeson/Types/ToJSON.hs
```diff
@@ -1291,6 +1273,7 @@ instance ToJSON1 Maybe where
 
 instance (ToJSON a) => ToJSON (Maybe a) where
     toJSON = toJSON1
+    omitField = null
     toEncoding = toEncoding1
```
friedbrice, line 1277: A `Maybe a` will be omitted when (1) `omitNothingFields` is `True`, and (2) when the maybe is empty.

This preserves the current `ToJSON` behavior of `Maybe a` without needing overlapping instances.

src/Data/Aeson/Types/FromJSON.hs
```diff
@@ -1517,6 +1518,7 @@ instance FromJSON1 Maybe where
 
 instance (FromJSON a) => FromJSON (Maybe a) where
     parseJSON = parseJSON1
+    omittedDefault = Just Nothing
```
friedbrice, line 1521: A field absent in JSON will be populated by `Nothing` in the Haskell representation, preserving the current behavior of `FromJSON` for `Maybe a` fields, without needing overlapping instances.

src/Data/Aeson/Types/ToJSON.hs
```diff
@@ -176,6 +176,9 @@ class GToJSON' enc arity f where
     -- and 'liftToEncoding' (if the @arity@ is 'One').
     gToJSON :: Options -> ToArgs enc arity a -> f a -> enc
 
+    gOmitField :: proxy enc -> proxy' arity -> f a -> Bool
```
phadej, line 179: Is idea of having `gOmitField` and `gOmittedDefault` to inherit the default value in wrapper types like `newtype`s or singular `data Wrapper = Wrap SomeOtherType` wrappers?

I don't think it's necessary.

It also is (or at least was) to prefer single method `G*` classes so they are inlined better.

TL;DR I'd rather not add generics for `omitField` nad `omittedField`. It will make patch smaller and we won't need to worry about generically derived code performance.

src/Data/Aeson/Types/ToJSON.hs
```diff
@@ -1291,6 +1273,7 @@ instance ToJSON1 Maybe where
 
 instance (ToJSON a) => ToJSON (Maybe a) where
     toJSON = toJSON1
+    omitField = null
```
phadej, line 1276: `isNothing` (it's the same as `null`, but more direct).

phadej (commented):
Looks good. Having `omitField :: a -> Bool` is indeed simpler that what I had in #839.

There probably should be something in `ToJSON1` and `FromJSON1` classes, as e.g. instance for `Fix` relies on them. How `Fix Maybe` should be serialised?

src/Data/Aeson/Types/ToJSON.hs
friedbrice: Okay, I'll see about removing them :-)

friedbrice: Done.

friedbrice:
I was able to get the generic classes down to one method like you suggested. I still need to work on the `ToJSON1` logic.

src/Data/Aeson/Types/FromJSON.hs
```diff
@@ -251,7 +251,10 @@ class GFromJSON arity f where
     -- or 'liftParseJSON' (if the @arity@ is 'One').
     gParseJSON :: Options -> FromArgs arity a -> Value -> Parser (f a)
 
+class GOmittedDefault arity f where
```
phadej, line 254: I probably was unclear: Don't add `gOmittedDefault` nor `gOmitField`. The generic deriving is cumbersome already. Unwrapping single constructors isn't a huge task to do manually, and I argue it's better to always be explicit which value will be omitted.

phadej:
I marked this as draft.

Please ping when you seem happy again.

tests/UnitTests/OptionalFields.hs
```diff
@@ -0,0 +1,118 @@
...
+-- c.f. https://github.com/haskell/aeson/pull/839#issuecomment-782453()6()
+data POC = POC
+  { x :: Nullable Int -- Field is required, but can be null.
+  , y :: Undefineable Int -- Field is optional, but cannot be null.
+  , z :: NullOrUndefineable Int -- Field is optional, and can be null.
+  }
+  deriving (Eq, Show, Generic)
+
+instance ToJSON POC where
+  toJSON = genericToJSON opts
```
phadej, line 22: missing `toEncoding`

friedbrice:
@phadej I'm pretty happy with it, now. I'm interested in your feedback :-)

friedbrice:
Oh, also, @phadej, I do have some ideas about changes to `To`/`FromJSON1`, but I'd like to know more about what you have in mind before attempting that.

phadej (commented):
This looks good.

I have further things to do for this, but I think it's easier and fasterfor me to them myself rather than try to explain them.

Could you squash your changes into own commit (and rebase on recent `master`), so I can continue from there.

tests/Encoders.hs
```diff
@@ -296,29 +296,6 @@ gOptionFieldParseJSON = genericParseJSON optsOptionField
 thMaybeFieldToJSON :: MaybeField -> Value
 thMaybeFieldToJSON = $(mkToJSON optsOptionField 'MaybeField)
 
-
---------------------------------------------------------------------------------
--- IncoherentInstancesNeeded
---------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
friedbrice: Thanks for your help and ideas, btw, @parsonsmatt. The TH was trivially-easy, but the Generics threw me for a loop.

src/Data/Aeson/Types/ToJSON.hs
```diff
@@ -317,6 +316,33 @@ class ToJSON a where
     toEncodingList :: [a] -> Encoding
     toEncodingList = listEncoding toEncoding
 
+    -- | Defines when it is acceptable to omit a field of this type from a record?
```
friedbrice, line 319:
```suggestion
    -- | Defines when it is acceptable to omit a field of this type from a record.
```

PR #1054, opened 2023-07-31
fix KeyMap.union doc

ncfavier:


phadej:
Thanks. I merged this manually (duplication the fix for other `union` implementation as well).

Issue #1061, opened 2023-08-29
CI: timeout of 60min a bit tight (or too many constraint sets)


andreasabel:
This run got canceled after 60 min: https://github.com/haskell/aeson/actions/runs/6010560497/job/16302233028?pr=1055
L'addition:
-  2 min setup
- 15 min build w/o tests
-  5 min build
-  2 min haddock
- 14 min constraint-set text-2.0
-  8 min constraint-set text-1.2
- 13 min+ constraint-set bytestring-0.11.2.0 (got cancelled)

And the hour is over...

phadej:
Resolved in https://github.com/haskell/aeson/pull/1071

PR #1069, opened 2023-09-28
Show everything builds `text-2.1` and `containers-0.7`

Vlix:
Just making this PR to run the CI. If that is all green, this PR should be proof `aeson-2.2.0.0` can be revised to allow `text-2.1` and `containers-0.7`


phadej:
This will be done as part of GHC-9.8 compatible release.

Not sure about `containers-0.7` though, as even with GHC-9.8 job, CI won't pick it up by default and the CI is already too slow for testing bumps in boot libs, #1061

Issue #1049, opened 2023-06-28
A line in the descriptions of `.:?=` and `.:!=` got mixed up?


Vlix:
The `.:!=` description says "...or the field is null and...", which is then immediately contradicted with the line after.
And in the descriptions of `.:?` and `.:!`, the "or the field is null" line is in the `.:?` description.

PR #1085, opened 2024-01-17
Fix documentation of .:!=

michaelpj:
I think this bit belongs in `.:?=`, by analogy with `.:!` and `.:?`.

Resolves https://github.com/haskell/aeson/issues/1049


phadej:
Is this issue noted in #1049?

michaelpj:
Ah yes, I think it's that issue. I just noticed it myself.

PR #1087, opened 2024-03-09
Link to updated time parser library

fishtreesugar:
since 2.2, aeson use [text-iso8601](https://hackage.haskell.org/package/text-iso8601) instead of [attoparsec-iso8601](https://hackage.haskell.org/package/attoparsec-iso8601)


phadej:
Thanks

PR #1088, opened 2024-03-21
Allow th-abstraction-0.7

phadej:


Kleidukos:
Would it be possible to perform a revision to allow people to transition to QuickCheck 2.15?

PR #1091, opened 2024-04-06
Add Data.Complex instances

TravisWhitaker:
There are loads of utility functions for defining the included instances. I guessed at which ones should be used for `Data.Complex`, but if I guessed incorrectly please say so.

I used a two-element array as the representation because that's how I've seen APIs in the wild do it, but perhaps `{i : ..., q : ...}` or `{real : ..., imag : ...}` would be preferred?


TravisWhitaker:
@phadej any thoughts on this?

TravisWhitaker:
Ah, sorry for missing `*JSON1`.

I'm curious what you'd think of having `{real : ..., imag : ...}` as the default instance, and then providing instances for the other two common ways via newtype instances (similar to the existing `DotNetTime` newtype).

phadej:
>  (similar to the existing DotNetTime newtype).
 
 Not worth the trouble.

Lysxia:
The representation as a 2-array of x+iy coordinates looks reasonable enough to me.

PR #1101, opened 2024-06-04
Add instances for  Data.Monoid 

AliceRixte:
Most of Data.Monoid newtypes already had an instance. I added the following ones : 

- Product
- Sum
- All
- Any

### DIFF
--- a/src/Data/Aeson/Types/FromJSON.hs
+++ b/src/Data/Aeson/Types/FromJSON.hs
@@ -2382,66 +2382,44 @@ instance FromJSON a => FromJSON (Monoid.Last a) where
     omittedField = omittedField1
 
 instance FromJSON1 Semigroup.Min where
-    liftParseJSON _ p _ a = coerce (p a)
-
-    liftParseJSONList _ _ p a = coerce (p a)
+    liftParseJSON _ p _ = coerce p
 
     liftOmittedField = coerce
 
 instance (FromJSON a) => FromJSON (Semigroup.Min a) where
     parseJSON = parseJSON1
 
-    parseJSONList = liftParseJSONList omittedField parseJSON parseJSONList
-
-    omittedField = omittedField1
-
 instance FromJSON1 Semigroup.Max where
-    liftParseJSON _ p _ a = coerce (p a)
+    liftParseJSON _ p _ = coerce p
 
-    liftParseJSONList _ _ p a = coerce (p a)
     liftOmittedField = coerce
 
 instance (FromJSON a) => FromJSON (Semigroup.Max a) where
     parseJSON = parseJSON1
 
-    parseJSONList = liftParseJSONList omittedField parseJSON parseJSONList
-    omittedField = omittedField1
-
 instance FromJSON1 Semigroup.First where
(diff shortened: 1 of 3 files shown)


src/Data/Aeson/Types/FromJSON.hs
```diff
@@ -2334,6 +2334,48 @@ instance FromJSON1 Monoid.Dual where
 instance FromJSON a => FromJSON (Monoid.Dual a) where
     parseJSON = parseJSON1
 
+instance FromJSON1 Monoid.Sum where
+    liftParseJSON _ p _ a = coerce (p a)
```
phadej, line 2338: I'd expect these to look like instances for `Down` or `Dual` above.

AliceRixte: I used Down instances as a model instead of Min, which was the first model I used. 

I took the opportunity to uniformize all newtype instances from Data.Monoid, including Dual, Min, Max, WrappedMonoid, First and Last. I can remove my last commit if you don't like it.

Cheers !

src/Data/Aeson/Types/ToJSON.hs
```diff
@@ -2100,12 +2100,42 @@ instance ToJSON a => ToJSON (Monoid.Dual a) where
     toEncoding = toEncoding1
     omitField = omitField1
 
+instance ToJSON1 Monoid.Sum where
+    liftToJSON _ t _ = t . Monoid.getSum
```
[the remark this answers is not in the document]
AliceRixte: Same over here : I took the opportunity to use coerce in all similar instances from Data.Monoid

src/Data/Aeson/Types/FromJSON.hs
```diff
@@ -2352,67 +2382,45 @@ instance FromJSON a => FromJSON (Monoid.Last a) where
...
     parseJSON = parseJSON1
 
-    parseJSONList = liftParseJSONList omittedField parseJSON parseJSONList
-    omittedField = omittedField1
-
 instance FromJSON1 Semigroup.WrappedMonoid where
-    liftParseJSON _ p _ a = coerce (p a)
+    liftParseJSON _ p _ = coerce p
 
-    liftParseJSONList _ _ p a = coerce (p a)
```
phadej, line 2407: this removals are wrong.

AliceRixte: I added it back, and made it coherent with the rest of the code

AliceRixte: I also wonder whether those two lines : 

```
parseJSONList = liftParseJSONList omittedField parseJSON parseJSONList
omittedField = omittedField1
```

should be added to all the FromJSON instances from Data.Monoid. If you think so I can do this.

phadej:
To make this easier I made https://github.com/haskell/aeson/pull/1104

So now the instances are simply using `DerivingVia` and are a lot more trivial. Sorry for causing the conflict, but hopefully now your patch will be simpler to write.

AliceRixte:
Pretty neat ! It was simpler for me to create a new pull request rather than cleaning up this one. Thanks a lot !

Issue #1059, opened 2023-08-12
Could not deduce `RecordFromJSON'` and `RecordToPairs` in `aeson-2.2.0.0`


julmb:
I have the following code
```hs
{-# LANGUAGE DeriveGeneric #-}

import Data.Aeson
import GHC.Generics

data Item f a = Item { root :: a, children :: f a } deriving Generic1

instance FromJSON1 f => FromJSON1 (Item f) where liftParseJSON = genericLiftParseJSON defaultOptions
instance ToJSON1 f => ToJSON1 (Item f) where liftToJSON = genericLiftToJSON defaultOptions
instance (FromJSON1 f, FromJSON a) => FromJSON (Item f a) where parseJSON = parseJSON1
instance (ToJSON1 f, ToJSON a) => ToJSON (Item f a) where toJSON = toJSON1

data Content f a = Content { items :: f (Item f a) } deriving Generic1

instance (Functor f, FromJSON1 f) => FromJSON1 (Content f) where liftParseJSON = genericLiftParseJSON defaultOptions
instance (Functor f, ToJSON1 f) => ToJSON1 (Content f) where liftToJSON = genericLiftToJSON defaultOptions
instance (Functor f, FromJSON1 f, FromJSON a) => FromJSON (Content f a) where parseJSON = parseJSON1
instance (Functor f, ToJSON1 f, ToJSON a) => ToJSON (Content f a) where toJSON = toJSON1

main :: IO ()
main = return ()
```

With `aeson-2.1.2.1`, this code works perfectly fine. With `aeson-2.2.0.0`, I get the following errors:
```
Main.hs:15:82: error:
    • Could not deduce (aeson-2.2.0.0:Data.Aeson.Types.FromJSON.RecordFromJSON'
                          One (f :.: Rec1 (Item f)))
        arising from a use of ‘genericLiftParseJSON’
      from the context: (Functor f, FromJSON1 f)
        bound by the instance declaration at Main.hs:15:10-58
    • In the expression: genericLiftParseJSON defaultOptions
      In an equation for ‘liftParseJSON’:
          liftParseJSON = genericLiftParseJSON defaultOptions
      In the instance declaration for ‘FromJSON1 (Content f)’
   |
15 | instance (Functor f, FromJSON1 f) => FromJSON1 (Content f) where liftParseJSON = genericLiftParseJSON defaultOptions
   |                                                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

PR #1103, opened 2024-06-04
Handle Functor Composition in Generic Instances

julmb:
This fixes all cases of #1059 that are known to me.

The changes are minimally-invasive with regard to the existing setup. However, they only allow the pattern `f :.: Rec1 g` for composition. I do not know if this covers all the cases that can arise with `DeriveGeneric` or if there is maybe a more clever way to set things up (see https://github.com/haskell/aeson/issues/1059#issuecomment-1675966700).

I am submitting this as a draft pull request since I still need to check some things and add regression tests. Since I barely understand the existing infrastructure, I would be grateful for early feedback on whether the proposed implementation is on the right track.

Once completed, I would only want this to be merged after someone more knowledgeable takes a critical look at it.

Tasks:
- [x] add missing instances
- [ ] check if optional fields are handled correctly
- [ ] add regression test

### DIFF
--- a/src/Data/Aeson/Types/FromJSON.hs
+++ b/src/Data/Aeson/Types/FromJSON.hs
@@ -1452,6 +1452,13 @@ instance {-# OVERLAPPING #-}
       recordParseJSONImpl (guard (allowOmittedFields opts) >> fmap Par1 o) gParseJSON args obj
     {-# INLINE recordParseJSON' #-}
 
+instance {-# OVERLAPPING #-}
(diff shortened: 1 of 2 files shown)


julmb:
This does not work in all cases since a field like `bar :: Maybe (Maybe (Maybe a))` will be represented as `Maybe :.: (Maybe :.: Rec1 Maybe)`, which the instance `RecordFromJSON' One (S1 s (f :.: Rec1 g))` cannot handle. An instance like `RecordFromJSON' One (S1 s (f :.: g))` does not work since there is no type class that provides field omission functionality on generic representations.

PR #1108, opened 2024-06-12
Handle Functor Composition in Generic Instances

julmb:
Second attempt at fixing #1059 after #1103 did not cover all cases. First discussion of the approach in https://github.com/haskell/aeson/issues/1059#issuecomment-2158531440.

Tasks:
- [x] add `GOmitFromJSON` and `GOmitToJSON` type classes and instances
- [x] replace `S1 s *` instances with single generic implementation based on `GOmitFromJSON` and `GOmitToJSON`
- [ ] investigate field omission propagation in existing instances `GFromJSON a (f :.: g)` and `GToJSON' e a (f :.: g)`
- [ ] inline `recordParseJSONImpl` that now has just a single call site
- [ ] add regression tests

I would be grateful for feedback by people more knowledgeable about the existing infrastructure than me.

### DIFF
--- a/src/Data/Aeson/Types/FromJSON.hs
+++ b/src/Data/Aeson/Types/FromJSON.hs
@@ -1013,29 +1013,17 @@ instance (FromJSON a) => GFromJSON arity (K1 i a) where
     gParseJSON _opts _ = fmap K1 . parseJSON
     {-# INLINE gParseJSON #-}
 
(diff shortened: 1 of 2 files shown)


julmb:
Rebased to `master`.

julmb:
I would still like to get this merged. Is there something I can or should do to help move this forward?

PR #1117, opened 2024-09-10
fix doc reference in Data.Aeson.Decoding.Text

exaexa:
A simple doc fix. I don't think a separate version bump or changelog entry is needed, best release this with any next version.


phadej:
Done in #1121

PR #1125, opened 2025-01-03
Add `Data.Aeson.Key.length`

### LINKED ISSUE #1123: Add `Data.Aeson.Key.length`
The JSON spec requires UTF-8 encoding ([source](https://www.rfc-editor.org/rfc/rfc8259.html#section-8.1)).  So the notion of the length of a key, defined as number of code points, seems well-defined.

The best way I've found is `Data.Text.length . Data.Aeson.Key.toText` which requires keeping up with current (unspecified) implementation to avoid extra work, plus adding `text` to explicit dependencies if not already there.

Happy to toss in a PR if deemed reasonable.

sjshuck:
Closes #1123.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 For the latest version of this document, please see [https://github.com/haskell/aeson/blob/master/changelog.md](https://github.com/haskell/aeson/blob/master/changelog.md).
 
+## Upcoming
+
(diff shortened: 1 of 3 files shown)


phadej:
Use `Data.Text.length . Data.Aeson.Key.toText`

PR #1133, opened 2025-03-16
Rework TH instances context

stevladimir:
Instead of deriving instances for all datatype variables iterate over constructors generating instances based on how variables are used. Does the right thing in cases like
*  Phantom type variables
  Does not add corresponding instance to the context for unused variables
*  Type families
  Instead of generating `(ToJSON a)` for `data Foo a = Foo (SomeTypeFamily a)` generates `ToJSON (SomeTypeFamily a)`

Note that it slightly changes the behavior in cases like
``` haskell
data Foo a = Foo a deriving (ToJSON)

data Bar a = Bar (Foo a)
deriveToJSON defaultOptions ''Bar
```
will result in `ToJSON (Foo a)` constraint rather than `ToJSON a` (current behavior). I believe the former is more precise as that is exactly what generated encoder wants. The small downside is requirement for FlexibleContext extension.

TODO and open questions:
* [ ] How to test TH?
  Are there any helpers to get easy access to instance context. I think it would be nice to have something like
``` haskell
$(getContext $ deriveToJSON ...) `shouldBe` "ToJSON a, ToJSON b"
```
Can add if required. But maybe there are already established approach to test that. 
* [ ] Which test cases to add?
  I sketched up some cases, but that list is not exhaustive. And they should be reworked according to previous point 
* [ ] Should we test both From and To classes?
* [ ] Add Changelog

### DIFF
--- a/src/Data/Aeson/TH.hs
+++ b/src/Data/Aeson/TH.hs
@@ -114,6 +114,8 @@ module Data.Aeson.TH
 
 import Data.Aeson.Internal.Prelude
 
+import Control.Monad
+import Data.Bifunctor
 import Data.Char (ord)
 import Data.Aeson (Object, (.:), FromJSON(..), FromJSON1(..), FromJSON2(..), ToJSON(..), ToJSON1(..), ToJSON2(..))
 import Data.Aeson.Types (Options(..), Parser, SumEncoding(..), Value(..), defaultOptions, defaultTaggedObject)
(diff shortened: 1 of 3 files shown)


stevladimir:
Should I explicitly ping some of the maintainers? No rush - asking just to make sure someone will eventually put an eye on this.
It is mostly ready except for some minor things mentioned in TODO list, but it makes sense to get some feedback (maybe it is no go at all) before proceeding with them.
@phadej maybe you can give a look or recommend someone to ping?

PR #1139, opened 2025-07-03
Fix #1138: Check for control characters in text literals everywhere

### LINKED ISSUE #1138: Tab in a JSON string isn't catched after an escaped quote
Consider two examples:

```json
["	\""]
```
and
```json
["\"	"]
```
(the spaces in the string literal are tabs).

The first one is an error:

> Unespected control character while parsing string literal

The second one produces a string `"\"\t"`, but should fail as far as I know.

phadej:
The first option from https://github.com/haskell/aeson/issues/1138#issuecomment-3031760070

### DIFF
--- a/aeson.cabal
+++ b/aeson.cabal
@@ -176,6 +176,7 @@ test-suite aeson-tests
     Regression.Issue571
     Regression.Issue687
     Regression.Issue967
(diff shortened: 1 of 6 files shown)


Lysxia (approved):
lgtm

PR #1142, opened 2025-07-31
Add option tagAsContentsFieldName to TaggedObject SumEncoding

wi11dey:
Hello,

Thanks for making aeson.

I've added a flag tagAsContentsFieldName which will allow encoding sum types as
```json
{
  "tag": "constructorName"
  "constructorName": { ... contents ... }
}
```
This style is used in type encoding standards such as https://palantir.github.io/conjure/#/docs/spec/wire?id=_54-union-json-format.

Initially I was thinking to make `contentsFieldName` be a `String -> String` and provide it the constructor name, which would be flexible enough to allow the default behavior of `const "contents"` and also be able to dynamically use the constructor name (and even modify it), but this would break backcompat and also not be serializable to the TH implementation.

Please let me know if you'd like any changes to the proposed API

### DIFF
--- a/src/Data/Aeson/TH.hs
+++ b/src/Data/Aeson/TH.hs
@@ -407,11 +407,14 @@ sumToValue letInsert target opts multiCons nullary conName value pairs
         case sumEncoding opts of
           TwoElemArray ->
             array target [conStr target opts conName, value]
-          TaggedObject{tagFieldName, contentsFieldName} ->
+          TaggedObject{tagFieldName, contentsFieldName, tagAsContentsFieldName} ->
             -- TODO: Maybe throw an error in case
             -- tagFieldName overwrites a field in pairs.
             let tag = pairE letInsert target tagFieldName (conStr target opts conName)
-                content = pairs contentsFieldName
+                contentsFieldName' = if tagAsContentsFieldName
+                                     then conString opts conName
+                                     else contentsFieldName
+                content = pairs contentsFieldName'
             in fromPairsE target $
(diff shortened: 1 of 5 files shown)


julmb:
I could use something like this in one of my projects. However, the constructor
```hs
TaggedObject { tagFieldName :: String, contentsFieldName :: String, tagAsContentsFieldName :: Bool }
```
admits the semantically questionable value `TaggedObject "type" "item" True`. If `tagAsContentsFieldName` is `True`, then the value in `contentsFieldName` is meaningless. I feel like it would be better to have
```hs
data ContentsFieldName = FromTag | Fixed String
... TaggedObject { tagFieldName :: String, contentsFieldName :: ContentsFieldName } ...
```
Alternatively, instead of `ContentsFieldName`, we could also use the isomorphic `Maybe String`, with a documentation note that a `contentsFieldName` value of `Nothing` means that the contents field gets the tag value as its name.

We could be even more extreme and not change any of the types in `TaggedObject` and simply treat `contentsFieldName` being the empty string as an indicator for using the tag value as content field name.

wi11dey:
> We could be even more extreme and not change any of the types in `TaggedObject` and simply treat `contentsFieldName` being the empty string as an indicator for using the tag value as content field name.

I like this. I've updated to the PR to this behavior. Let me know if it looks good to you

julmb:
I can't comment on the implementation, but from a design perspective, I would use it in my projects. Let's see what the maintainers have to say.

PR #1143, opened 2025-07-31
Add allNullaryConstructorTagModifier option

wi11dey:
Often, enums (all nullary contructors) are encoded differently than sum types.

For example, if I have
```haskell
data State = Launching | Launched | Terminated | MarkedForDeletion
```

I might want to have these states encoded as `LAUNCHING`, `LAUNCHED`, `TERMINATED`, `MARKED_FOR_DELETION`, respectively.

However, using the existing `constructorTagModifier` is not satisfactory for this because I don't necessarily want my sum types, e.g.

```
data Employee = Manager { subordinates :: [Employee] } | Junior
```

to be encoded the same way as an enum.

I've added an optional field `allNullaryConstructorTagModifier` (it's `Maybe` so that there's no behavior change to any existing usages of Aeson, in case someone is depending on `constructorTagModifier` also applying to all-nullary datatypes). When it's `Nothing`, everything is the same as today. When it's non-`Nothing` it's used instead of `constructorTagModifier` for all-nullary datatypes, for the purposes of handling enums differently than sum types.

Please let me know if you'd like any changes to the api

### DIFF
--- a/src/Data/Aeson.hs
+++ b/src/Data/Aeson.hs
@@ -131,6 +131,7 @@ module Data.Aeson
     , fieldLabelModifier
     , constructorTagModifier
     , allNullaryToStringTag
+    , allNullaryConstructorTagModifier
     , omitNothingFields
     , allowOmittedFields
     , sumEncoding
--- a/src/Data/Aeson/TH.hs
+++ b/src/Data/Aeson/TH.hs
@@ -335,7 +335,11 @@ consToValue target jc opts instTys cons = autoletE liftSBS $ \letInsert -> do
       -- forgotten.
       [con] | not (tagSingleConstructors opts) -> [argsToValue letInsert target jc tvMap opts False con]
       _ | allNullaryToStringTag opts && all isNullary cons ->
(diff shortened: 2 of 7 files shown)


julmb:
Since your `State` and `Employee` types are different types, why not just use a different `Options` value (with a different `constructorTagModifier`) for each of the types? Is it important to you that the same `Options` value can act differently on different types?

wi11dey:
@julmb That's fair, though yes I would like to reuse the same `Options` value as I'm using `DerivingVia` to derive `FromJSON` for a bunch of different types. I think the same logic of just using a different `Options` value could also apply to the existing `allNullaryToStringTag`, where you could just set your sum encoding to `UntaggedValue` for data types you know have all nullary constructors. If you feel strongly about not adding this though, I can live without it and close

julmb:
> If you feel strongly about not adding this though, I can live without it and close

Oh, this might be a misunderstanding, I am not a maintainer of `aeson` and I have no say in whether this PR gets accepted. I was just asking out of my own curiosity.

PR #1161, opened 2026-02-22
chore: fix typos

### LINKED ISSUE #1116: Mispellings across code
`unespected` is a spelling error in at least 4 places I can see, plus tests.

Here is an example: https://github.com/haskell/aeson/blob/c93d60d82388c184a9cf240f016a728c9d0e758f/src/Data/Aeson/Decoding/ByteString.hs#L164C19-L164C29

It would be good to remove such spelling errors.

e-kwsm:
fix #1116

### DIFF
--- a/benchmarks/bench/UnescapePureText1.hs
+++ b/benchmarks/bench/UnescapePureText1.hs
@@ -2,7 +2,7 @@
 --
 -- WARNING: This file is security sensitive as it uses unsafeWrite which does
 -- not check bounds. Any changes should be made with care and we would love to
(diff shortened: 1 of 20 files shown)


phadej:
Please separate comment and error message changes. We consider latter as breaking changes, so they cannot be merged as easily. (Tests needs to change == breaking change)

e-kwsm:
Thank you for your review. a53ef3ed743829d719640fcfdffdb58b1ad78c61 is the *breaking* change.

PR #1165, opened 2026-04-10
Add a `FromJSONKey` instance for `Data.Fixed`

VictorCMiraldo:
We have a `ToJSONKey (Fixed a)` instance:

https://github.com/haskell/aeson/blob/45d31f1bd9a0edbd6ba55fbcdd5082b159d33106/src/Data/Aeson/Types/ToJSON.hs#L1432-L1433

But there's no corresponding `FromJSONKey (Fixed a)` available. Because we have a `FromJSONKey Double`, it seems like this was just never added and never needed. I now have some `Map`s with fixed-point numbers as their keys and therefore would benefit from a `FromJSONKey (Fixed a)` instance. A lot of the scientific parsing functionality is internal to `aeson` and not exported, so I can't just add it the orphan instance on my project and call it a day, plus, if I can fix it for everyone, than it's better for everyone! :)

I tried reusing the `Scientific` parsing as much as possible and checked it does indeed reject large exponents in a `cabal repl`:

```
ghci> x = Object (KM.fromList [("4",String "a"),("5.32e4300",String "b")])
ghci> fromJSON x :: Result (M.Map (Fixed E12) String)
Error "found a number with exponent 4298, but it must not be greater than 1024"
```

### DIFF
--- a/src/Data/Aeson/Types/FromJSON.hs
+++ b/src/Data/Aeson/Types/FromJSON.hs
@@ -216,6 +216,12 @@ parseScientificText = scanScientific
     (\sci rest -> if T.null rest then return sci else fail $ "Expecting end-of-input, got " ++ show (T.take 10 rest))
     fail
 
+parseBoundedScientificText :: Text -> Parser Scientific
+parseBoundedScientificText t = parseScientificText t >>= rejectLargeExponent
+  where
+    rejectLargeExponent :: Scientific -> Parser Scientific
+    rejectLargeExponent s = withBoundedScientific' pure (Number s)
+
 parseIntegralText :: Integral a => String -> Text -> Parser a
 parseIntegralText name t =
     prependContext name $
@@ -1729,6 +1735,10 @@ instance (FromJSON a, Integral a) => FromJSON (Ratio a) where
 -- 'withScientific' if you want to allow larger inputs.
 instance HasResolution a => FromJSON (Fixed a) where
     parseJSON = prependContext "Fixed" . withBoundedScientific' (pure . realToFrac)
+
+instance HasResolution a => FromJSONKey (Fixed a) where
+  fromJSONKey = FromJSONKeyTextParser $ \t ->
+      realToFrac <$> parseBoundedScientificText t
 
 instance FromJSON Int where
     parseJSON = parseBoundedIntegral "Int"
--- a/src/Data/Aeson/Types/FromJSON.hs
+++ b/src/Data/Aeson/Types/FromJSON.hs
@@ -216,21 +216,21 @@ parseScientificText = scanScientific
     (\sci rest -> if T.null rest then return sci else fail $ "Expecting end-of-input, got " ++ show (T.take 10 rest))
     fail
 
-parseBoundedScientificText :: Text -> Parser Scientific
-parseBoundedScientificText t = parseScientificText t >>= rejectLargeExponent
+parseBoundedScientificTextTo :: (Scientific -> Parser a) -> String -> Text -> Parser a
+parseBoundedScientificTextTo toResultType name t =
+    prependContext name $
+            parseScientificText t
+        >>= rejectLargeExponent
+        >>= toResultType
   where
     rejectLargeExponent :: Scientific -> Parser Scientific
     rejectLargeExponent s = withBoundedScientific' pure (Number s)
 
+parseBoundedScientificText :: String -> Text -> Parser Scientific
+parseBoundedScientificText = parseBoundedScientificTextTo pure
+
 parseIntegralText :: Integral a => String -> Text -> Parser a
-parseIntegralText name t =
-    prependContext name $
-            parseScientificText t
-        >>= rejectLargeExponent
-        >>= parseIntegralFromScientific
-  where
-    rejectLargeExponent :: Scientific -> Parser Scientific
-    rejectLargeExponent s = withBoundedScientific' pure (Number s)
+parseIntegralText = parseBoundedScientificTextTo parseIntegralFromScientific
 
 parseBoundedIntegralText :: (Bounded a, Integral a) => String -> Text -> Parser a
 parseBoundedIntegralText name t =
@@ -1738,7 +1738,7 @@ instance HasResolution a => FromJSON (Fixed a) where
 
 instance HasResolution a => FromJSONKey (Fixed a) where
   fromJSONKey = FromJSONKeyTextParser $ \t ->
-      realToFrac <$> parseBoundedScientificText t
+      realToFrac <$> parseBoundedScientificText "Fixed" t
 
 instance FromJSON Int where
     parseJSON = parseBoundedIntegral "Int"
--- a/tests/PropertyKeys.hs
+++ b/tests/PropertyKeys.hs
@@ -6,7 +6,7 @@ module PropertyKeys ( keysTests ) where
 import Prelude.Compat
 
 import Control.Applicative (Const)
-import Data.Fixed
+import Data.Fixed (Fixed, E3, E6)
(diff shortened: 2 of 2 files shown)


src/Data/Aeson/Types/FromJSON.hs
```diff
@@ -216,6 +216,12 @@ parseScientificText = scanScientific
     (\sci rest -> if T.null rest then return sci else fail $ "Expecting end-of-input, got " ++ show (T.take 10 rest))
     fail
 
+parseBoundedScientificText :: Text -> Parser Scientific
+parseBoundedScientificText t = parseScientificText t >>= rejectLargeExponent
+  where
+    rejectLargeExponent :: Scientific -> Parser Scientific
+    rejectLargeExponent s = withBoundedScientific' pure (Number s)
```
phadej: This duplicates too much of `parseIntegralText`; I don't like this duplication.

tests/PropertyKeys.hs
```diff
@@ -6,6 +6,7 @@ module PropertyKeys ( keysTests ) where
 import Prelude.Compat
 
 import Control.Applicative (Const)
+import Data.Fixed
```
phadej: Wildcard import. Bad.

phadej (changes requested):

tests/PropertyKeys.hs
VictorCMiraldo: Respectfully disagree. Happy to oblige. :)

src/Data/Aeson/Types/FromJSON.hs
VictorCMiraldo: Reasonable! I refactored `parseIntegralText`; how does it look now?

VictorCMiraldo:
@phadej do you think we could merge this? Would be nice to eventually upstream these changes.

VictorCMiraldo:
Gentle reminder to @phadej or @Lysxia to have a look?

If this PR is undesired, then we should at least remove the `ToJSONKey Data.Fixed` instance too, keeping the
`ToJSONKey` and `FromJSONKey` instances symmetric. I'd be happy to do that if that's the preferred course of action.

Lysxia:
I'm inclined to accept this, but still the fact that you can have a JSON object with multiple JSON keys parse into the same Haskell key makes me uneasy.

@VictorCMiraldo  Does this not happen in your use case?

We do have the precedent of `FromJSONKey Double` and `ToJSONKey (Fixed a)` so my current thinking is that it's too late to fix and we should rather lean into aeson being Pareto efficient (= it does its job good enough for most, even if imperfectly). But I'm open to hearing thoughts about issues the non-injectivity of `FromJSONKey` might cause and ideas of how to mitigate them.

stevladimir:
In recent GHCs one can add warnings to instances (https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/pragmas.html#warning-and-deprecated-pragmas). Also one can specify warning group, so I can imagine something like
``` haskell
instance
#if MIN_VERSION_GLASGOW_HASKELL(...)
  {-# WARNING in "x-aeson-non-injective" "Blah blah blah" #-}
#endif
  FromJSONKey (Fixed a)
```
And then those who still want to rely on this instance could add into .cabal `ghc-options: -Wno-x-aeson-non-injective` or per-file pragma.

Ofc currently this works only for recent GHCs, but situation will improve with time

VictorCMiraldo:
> I'm inclined to accept this, but still the fact that you can have a JSON object with multiple JSON keys parse into the same Haskell key makes me uneasy.
> @VictorCMiraldo Does this not happen in your use case?

I really dislike that too... This PR does not introduce the non-injectivity though. The problem runs deeper and it's about relying on `scientific` under the hood. Even with `To/FromJSONKey Int32`, the keys `"1000"` and `"1e3"` parse to the same value.

> And then those who still want to rely on this instance could add into .cabal ghc-options: -Wno-x-aeson-non-injective or per-file pragma.

Then you'll have to add that to all numerical instances of `To/FromJSONKey` too!

---

Some context might be useful: In my day job, we were exporting some `Map Pico Whatever`, and it was going fine (`aeson` provides a `ToJSONKey (Fixed e)` instance). At some point, we started needing to import these maps too and that's when I noticed there was no corresponding `FromJSONKey`. At first, I wanted to write my own instance, in our own codebase, but all the nice and safe number-parsing functions from `aeson` are not exported and memory safety for us is a real issue. Those maps could be created by malicious users, so I don't want to try parsing `"1e123123123123"`. 

I think that the `to` and `from` family of instances provided by `aeson` should to be symmetric: for every `t`, there exists `ToJSONKey t` iff there exists `FromJSONKey t`. I see two ways forward:

(A) Given that the non-injectivity comes from parsing numerical keys first as `scientific` then converting to their designated type, pragmatically, I think it's best is to accept this quirk and merge this PR, since it fixes a real inconsistency in the library API. This would be backwards compatible. This is why I made this PR.

(B) One could also argue that this non-injectivity is pretty bad (which, personally, I believe it is!) and then we reject this PR and make a different PR to (a) delete the `ToJSONKey (Fixed ...)`, (b) expose the nice memory-safe numerical-parsing functions together with docs for users to make their own `To/FromJSONKey` instances. This would *not* be backwards compatible but would help us remove one source of non-injectivity from `aeson` the library and let users choose by themselves if it's something they care about. 

Personally, I think that trying to parse `"1.23e3"` through `FromJSONKey Int32` should be a parse error, but it's not:
image

But this is way beyond the scope of this PR

stevladimir:
> Then you'll have to add that to all numerical instances of `To/FromJSONKey` too!

Yeah. The point is that you have means to keep questionable instances (to be backwards compatible), but inform the users that they are unsafe in some sense

phadej:
The

```json
{ "a": 1, "a": 2 }
```

is "valid" JSON, where `aeson` simply doesn't preserve duplicate key. [RFC](https://www.rfc-editor.org/info/rfc8259/#section-4) says just

> The names within an object SHOULD be unique.

So `aeson` cannot reject such values. (Though it makes subjective choice to just drop another one; that said using raw decoding you can deal with duplicates).

```json
{ "1000": 1, "1e3": 2}
```

where you interpret object keys as numerals is not much different. There may be valid reasons to have those separate, but arguably if those keys are to be interpreted numerically, you can just drop them.

So IMHO, either `aeson` should have a way to deal with duplicate keys (in `Value`) or just don't consider non-injective `ToJSONKey` to be a real issue. If it's an issue in some practical application, people may use own instance (or go even further to lower level implementation). 

As an extreme example consider

```
echo '{"\x20": 1, "\u0020": 2}' | python3 -m json.tool
{
    " ": 2
}
```

Someone might argue that these keys are different and should be distinguishable by the library because their literals are different. I'd ignore that person opinion.

phadej:
> Personally, I think that trying to parse "1.23e3" through FromJSONKey Int32

That is a fair take, but OTOH `1.23e3` is valid for `FromJSONKey Int32`; and someone else (e.g. myself) might like that parsers of values and keys agree when they can.

As I pointed out previously, `Value` doesn't really distinguish how different literals were encoded. For numerals for some degree, as `Scientific` values are not normalised, but I would like for `aeson` to not rely on that too hard. There is exponent check but whether it's `1.23e4` or `12.3e3` doesn't matter much for that, as the bound is arbitrary.

VictorCMiraldo:
@phadej makes a good point, I change my mind: I actually agree with it. Numbers are not special, the examples with `"a"` and `"\x20"` make a strong case. I guess this means this PR is the right way forward, the non-injectivity is a non-issue.

Lysxia:
Let's go with this. Thanks!

PR #1182, opened 2026-07-09
Clarify comment about exceptions for decodeFile

### LINKED ISSUE #1181: What exception is thrown when `decodeFile` is used
Documentation says 

> "Throws an 'Exception' when the file is missing."

Is it `AesonException`?

What happens if user has no permissions to access the file?

Please clarify

Lysxia:
Close #1181

### DIFF
--- a/src/Data/Aeson.hs
+++ b/src/Data/Aeson.hs
@@ -200,7 +200,10 @@ encodeFile fp = L.writeFile fp . encode
 -- This function parses immediately, but defers conversion.  See
 -- 'json' for details.
 --
--- Throws an 'Exception' when the file is missing.
+-- Throws an exception when the file cannot be accessed
+-- (e.g., it is missing or permissions are invalid).
+-- This function uses 'Data.ByteString.readFile' without handling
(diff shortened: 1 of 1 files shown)


src/Data/Aeson.hs
```diff
@@ -200,7 +200,9 @@ encodeFile fp = L.writeFile fp . encode
 -- This function parses immediately, but defers conversion.  See
 -- 'json' for details.
 --
--- Throws an 'Exception' when the file is missing.
+-- Throws an 'Control.Exception.IOException' when the file cannot be accessed
```
phadej: https://hackage-content.haskell.org/package/bytestring-0.12.2.0/docs/Data-ByteString.html#v:readFile doesn't document any particular exception behavior.

Thus I'd rather not explain something which is not actually documented to be the case.

I'd rather say `decodeFileStrict` uses `readFile` and doesn't handle any exceptions thrown by the latter.

If users then wonder what `B.readFile` may throw, they should ask `bytestring` maintainers.

phadej (approved):

PR #1185, opened 2026-07-09
Add comment about JSONPath accumulated in reverse in Parser

phadej:
I was wondering how this works. it took a while to find where `reverse` happens.

It might be good idea to define `newtype ReversedJSONPath = ReversesJSONPath JSONPath`, and use that in `Parser`. That way the `reverse` would be harder to forget, but such change might be breaking, i'm not sure how much of internals is exposed.


Lysxia (approved):
Thanks!

PR #1186, opened 2026-07-09
Keep track of indices in listParser errors

### LINKED ISSUE #1184: listParser doesn't update path index
```haskell
*PoeNinja Data.Vector Data.Maybe> Data.Aeson.Types.iparse (Data.Aeson.Types.listParser (Data.Aeson.withObject "Foo" $ \obj -> obj Data.Aeson..: "a")) (fromJust (Data.Aeson.decode "[{\"a\": true }]") :: Value) :: Data.Aeson.Types.IResult [Int]
IError [Key "a"] "parsing Int failed, expected Number, but encountered Boolean"
```

There is no `Index 0` in the path. I think it affects `[a]` instances as these use `listParser`.

Lysxia:
Close #1184

### DIFF
--- a/src/Data/Aeson/Types/FromJSON.hs
+++ b/src/Data/Aeson/Types/FromJSON.hs
@@ -690,7 +690,7 @@ omittedField2 = liftOmittedField2 omittedField omittedField
 
 -- | Helper function to use with 'liftParseJSON'. See 'Data.Aeson.ToJSON.listEncoding'.
 listParser :: (Value -> Parser a) -> Value -> Parser [a]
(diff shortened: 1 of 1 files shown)


Lysxia:
I wasn't sure whether changing error messages required bumping the major version number but there is already multiple precedent in aeson where that only bumps the minor version number.

phadej:
A small nitpick: This change doesn't make any tests fail, though there is a part test suite testing the error messages. I'd liked to see a test change and/or addition as well!

Lysxia:
Good idea. I should make a PR template to remind myself to add tests.

PR #1190, opened 2026-08-28
Fix optimization-dependent invalid Map decoding

### LINKED ISSUE #1169: unsafe rewrite rule for Functor instance of FromJSONKeyFunction
## Summary

I noticed aeson parsing into an invalid `Map k v`. This only happened with optimisations enabled and not in ghci.

The combination of `FromJSONKeyCoerce`, the `Functor` instance of `FromJSONKeyFunction` and the rewrite rule https://github.com/haskell/aeson/blob/cd59793ca2405353ecee65f25b18ad7610104914/src/Data/Aeson/Types/FromJSON.hs#L486-L489 can produce invalid maps.

## To reproduce

Have newtype `MyKey` with an Ord instance which is incompatible to that of `Text`. e.g. `deriving via Down Text`.
```haskell
newtype MyKey = MyKey Text deriving Ord via (Down Text)
```
Implement:

```haskell
instance FromJSONKey MyKey where
  fromJSONKey = MyKey <$> fromJSONKey
```

When you now parse an object with multiple keys into something of the type `Map MyKey a` with aeson the resulting Map will be invalid (a.k.a. `Data.Map.Strict.valid`) will be `False`.

## Explanation

The functor instance of FromJSONKeyFunction in aeson is correctly implemented as:

```haskell
   MyKey <$> fromJSONKey
== fmap MyKey FromJSONKeyCoerce
== FromJSONKeyText (MyKey . coerce)
== FromJSONKeyText coerce
```

But with above rewrite rule I get:

```haskell
   MyKey <$> fromJSONKey
== fmap coerce FromJSONKeyCoerce
== coerce FromJSONKeyCoerce
```

which is wrong because `FromJSONKeyCoerce` explicitely says that it expects the `Ord` instance to be compatible.

## Full reproducer

This reproducer works with ghc 9.10.3 and aeson 2.2.3.0
```haskell
{-# LANGUAGE DerivingVia #-}

import Data.Aeson
import Data.Map.Strict
import Data.Ord
import Data.Text

newtype MyKey = MyKey Text
    deriving (Eq, Ord) via (Down Text)
    deriving (FromJSON) via Text

instance FromJSONKey MyKey where
    fromJSONKey = MyKey <$> fromJSONKey

myMap :: Maybe (Map MyKey Text)
myMap = decode "{\"a\":\"a\",\"b\":\"b\"}"

main :: IO ()
main = print $ valid <$> myMap
```

```bash
> ghc -O0 test.hs
> ./test
Just True

> ghc -O1 test.hs
> ./test
Just False
```

PaleRoses:
Fixes #1169.

Remove the unsound `"FromJSONKeyCoerce: fmap coerce"` RULE. `mapKeyO` reuses the source tree only when mapped keys stay strictly ordered; otherwise it rebuilds the map.

Tests cover reversed ordering and collisions. The full suite passes at `-O0` and `-O2`. All 32 Map benchmarks pass the 5% regression check; affected cases are 29–41% faster.


src/Data/Aeson/Internal/Functions.hs
```diff
@@ -35,8 +35,23 @@ mapKey :: (Eq k2, Hashable k2) => (k1 -> k2) -> H.HashMap k1 v -> H.HashMap k2 v
...
 
+    -- The outer 'Nothing' is an ordering obstruction; the inner one is the
+    -- empty suffix before its greatest projected key is seen.
+    checkKeyOrder sourceKey _ maybeNextKey = do
+        nextKey <- maybeNextKey
+        let currentKey = fk sourceKey
+        if maybe True (currentKey <) nextKey
+        then Just (Just currentKey)
+        else Nothing
+{-# INLINE mapKeyO #-}
```
Lysxia, line 57: I don't think this is the right place for this kind of optimization.

- It's orthogonal to fixing the main issue at hand.
- When the keys are not ordered you end up paying the cost of applying `fk` twice to each key, which might be expensive if it's doing nontrivial parsing. I'm also not sure this optimization is worthwhile since `mapKeysWith` already takes care of running in linear time if the mapping is monotonic.

tests/UnitTests/FromJSONKey.hs
```diff
@@ -1,49 +1,74 @@
...
     deriving (FromJSONKey)
 
 newtype MyText' = MyText' Text
+    deriving FromJSON via Text
 
 instance FromJSONKey MyText' where
     fromJSONKey = fmap MyText' fromJSONKey
-    fromJSONKeyList = error "not used"
+
+newtype ReverseText = ReverseText Text
```
Lysxia, line 27: Why not use `Down Text`?

Update: there is no `FromJSON` for `Down`

tests/UnitTests/FromJSONKey.hs
```diff
@@ -1,49 +1,74 @@
...
+
+assertDecodedMapIsValid :: Assertion
+assertDecodedMapIsValid = fmap Map.valid decodedMap @?= Just True
+  where
+    decodedMap = decode "{\"a\":\"a\",\"b\":\"b\"}" :: Maybe (Map ReverseText Text)
+
+assertDecodedCollidingMap :: Assertion
+assertDecodedCollidingMap = decodedMap @?= Just (Map.singleton (FoldedText "a") "upper")
+  where
+    decodedMap = decode "{\"A\":\"upper\",\"a\":\"lower\"}" :: Maybe (Map FoldedText Text)
```
Lysxia, line 74: This test relies on `KeyMap` being ordered. That's why it sometimes fails the flag `ordered-keymap` disabled.

